PR #9, opened 2016-06-15
Patch allow-non-breakable-words on '-'

michelebariani:
Hi,

I found that the allow-non-breakable-words is not working as I expected (not sure if we have the same expectations!) with long URLs preceeded by '- ' instead of '# '.

The patch implements (test included) that expectation.

HTH,
Michele

PS: I came up with the use case using yamllint on Ansible playbooks


adrienverge:
> I found that the allow-non-breakable-words is not working as I expected (not sure if we have the same expectations!) with long URLs preceeded by '- ' instead of '# '.

We do have the same expectations :)

Thanks for contributing @michelebariani!

PR #17, opened 2016-09-12
line_length: Allow mapping values with long unbreakable lines

allanlewis:
Previously, the unbreakable line check only worked for array items (`- foo`) or comments (`# foo`): this branch extends it to also work with mapping values (`key: foo`).

### DIFF
--- a/tests/rules/test_line_length.py
+++ b/tests/rules/test_line_length.py
@@ -113,7 +113,35 @@ class LineLengthTestCase(RuleTestCase):
                    'long_line: http://localhost/very/very/long/url\n'
                    '...\n', conf, problem=(2, 21))
 
+        conf = ('line-length: {max: 20, allow-non-breakable-words: yes}\n'
+                'trailing-spaces: disable')
+        self.check('---\n'
+                   'loooooooooong+word+and+some+space+at+the+end       \n',
+                   conf, problem=(2, 21))
+
+    def test_non_breakable_inline_mappings(self):
         conf = 'line-length: {max: 20, ' \
                'allow-non-breakable-inline-mappings: yes}'
         self.check('---\n'
-                   'long_line: http://localhost/very/very/long/url\n', conf)
+                   'long_line: http://localhost/very/very/long/url\n'
+                   'long line: http://localhost/very/very/long/url\n', conf)
+        self.check('---\n'
+                   '- long line: http://localhost/very/very/long/url\n', conf)
+
+        self.check('---\n'
+                   'long_line: http://localhost/short/url + word\n'
+                   'long line: http://localhost/short/url + word\n',
+                   conf, problem1=(2, 21), problem2=(3, 21))
+
+        conf = ('line-length: {max: 20,'
+                '              allow-non-breakable-inline-mappings: yes}\n'
+                'trailing-spaces: disable')
+        self.check('---\n'
+                   'long_line: and+some+space+at+the+end       \n',
+                   conf, problem=(2, 21))
+        self.check('---\n'
+                   'long line: and+some+space+at+the+end       \n',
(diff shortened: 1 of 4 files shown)


allanlewis:
This is breaking on Python 3: I'll fix it.

adrienverge:
Hey @allanlewis,

To be honest I don't really like this change. In my opinion, a one-line mapping should be split like this:

```yaml
# problem
long_line: http://localhost/very/very/very/very/very/very/long/url
```

```yaml
# no problem
long_line:
  http://localhost/very/very/very/very/very/very/long/url
```

However if you really think it's needed, I'm not opposed to adding it. What about adding a `allow-inline-mappings` options to the `line-length` rule?

allanlewis:
@adrienverge, I see what you mean. However, many of the cases in my project are in Ansible's `get_url` module, which has a `url` key. Adding a new-line and an indent with such a short key doesn't look nice, IMHO:
```yaml
get_url:
  url:
    http://localhost/very/very/very/very/very/very/long/url
```
I'd much prefer:
```yaml
get_url:
  url: http://localhost/very/very/very/very/very/very/long/url
```
So I'll refactor this as you suggest.

adrienverge:
Just as a note, there is another alternative (which I use often):

```yaml
get_url:
  url: "http://localhost/very/very/very/very/very\
        /very/very/very/very/very/long/url"
```

or equivalently:

```yaml
get_url:
  url:
    "http://localhost/very/very/very/very/very\
     /very/very/very/very/very/long/url"
```

allanlewis:
@adrienverge, I've implemented a version of your suggestion: please let me know your thoughts.

adrienverge:
Hi @allanlewis,

The commits look clean, and I find your option naming great! I'm also OK to
enable it in `relaxed` conf.

But I see four issues in your implementation:

1. Because of `yaml.safe_load`, your implementation crashes yamllint when a dict
   spans on more than one line. For example yamllint crashes on this valid input:

   ```yaml
   - {a: "http://localhost/very/very/very/very/very/very/long/url"
      }
   ```

2. Because of `popitem()`, it does not check inline dict that have more than one
   key/value pair. Example: this is reported OK (but should not):

   ```yaml
   dict:
     {a: long long long long long long long, b: nospace}
   ```

3. When the mapping is in a sequence, your new option does not work. For
   example this is still reported as too long:

   ```yaml
   - long_line: http://localhost/very/very/long/url
   ```

4. Currently extra spaces at the end of a line **do** trigger a `line_length`
   error (independently from the `trailing-spaces` rule).

   When enabling `allow-non-breakable-inline-mappings`, this behaviour changes.

   ```yaml
   # this triggers a line-length report
   long+line+and+some+spaces+at+the+end       <-- extra spaces
   # this does not
   long_line: and+some+space+at+the+end       <-- extra spaces
   ```

adrienverge:
I just fixed these problems (and added more test cases) in an extra commit, see https://github.com/adrienverge/yamllint/commits/allanlewis-improve-unbreakable

If it suits you, squash your commits and add mine to your PR, and I'll merge!

allanlewis:
Thanks, @adrienverge. FYI I enabled the "allow edits from maintainers" option: I wouldn't have minded if you'd added your commit directly. Anyway, your commit looks good and it certainly improves the robustness of this change - thanks for your help!

Merging this should unblock me to use `yamllint` in my project to lint all of my Ansible and Jenkins Job Builder configuration.

adrienverge:
> FYI I enabled the "allow edits from maintainers" option: I wouldn't have minded if you'd added your commit directly.

I wasn't aware of this feature. I'll try it next time!

Thanks for your contribution. I'll release a new version.

PR #19, opened 2016-09-22
Add the ability to run custom rules included in a project

pdericson:
This is a little rough but I figured I'd raise it now to gauge interest.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -15,6 +15,7 @@ If ``-c`` is not provided, yamllint will look for a configuration file in the
 following locations (by order of preference):
 
 - ``.yamllint`` in the current working directory
+- ``.yamllint/config`` in the current working directory
 - ``$XDG_CONFIG_HOME/yamllint/config``
 - ``~/.config/yamllint/config``
 
--- a/docs/custom_rules.rst
+++ b/docs/custom_rules.rst
@@ -1,0 +1,62 @@
+Custom Rules
+============
+
+There are times when you might like to add custom rules to your
+project.  This could be because the rules you'd like to enforce are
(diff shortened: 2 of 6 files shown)


adrienverge:
Hey @pdericson,

Nice work!

Although the idea of "pluggable rules" looks good, I don't like yamllint automatically loading rules from the current dir. This would allow untrusted sources (example: a pull request on GitHub) to run arbitrary Python code. Maybe a `--plugin` option would be better? And it would allow pluggable rules installed system-wide.

Could you give more details about your use-case? Is the "truthy" rule just an example, or something you really want to lint?

pdericson:
Hey @adrienverge, I'll rework my PR to make this explicitly opt-in like you suggest.  Regarding the "truthy" example - I have been thinking about creating an ansible linter for a while and so this example would be something I'd like to include.

adrienverge:
Hi @pdericson,

I can see the use of such a rule!

I guess the motivation is to prevent pyyaml from altering YAML source. Because by default, it turns

```yaml
1: y
2: yes
3: on
4: true
5: True
```
into
```json
{
    "1": "y", 
    "2": true, 
    "3": true, 
    "4": true, 
    "5": true
}
```
and
```yaml
y:    1
yes:  2
on:   3
true: 4
True: 5
```
into
```json
{
    "y": 1, 
    "true": 5
}
```

To be honest I don't really like the pluggable rules idea, but I would warmly welcome a `quote-truthy` rule contribution!

pdericson:
@adrienverge I agree, so far the only yaml linting idea I have had for ansible is the truthy rule which as you have pointed out can be made generic.

I'm going to replace this PR with #22

PR #22, opened 2016-09-28
Add rule: truthy, to forbid truthy values that are not quoted.

pdericson:


tests/test_cli.py
```diff
@@ -279,7 +279,7 @@ def test_run_non_ascii_file(self):
         # Make sure the default localization conditions on this "system"
         # support UTF-8 encoding.
         loc = locale.getlocale()
-        locale.setlocale(locale.LC_ALL, 'C.UTF-8')
+        locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
```
pdericson: `C.UTF-8` didn't work for me on OS X...

adrienverge: Please don't include this is this pull-request, we can do another one if needed.

(By the way `en_US.UTF-8` isn't supposed to be present on non-English-speaking systems, and I know it isn't on Debian build system.)

yamllint/conf/relaxed.yaml
```diff
@@ -26,3 +26,5 @@ rules:
   line-length:
     level: warning
     allow-non-breakable-inline-mappings: yes
+  truthy:
+    level: warning
```
adrienverge: I'd rather have it disabled in `relaxed` rule. I know a bunch of people that are really attached to their `yes`/`no` in Ansible...

tests/rules/test_truthy.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 Adrien Vergé
```
adrienverge: Hey! I didn't write this. ;)

yamllint/rules/truthy.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 Adrien Vergé
```
adrienverge: Please put your name here :)

tests/test_spec_examples.py
```diff
@@ -119,6 +120,9 @@ class SpecificationTestCase(RuleTestCase):
     'example-8.22': ('indentation: disable\n'),
     'example-10.1': ('colons: {max-spaces-before: 2}\n'),
     'example-10.2': ('indentation: {indent-sequences: no}\n'),
+    'example-10.5': ('truthy: disable\n'),
```
adrienverge: Why this one?

tests/test_spec_examples.py
```diff
@@ -65,6 +65,7 @@ class SpecificationTestCase(RuleTestCase):
     'example-2.16': ('empty-lines: {max-end: 1}\n'),
     'example-2.18': ('empty-lines: {max-end: 1}\n'),
     'example-2.19': ('empty-lines: {max-end: 1}\n'),
+    'example-2.21': ('truthy: disable\n'),
```
adrienverge: Why this one?

yamllint/rules/truthy.py
```diff
@@ -0,0 +1,56 @@
...
+          'FALSE', 'False',  # false is a boolean
+          'ON', 'On', 'on',
+          'OFF', 'Off', 'off']
+
+
+def check(conf, token, prev, next, nextnext, context):
+    if isinstance(token, yaml.tokens.ScalarToken):
+        if token.value in TRUTHY and token.style is None:
+            yield LintProblem(token.start_mark.line + 1,
+                              max(1, token.start_mark.column),
```
adrienverge: This should be `token.start_mark.column + 1` (and you can remove `max()`).

tests/rules/test_truthy.py
```diff
@@ -0,0 +1,33 @@
...
+        conf = 'truthy: disable'
+        self.check('---\n1: True\n', conf)
+        self.check('---\nTrue: 1\n', conf)
+
+    def test_enabled(self):
+        conf = 'truthy: enable\n'
+        self.check('---\n1: True\n', conf, problem=(2, 3))
+        self.check('---\nTrue: 1\n', conf, problem=(2, 1))
+        self.check('---\n1: "True"\n', conf)
+        self.check('---\n"True": 1\n', conf)
```
adrienverge: Could you add this?
```py
        self.check('---\n'
                   '!!seq [\n'
                   '  !!bool true, !!bool false,\n'
                   '  !!bool "false", !!bool "FALSE",\n'
                   '  !!bool "true", !!bool "True",\n'
                   '  !!bool "false", !!bool "FALSE",\n'
                   ']\n', conf)
        self.check('---\n'
                   '!!seq [\n'
                   '  !!bool true, !!bool True,\n'
                   ']\n', conf, problem=(3, 23))
```

tests/test_cli.py
pdericson: I have reverted this change, I can run the tests in docker so all good

```
docker run -v $PWD:/mnt -w /mnt python:2.7 python setup.py test
```

yamllint/rules/truthy.py
pdericson: I copied this from commas.py, I can see why it's invalid now :)

adrienverge:
That's better, thanks!

Have you got a chance to consider my other remarks? (the extra tests with `!!bool` and the spec examples you disable -- I think two of them are working fine and shouldn't be disabled)

tests/rules/test_truthy.py
```diff
@@ -0,0 +1,47 @@
...
+        self.check('---\n'
+                   '!!seq [\n'
+                   '  !!bool true, !!bool false,\n'
+                   '  !!bool "false", !!bool "FALSE",\n'
+                   '  !!bool "true", !!bool "True",\n'
+                   '  !!bool "false", !!bool "FALSE",\n'
+                   ']\n', conf)
+        self.check('---\n'
+                   '!!seq [\n'
+                   '  !!bool true, !!bool True,\n'
```
pdericson: @adrienverge I'm wondering if `!!bool True` should be allowed by the truthy rule.  I haven't figured out yet if the explicit tag type is available to inspect or is lost during parsing by pyyaml...

pdericson:
Sorry, my session ended abruptly yesterday, I was addressing feedback during my commute

adrienverge:
@pdericson this is great, thanks for your contribution!

ghiculescu:
just a quick heads up that the version of `yamllint` that apt-get installs doesnt include this rule. the pip version does.

adrienverge:
@ghiculescu Are you using Debian? Ubuntu?

@debalance is kindly maintaining packages for Debian, which means new versions should also end up in next Ubuntu versions. For really up-to-date versions, we would need an Ubuntu maintainer (help wanted!)

PR #23, opened 2016-10-01
feat(truthy): Allow explicit types

adrienverge:
With this change, we don't require quotes for truthy values that are explicitely typed. For instance, the following examples are all considered valid:

```yaml
string1: !!str True
string2: !!str yes
string3: !!str off
encoded: !!binary |
           True
           OFF
           pad==  # this decodes as 'N\xbb\x9e8Qii'
boolean1: !!bool true
boolean2: !!bool "false"
boolean3: !!bool FALSE
boolean4: !!bool True
boolean5: !!bool off
boolean6: !!bool NO
```

### DIFF
--- a/yamllint/rules/truthy.py
+++ b/yamllint/rules/truthy.py
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Use this rule to forbid truthy values that are not quoted.
+Use this rule to forbid truthy values that are not quoted nor explicitly typed.
 
 This would prevent YAML parsers to tranform ``[yes, FALSE, Off]`` into ``[true,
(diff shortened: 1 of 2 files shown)


adrienverge:
@pdericson what do you think?

pdericson:
@adrienverge very nice - I didn't realise until a few days ago that you could do explicit typing in yaml so I think this is really useful.

yamllint/rules/truthy.py
```diff
@@ -15,7 +15,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Use this rule to forbid truthy values that are not quoted.
+Use this rule to forbid truthy values that are not quoted nor explicitely
```
pdericson: s/explicitely/explicitly ?

adrienverge: oops

adrienverge:
Thanks @pdericson!

PR #27, opened 2016-10-24
Strict mode

jsok:
This PR adds a strict mode which returns a non-zero exit code when there are only warnings.


adrienverge:
I'm not really in favor of this change, because in my opinion the only purpose of having errors AND warnings is the non-zero return code.

But this PR is really nice and clean, and I have to admit adding this option cannot hurt.

jsok:
Thanks, I appreciate it.

PR #31, opened 2016-12-23
support for pre-commit from yelp

bootswithdefer:


adrienverge:
Squashed and merged. Thanks!

PR #38, opened 2017-02-17
Add min-spaces-inside-empty, max-spaces-inside-empty to braces and brackets

jhriggs:
This allows empty objects (e.g. `[]`, `{}`) to be handled separately from non-empty ones. In my case, I want `min-spaces-inside: 1` and `max-spaces-inside: 1` for braces and brackets, but I want empty items to have no spaces (`min-spaces-inside-empty: 0` and `max-spaces-inside-empty: 0`).

This has been done as backward-compatible as possible, defaulting the new settings to `-1` and using the non-empty configuration variable value in the the case of `-1`. Thus, users should see no difference in behavior of existing configurations unless one of the new variables is set.

### DIFF
--- a/tests/rules/test_braces.py
+++ b/tests/rules/test_braces.py
@@ -32,11 +32,19 @@ class ColonTestCase(RuleTestCase):
                    'dict7: {   a: 1, b, c: 3 }\n', conf)
 
     def test_min_spaces(self):
-        conf = 'braces: {max-spaces-inside: -1, min-spaces-inside: 0}'
+        conf = ('braces:\n'
+                '  max-spaces-inside: -1\n'
+                '  min-spaces-inside: 0\n'
+                '  max-spaces-inside-empty: -1\n'
+                '  min-spaces-inside-empty: -1\n')
         self.check('---\n'
                    'dict: {}\n', conf)
 
-        conf = 'braces: {max-spaces-inside: -1, min-spaces-inside: 1}'
+        conf = ('braces:\n'
+                '  max-spaces-inside: -1\n'
+                '  min-spaces-inside: 1\n'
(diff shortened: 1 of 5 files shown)


yamllint/conf/relaxed.yaml
```diff
@@ -6,9 +6,11 @@ rules:
   braces:
     level: warning
     max-spaces-inside: 1
+    max-spaces-inside-empty: -1
```
adrienverge: Did you mean `1`?

If you really meant `-1`, you can remove these, because the `relaxed` rule extends `default`.

adrienverge:
Good PR with tests and clear documentation, thanks!

Could you add tests with both `m*-spaces-inside` and `m*-spaces-inside-empty` set, to make sure it behaves as expected? For instance with:

```yaml
brackets:
  max-spaces-inside: 0
  min-spaces-inside-empty: 1
```

I would expect `[ ]` and `[42]` to pass and `[]` and `[ 42]` to fail.

yamllint/conf/relaxed.yaml
jhriggs: This was a remnant from before I added backward compatibility (setting the *-empty values to their non-empty equivalents with -1). I removed it.

jhriggs:
I added tests that I hope cover many of the mixed scenarios, and I removed the max-spaces-inside-empty settings from relaxed.

adrienverge:
That's great, thanks!

jhriggs:
@adrienverge - Any timeline on uploading a new package for this and a couple of the other recent merges? We would like to be able to run "official" packages. :-)

adrienverge:
@jhriggs Sure! v1.7.0 released :)

PR #41, opened 2017-02-28
Doc: Add completor Vim plugin

polyzen:


adrienverge:
Thanks for contributing!

Syntastic and Completor look like to different things to me (a background checker and a typeahead completor), so can you rephrase a bit (especially "If you instead use") before I merge?

polyzen:
:facepalm:.. Got so caught up with deciding how to add the information in, that I penned in the wrong plugin! Fixing.

PR #46, opened 2017-03-25
Add information about Emacs integration

krzysztof-magosa:
Hi.

I've recently created Flycheck plugin which integrates YAMLint into Emacs. I'm waiting for MELPA acceptance, but in meantime it can be easily installed directly from my GitHub repository. This pull request adds information about that plugin to editors section of documentation.

Cheers.


adrienverge:
Thanks!

PR #51, opened 2017-05-28
Use argparse mutually_exclusive_group for --config-file and --config-data

sedrubal:
This does the same as your solution :wink:


adrienverge:
Looks great, thanks!

Issue #43, opened 2017-03-01
Add configuration of files to ignore


Dbz:
A great feature for this project would be the ability to exclude certain files. This way, if you are adding yamllint to a project, you can ignore every file with issues until you get around to fixing them. Right now, the only way to achieve this is through comments on every file.

A good way to accomplish this is in the rules configuration:

``` YAML
rules:
  indentation:
    indent-sequences: false
    spaces: 2
    exclude:
    - 'configs/config_yaml.yml'
```

A second add-on feature would be to allow these excludes to be auto-generated. This second feature helps make it easy to integrate yamllint into an existing codebase.

Right now our project has 11400 errors, so these features would make yamllint an idea yaml linter for our project, but as it stands, there is too much noise to lint new yamls as they come.

adrienverge:
Hi!

Looks like a duplicate of #5.

> What you describe looks very useful. However I'm not sure such filtering should be part of yamllint's logic, especially because it would be redundant with common UNIX tools.
>
> For instance, what about using find?
>
> ```shell
> yamllint $(find foo -name '*.yml' ! -name 'b*.yml' ! -name 'd.yml')
> ```

Dbz:
Hey Adrien,

This is not a duplicate of #5. We want to exclude files from specific rules and not the linter in general, so this is not something that can be done with UNIX tools. I feel very strongly that this is a very important feature for a production yaml linter, and I want to give you insight to our pipeline:

1. Devs need to be able to run the linter easily.

We have ten dev teams around the company committing to this repository. They can be given the line (or script) like `yamllint $(find foo -name '*.yml' ! -name 'b*.yml' ! -name 'd.yml')` with fifty more file names included to run, but it causes some issues. For example, the is no up-to-date list of the work that needs to be done to clean these files up, and there is also no priority on the syntax vs. style issues. We want linting to be very easy, and we want refactoring to be easy.

2. We have indentation issues that are causing teams who work on the same yaml many problems. 

We want to use a linter to fix the most important problems in the yamls first. Inconsistencies are causing issues for many teams. We render partial yamls inside yamls, and indentation/duplicate keys, etc. are causing problems and need to be dealt with immediately. If we have to ignore entire files, then we may miss these issues as they pop up.

3. The CI pipeline needs to run the linter, and that requires constantly updating the yamllint run command.

[...]

PR #60, opened 2017-07-18
Change format of ignore patterns

sedrubal:
Use yaml lists:

```yaml
ignore:
  - generated
  - "*.template.yaml"
```

instead of multiline strings:

```yaml
ignore: |
  generated
  *.template.yaml
```

This also fixes the wrong error messages when using the new config
format with ymallit v1.8.0/v1.8.1.

The only problem I see is that this will break config files with the
latter ignore pattern format. But AFAIK there is no major distribution packaging >= v1.8.0:

- https://apps.fedoraproject.org/packages/yamllint
- https://packages.debian.org/sid/yamllint
- https://packages.ubuntu.com/yakkety/yamllint

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -125,20 +125,20 @@ Specific files can be ignored (totally or for some rules only) using a
 .. code:: yaml
 
  # For all rules
- ignore: |
-   *.dont-lint-me.yaml
-   /bin/
-   !/bin/*.lint-me-anyway.yaml
+ ignore:
+   - "*.dont-lint-me.yaml"
+   - /bin/
+   - "!/bin/*.lint-me-anyway.yaml"
 
  rules:
    key-duplicates:
-     ignore: |
-       generated
-       *.template.yaml
+     ignore:
+       - generated
+       - "*.template.yaml"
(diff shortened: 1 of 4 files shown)


adrienverge:
Hi @sedrubal,

I clearly understand the preference for a list rather than a text file, unfortunately now the feature is out and yamllint 1.8.1 is used in many projects (it is commonly installed via `pip` rather than Fedora / Debian repos; see Ansible, OpenStack etc.)

It would have been great to have this kind of remark when this feature was in "RFC", although I guess it was not advertized enough: https://github.com/adrienverge/yamllint/issues/43#issuecomment-306583928
For information (but it doesn't have much importance now) I first chose the file-like option to mimic a `.gitignore` file.

Anyway thanks to your [comment](https://github.com/adrienverge/yamllint/commit/df26cc0438f996c03a9b61aca3a8b35166b09a99#commitcomment-23168185) I changed the error message to be less confusing ;)

sedrubal:
All right. Sorry for not reading the discussion 🙈

PR #64, opened 2017-08-19
Cli: Coloring with colorama

kapsh:


kapsh:
I think that colored output is easier and more maintainable with colorama. And it is widely used library so this isn't bad to add such dependency.

yamllint/cli.py
```diff
@@ -125,6 +130,7 @@ def run(argv=None):
         sys.exit(-1)
 
     max_level = 0
+    underlined = '\033[4m'  # w/a - see colorama issue #38
```
kapsh: This is ugly hack, but if you consider changing file name style to bright (\e[1m) - this can be done like in the orher places.

adrienverge:
Hi @kapsh, thanks for contributing!

Unfortunately I'm not going to merge this because it makes the code a bit more complex, because `\033[2m` is standard across all languages, and because it adds one unneeded dependency (which means it would complicate packaging for Linux distros).

kapsh:
Hello, 
Thanks for the feedback but I cannot agree with you. Let me explain.

1. Expressions like '\033[Xm' are NOT "standard", they are only Linux (and other *nix ofc.) TTY control codes. Colorama though is really cross-platform because it brings you Windows support for cheap. 

colorama_win_example

2. '\033[Xm' sequences are error-prone when typing and they are mixing with payload in strings.

3. Escape sequences are just hard to remember. Can you quick tell what is "\033[41m033msome text\033[0m"? I can't.

4. Packaging is not an issue - this library already has packages anywhere (just have checked Debian, Ubuntu, RHEL/CentOS, Archlinux, Gentoo, etc.).

PR #71, opened 2017-09-19
New rule key-ordering

### LINKED ISSUE #67: [Proposal] Key ordering rule
In order to enforce uniform key ordering in a mapping I want to have a rule for checking lexicographical sorting.

JFKnaufAncud:
closes #67


tests/rules/test_key_ordering.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 Adrien Vergé
```
adrienverge: # Copyright (C) 2017 Johannes F. Knauf
;-)

tests/rules/test_key_ordering.py
```diff
@@ -0,0 +1,67 @@
...
+                   'second: before_first\n'
+                   'at: root\n', conf,
+                   problem=(3, 1))
+        self.check('---\n'
+                   'nested but OK:\n'
+                   '  second: {first: 1}\n'
+                   '  third:\n'
+                   '    second: 2\n', conf)
+
+    def test_key_tokens_in_flow_sequences(self):
```
adrienverge: Is this needed?

JFKnaufAncud: TBH I don't know. I just included that potentially pathological case, because it was already present in your 'key-duplicates' tests. As it is valid YAML, and otherwise might easily be forgotten, I would keep it.

tests/rules/test_key_ordering.py
JFKnaufAncud: Updated the feature branch

adrienverge:
I pushed a commit to your branch to add more tests and update documentation.

Thanks for your contribution!

PR #77, opened 2017-10-16
Fix indentation rule for key following empty list

imtayadeway:
If a key-value pair follows an empty list, i.e.:

```yaml
a:
-
b: c
```

yamllint will complain:

```
warning  wrong indentation: expected 2 but found 0  (indentation)
```

This is because it is expecting the second key to be a continuation of
the block entry above:

```yaml
a:
-
  b: c
```

However, both are perfectly valid, though structurally different.


adrienverge:
Thanks for reporting this bug and fixing it!

I allowed myself to push a commit to your branch to add more tests cases.

imtayadeway:
@adrienverge thanks so much for the quick turnover and for adding extra test cases! :heart:

PR #81, opened 2017-10-27
Better color support check.

### LINKED ISSUE #79: Output is ugly on Windows
The console formatting stuff in the output should be excluded when on Windows. The output looks like this:

```
C:\Users\waylan\Desktop>yamllint sample.yml
←[4msample.yml←[0m
  ←[2m1:1←[0m       ←[33mwarning←[0m  missing document start "---"  ←[2m(document-start)←[0m
  ←[2m4:9←[0m       ←[31merror←[0m    wrong indentation: expected 10 but found 8  ←[2m(indentation)←[0m
```

... which is pretty hard to read.

Of course the `-f parsable` format works okay as a work-around, but that shouldn't be necessary.

```
C:\Users\waylan\Desktop>yamllint -f parsable sample.yml
sample.yml:1:1: [warning] missing document start "---" (document-start)
sample.yml:4:9: [error] wrong indentation: expected 10 but found 8 (indentation)
```

waylan:
Not all systems have `isatty` attribute on `sys.stdout` so check for
existance of attribute before checking value. Also don't use color in
Windows unless environ indicates support. Apparently, Windows can indicate
support by either the presence of `ANSICON` environ variable or if the
`TERM` environ variable is set to `ANSI`. Fixes #79.

No additional tests added, as the relevant tests use fcntl, which is a
Unix only lib. In fact, the tests won't even run in Windows (import error
for `fcntl` module in `test.cli.py`).

### DIFF
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -16,9 +16,9 @@
 
 from __future__ import print_function
 
-import os.path
+import os
 import sys
-
+import platform
 import argparse
 
 from yamllint import APP_DESCRIPTION, APP_NAME, APP_VERSION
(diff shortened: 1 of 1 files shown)


adrienverge:
Looks good, thanks!

Can you:
- rebase on latest master (I have pushed a [commit](https://github.com/adrienverge/yamllint/commit/d422274) to fix E722 errors)
- fix E501 errors in `yamllint/cli.py`?

waylan:
Requested changes made. Although it was sure easier to read with the longer lines. I switched all my projects to 119 char line length (width used by most code hosting site, including GitHub) some time ago and have never looked back.

PR #89, opened 2017-12-07
Feature: a new rule for checking octal numbers

ScienJus:
Hi, i want to add a new rule for checking the values should not be a octal number. like:

```
user-citycodes: 010
```

`010` is the city code of beijing. but in some system, it is converted to the decimal number `8`.

so if we add the config `octal-values: enable`. `user-citycodes: 010` will fail. but `user-citycodes: '010'` or `user-citycodes: 010,0571` will pass.

### DIFF
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -99,6 +99,11 @@ new-lines
 
 .. automodule:: yamllint.rules.new_lines
 
+octal-values
+------------
+
+.. automodule:: yamllint.rules.octal_values
+
 trailing-spaces
 ---------------
 
--- a/tests/rules/test_octal_values.py
+++ b/tests/rules/test_octal_values.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 Adrien Vergé
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from tests.common import RuleTestCase
+
+
+class OctalValuesTestCase(RuleTestCase):
+    rule_id = 'octal-values'
+
+    def test_disabled(self):
+        conf = ('octal-values: disable\n'
+                'new-line-at-end-of-file: disable\n'
+                'document-start: disable\n')
+        self.check('user-city: 010', conf)
+        self.check('user-city: 0o10', conf)
+
+    def test_implicit_octal_values(self):
+        conf = ('octal-values: {forbid-implicit-octal: true}\n'
+                'new-line-at-end-of-file: disable\n'
+                'document-start: disable\n')
+        self.check('user-city: 010', conf, problem=(1, 15))
+        self.check('user-city: abc', conf)
+        self.check('user-city: 010,0571', conf)
+        self.check("user-city: '010'", conf)
+        self.check('user-city: "010"', conf)
+        self.check('user-city:\n'
+                   '  - 010', conf, problem=(2, 8))
+        self.check('user-city: [010]', conf, problem=(1, 16))
+        self.check('user-city: {beijing: 010}', conf, problem=(1, 25))
+        self.check('explicit-octal: 0o10', conf)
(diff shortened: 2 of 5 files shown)


adrienverge:
Hi,

This looks like a nice new rule, thanks!

However I have a few remarks:

- Please keep alphabetical order in `rules.rst` and `__init__.py`.

- Please rebase on latest `master` to avoid many styling errors that I just fixed (sorry about that).

- Can you put your own name in copyright headers?

- The doc is a bit too brief. What do you think of this?

  ```
  Use this rule to prevent values with octal numbers. In YAML, numbers that
  start with ``0`` are interpreted as octal, but this is not always wanted.
  For instance ``010`` is the city code of Beijing, and should not be
  converted to ``8``.
  ```

- You need to add more tests! Both positive and negative tests. Examples:

  ```yaml
  a: 010
  b:
    - 010
  c: [010]
  d: {1: 010}
  ```

  ```yaml
  explicit octal: 0o10
  ```
  
  ```yaml
  zero: 0
  ```
  
  ```yaml
  hex value: 0x10
  ```

  ```yaml
  number values:
  - 0.10
  - .01
  - 0e3
  ```

- All rules should be referenced in `conf/default.yaml`, so could you add `octal-value: disable` in it? (please keep alphabetical order)

- `octal-values: enable` in config is a bit counter-intuitive: it looks like we *enable* (allow) octal values, but it's doing the opposite.

  What about this?
  ```yaml
  octal-values:
    forbid-implicit-octal: true  # for 010
    forbid-explicit-octal: true  # for 0o010
  ```

Feel free to ask if you need help!

ScienJus:
Hi @adamchainz . 

I have reworked on this pr, but testing failed in any non-python 2.6 env because of `The command "if [[ $TRAVIS_PYTHON_VERSION != 2.6 ]]; then flake8 .; fi" exited with 1.` .

And i have no idea why the test coverage is so much lower than the master branch.

adamchainz:
I think you meant @adrienverge

docs/rules.rst
```diff
@@ -107,4 +112,4 @@ trailing-spaces
 truthy
 ---------------
 
-.. automodule:: yamllint.rules.truthy
+.. automodule:: yamllint.rules.truthy
```
adrienverge: Please remove this change

tests/rules/test_octal_values.py
```diff
@@ -0,0 +1,75 @@
...
+        conf = ('octal-values: disable\n'
+                'empty-lines: disable\n'
+                'new-line-at-end-of-file: disable\n'
+                'document-start: disable\n')
+        self.check('user-city: 010', conf)
+        self.check('user-city: 0o10', conf)
+
+    def test_implicit_octal_values(self):
+        conf = ('octal-values: {forbid-implicit-octal: true}\n'
+                'empty-lines: disable\n'
```
adrienverge: same here

tests/rules/test_octal_values.py
```diff
@@ -0,0 +1,75 @@
...
+
+from tests.common import RuleTestCase
+
+
+class OctalValuesTestCase(RuleTestCase):
+    rule_id = 'octal-values'
+
+    def test_disabled(self):
+        conf = ('octal-values: disable\n'
+                'empty-lines: disable\n'
```
adrienverge: Can you remove this if this is not used?

yamllint/rules/octal_values.py
```diff
@@ -0,0 +1,93 @@
...
+        return
+
+    if conf['forbid-implicit-octal']:
+        if isinstance(token, yaml.tokens.ScalarToken):
+            if not token.style:
+                val = token.value
+                if val.isdigit() and len(val) > 1 and val[0] == '0':
+                    yield LintProblem(
+                        token.start_mark.line + 1, token.end_mark.column + 1,
+                        'implicit octal value "%s" in block mapping' %
```
adrienverge: It's not always in a block mapping (a block mapping is a multi-line dictionnary)!

You could write `forbidden implicit octal value "%s"`...

yamllint/rules/octal_values.py
```diff
@@ -0,0 +1,93 @@
...
+                        token.value)
+
+    if conf['forbid-explicit-octal']:
+        if isinstance(token, yaml.tokens.ScalarToken):
+            if not token.style:
+                val = token.value
+                if len(val) > 2 and val[:2] == '0o' and val[2:].isdigit():
+                    yield LintProblem(
+                        token.start_mark.line + 1, token.end_mark.column + 1,
+                        'explicit octal value "%s" in block mapping' %
```
adrienverge: `forbidden explicit octal value "%s"`?

ScienJus:
@adrienverge 

All changes have been completed, Do I need to manually squash commits?

adrienverge:
yes please!

ScienJus:
@adrienverge Done. Next time you can try rebasing and squashing commits by github's features

adrienverge:
Thanks for this feature!

PR #90, opened 2017-12-07
88 anchor references seen as duplicate key

### LINKED ISSUE #88: anchor references seen as duplicate key
I'm not sure if this is a feature or a bug! :-)

When running yamllint on a yaml file with multiple anchors referenced, you get a duplicate key error. 

EG:

Given the following file:
```yaml
---
anchor_one: &anchor_one
  sequence_one:
    - sequence_one
    - sequence_two

anchor_two: &anchor_two
  sequence_two:
    - sequence_three
    - sequence_four

main:
  <<: *anchor_one
  <<: *anchor_two
  unicorn_settings:
    - non-sequence-sequence
```

Upon running yamllint you get the following error:
```bash
test.yml
  14:3      error    duplication of key "<<" in mapping  (key-duplicates)
```

Python thinks it's "valid" and has the desired data structure:

```python
>>> import yaml, pprint
>>> pp = pprint.PrettyPrinter(indent=2)
>>> file = open("test.yml", "r")
>>> yaml_dump = yaml.load(file)
>>> pp.pprint(yaml_dump)
{ 'anchor_one': { 'sequence_one': ['sequence_one', 'sequence_two']},
  'anchor_two': { 'sequence_two': ['sequence_three', 'sequence_four']},
  'main': { 'sequence_one': ['sequence_one', 'sequence_two'],
            'sequence_two': ['sequence_three', 'sequence_four'],
            'unicorn_settings': ['non-sequence-sequence']}}
```

[As near as I can tell, it seems to be supported by the YAML spec:](http://www.yaml.org/spec/1.2/spec.html#id2765878):

*When composing a representation graph from serialized events, an alias node refers to the most recent node in the serialization having the specified anchor. Therefore, anchors need not be unique within a serialization.*

neekz0r:
fixes #88 

LMK if you spot anything fishy. I couldn't dedicate much time to tracing back so my method may be wrong --- but all unit tests pass so if there is something wrong, it's not covered.

### DIFF
--- a/tests/rules/test_key_duplicates.py
+++ b/tests/rules/test_key_duplicates.py
@@ -78,6 +78,15 @@ class KeyDuplicatesTestCase(RuleTestCase):
                    '    duplicates with\n'
                    '    many styles\n'
                    ': 1\n', conf)
+        self.check('---\n'
+                   'Merge Keys are OK:\n'
+                   'anchor_one: &anchor_one\n'
+                   '  one: one\n'
+                   'anchor_two: &anchor_two\n'
+                   '  two: two\n'
+                   'anchor_reference:\n'
+                   '  <<: *anchor_one\n'
+                   '  <<: *anchor_two\n', conf)
 
     def test_enabled(self):
         conf = 'key-duplicates: enable'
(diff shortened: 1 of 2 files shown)


adrienverge:
Hi @neekz0r, thanks a lot!

Do you think it would be possible to handle this case?

```yaml
duplicate keys:
  <<: {same key: value 1}
  <<: {same key: value 2}
```

keymon:
I would merge this PR directly. For the case described in  https://github.com/adrienverge/yamllint/pull/90#issuecomment-350237615, 

I think we should open a new issue and solve it in other PR. Currently we do not support this case anyway:

```
duplicate keys:
  <<: {same key: value 1}
  same key: value 2
```

neekz0r:
FYI, I'm still "here", but my time is limited to work on the expanded request (low priority for us at JLR, marked as a "nice to have").

So I would tend to agree with @keymon for selfish reasons.

adrienverge:
@keymon @neekz0r you're right.

I've just rebased on `master` and adapted the commits a bit.

Thanks for your contribution!

PR #104, opened 2018-04-02
removed leading slashes from 'ignore: |' directories

Eimert:
[Solved this issue](https://github.com/metacloud/molecule/issues/1228) and want to update the documentation. Occurs when yamllint is used with molecule.


adrienverge:
I admit this is clearer like this. Thanks!

PR #106, opened 2018-04-05
Clarify documentation on the 'truthy' rule

adamchainz:
I like the 'truthy' rule but its documentation and message have confused several of my colleagues (e.g. @flatnine and @seporaitis ). I've tried rewriting its docs and message to be clearer.

### DIFF
--- a/yamllint/rules/truthy.py
+++ b/yamllint/rules/truthy.py
@@ -15,11 +15,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """
-Use this rule to forbid truthy values that are not quoted nor explicitly typed.
+Use this rule to forbid non-explictly typed truthy values other than ``true``
+and ``false``, for example ``YES``, ``False`` and ``off``.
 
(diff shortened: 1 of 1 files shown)


adrienverge:
Thanks, it's always better when written by a real English speaker!

However I'd like to keep the mention to `on`/`off` and the reason why this rule is useful.
What do you think of this?


```diff
 Use this rule to forbid non-explictly typed truthy values other than ``true``
-and ``false``, for example ``YES`` and ``False``.
+and ``false``, for example ``YES``, ``False`` or ``off``.
+
+This can be useful to prevent YAML parsers from transforming ``[yes, FALSE,
+Off]`` into ``[true, false, false]`` or ``{y: 1, yes: 2, on: 3, true: 4,
+True: 5}`` into ``{y: 1, true: 5}``.
```

> Test failure seems to not be my fault...

Yes I know, I'm sorry about that... I fixed it temporarily but it came back, and I lack time to dig into it now.

adamchainz:
Updated. I reworded a little to point out the transformation can be surprising.

PR #110, opened 2018-04-06
Feature: Quoted value rule

jurajseffer:
We have a need to make sure all values are quoted, regardless of type as the processor cannot deal with numbers, booleans etc.

Questions:
- Is there a way to make the `quote-type` argument optional with `any` being the default? I could not see a way to do that.
- Why when I used an actual boolean in my test `truthy` rule interfered and I got `-  4:10: truthy value is not quoted (truthy),` as a failure for that test? See https://travis-ci.org/jurajseffer/yamllint/jobs/363218946

Thanks for the excellent library :)


yamllint/rules/quoted.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if isinstance(prev, yaml.tokens.ValueToken):
+            quote_type = conf['quote-type']
+            if ((quote_type == 'single' and token.style != "'") or
+                    (quote_type == 'double' and token.style != '"') or
+                    (quote_type == 'any' and token.style is None)):
+                yield LintProblem(
+                    token.start_mark.line + 1,
+                    token.start_mark.column + 1,
+                    "value is not quoted (using %s quote)" % quote_type
```
adrienverge: `value` -> `string value`?

Also please don't use parentheses in the error message, it interfers with parsers and would result in `(using " quote) (quoted)` which is unclear... In my opinion, `string value is not quoted` is perfect!

yamllint/rules/quoted.py
```diff
@@ -0,0 +1,62 @@
...
+   ::
+
+    foo: "bar"
+    bar: 'foo'
+
+   the following code snippet would **FAIL**:
+   ::
+
+    foo: bar
+    bar: 123
```
adrienverge: Adding `123` in the *failure* example is a bit confusing... I'd rather have documentation as clear as possible, for instance:

**PASS**:
```yaml
foo: "bar"
bar: 'foo'
number: 123
boolean: true
```

**FAIL**:
```yaml
foo: bar
```

PR #114, opened 2018-05-23
Added configuration for the syntax checker

aohorodnyk:
Hello,

I've tried to use yamlliner for my needs, but I can't.
Yaml files in my projects are contain some placeholders, as the result yamlliner was crashed with syntax error.

I decided to configure syntax checker level or disable it at all, but didn't find any configuration options for that.

So I've made changes and want to contribute it to official yamlliner.

I've added optional `syntax-checker` configuration with optional sub-options: `enabled` and `level`, example for this configuration:
```yaml
syntax-checker:
    enabled: false
    level: warning
rules:
    ...
```
Default configuration options (if you don't send some of there or all options) are:
```yaml
syntax-checker:
    enabled: true
    level: error
rules:
    ...
```

### DIFF
--- a/yamllint/config.py
+++ b/yamllint/config.py
@@ -31,7 +31,7 @@ class YamlLintConfig(object):
         assert (content is None) ^ (file is None)
 
         self.ignore = None
-        self.syntax_checker = None
+        self.syntax_checker = {'enabled': True, 'level': 'error'}
 
         if file is not None:
             with open(file) as f:
@@ -109,8 +109,6 @@ class YamlLintConfig(object):
                 raise YamlLintConfigError('invalid config: syntax-checker.level should be "error" or "warning"')
 
             self.syntax_checker = syntax_checker
-        else:
-            self.syntax_checker = {'enabled': True, 'level': 'error'}
 
     def validate(self):
         for id in self.rules:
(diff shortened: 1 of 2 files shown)


aohorodnyk:
I really want to deliver this feature to the YamlLint checker. Could you write me what do you think about that?
I'll fix tests if all right for you.
Thanks

adrienverge:
Hi,

I'm not sure this is a good idea, because most yamllint rules need a valid YAML syntax to run.
If you get a working prototype, that doesn't break other rules, I'd be OK to review your PR. In any case this feature would need many tests to make sure it behaves correctly in all corner cases.

By the way, maybe using the same config format as for other rules would be more intuitive, what do you think?
```yaml
rules:
  syntax: disable
```
```yaml
rules:
  syntax:
    level: error
```

aohorodnyk:
Hi,

Sorry for the late response, I thought about additional features for resolving the issue with full disable of syntax parser.

I added the config to the root instead of rules block, because of code architecture. Following YamlLint architecture, Syntax Parser isn't part of rules and a part of the core functionality.

Why do I want to disable syntax checker? YAML parser from the Symfony framework has some breaks of the YAML syntax, but we want to check some things. So I added an ability to disable syntax error output.

Also, I want to add regexp expressions for syntax checking functionality, so we could disable part of syntax errors. What do you think about that?

aohorodnyk:
As I see, actually this PR does not relevant for me :)
It's strange, but I guess we could close it.
Thanks for the support.

PR #117, opened 2018-05-29
EPEL on CentOS

sixdrum:
I noticed that yamllint isn't in the base repo for CentOS 7, so I added a note about requiring EPEL.


adrienverge:
I squashed and slightly updated your commits.

Thanks!

Issue #107, opened 2018-04-05
Override config via environment


jbergstroem:
When playing with dockerfiles and multiple repos (main usecase for me is part of CI) it would be nice to be able to override specific config settings via environment; for instance `YAMLLINT_CONFIG="line-length.max=81" yamllint foo.yaml`. This makes it easier to override a few settings without having to check in a config in each repo you're testing against.

adrienverge:
I agree it could be cool. But I'd rather have a YAML-formatted config, like for the `-d` option.

By the way, did you know you can already use this?
```shell
YAMLLINT_CONFIG="{extends: relaxed, rules: {line-length: {max: 81}}}"
yamllint -d "$YAMLLINT_CONFIG" foo.yaml
```

jbergstroem:
> @adrienverge said: did you know you can already pass config as env?

Didn't. Thanks!

> @adrienverge said: `YAMLLINT_CONFIG=..`

I considered format as well; suggested something different than YAML in my original comment because I felt a simpler dot notation was "closer" to your average bash experience. I guess it doesn't scale as well, but the idea was to have the option of changing "just the one thing"; not really passing a full config.

I guess there are two minor opportunities for improvements here. Consider a use case similar to:
```dockerfile

# Override while calling this docker instance in your CI pipeline
ENV YAML_ARGS=""

# ...

CMD ["yamllint", "-d '${YAML)ARGS}'"]

```
..and overriding this from your ci script.

1.  Allowing "empty" configs:
```console
$ docker run --rm -v $(PWD):/yaml sdesbure/yamllint yamllint -d "" .gitlab-ci.yml
invalid config: not a dict
```

2. Avoiding `-d` by reading a defined environment variable

I guess both can be handled in wrappers but I thought it be worth bringing it up. Feel free to close this issue unless you feel we should make either improvement actionable.

adrienverge:
Hey,

Both 1. and 2. make sense!
Contributions are welcome (they need to come with tests).

grzesuav:
@adrienverge When I have config in file and want just to override one rule, can I do it with existing mechanism somehow ?

adrienverge:
@grzesuav you can do this:
```shell
YAMLLINT_CONFIG="{extends: /my/config/file.yaml, rules: {line-length: {max: 20}}}"
yamllint -d "$YAMLLINT_CONFIG" foo.yaml
```

grzesuav:
ok, thanks :) 

[...]

PR #119, opened 2018-06-28
cli: read optional config from YAMLLINT_CONFIG

jbergstroem:
This is a "minor" abi break (?) since empty configs are disregarded instead of failing.

Fixes: https://github.com/adrienverge/yamllint/issues/107

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -165,7 +165,19 @@ class CommandLineTestCase(unittest.TestCase):
 
         out, err = sys.stdout.getvalue(), sys.stderr.getvalue()
         self.assertEqual(out, '')
-        self.assertRegexpMatches(err, r'^invalid config: not a dict')
+        self.assertRegexpMatches(err, r'No such file or directory')
+
+    def test_run_with_environment_variable(self):
+        sys.stdout, sys.stderr = StringIO(), StringIO()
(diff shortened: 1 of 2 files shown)


jbergstroem:
So, my minor logic change now left us with a few uncovered branches. Suggestions on how to tackle this? Add tests? Different route for handling the parsing of environment variables?

jbergstroem:
@adrienverge any input on above?

adrienverge:
Hi Johan, thanks, this looks good.

The API break is not problematic in my opinion, as empty config data resulted in an error.

Could you explain your comment on the *few uncovered branches*? I didn't get it.

About tests:
- `test_run_with_empty_config()` should be changed to run without error (on an existing file), as this is the new behaviour.
- Please add a case for checking that `-d` option overrides the environment.

adrienverge:
Also, using `-d ''` should be a way to cancel the use of `$YAMLLINT_CONFIG`, what do you think? We would need a test for that.

adrienverge:
@jbergstroem if you want to continue on this pull request, using `YAMLLINT_CONFIG_DATA` instead of `YAMLLINT_CONFIG` would be consistent with the new feature from https://github.com/adrienverge/yamllint/pull/255#pullrequestreview-400674082.

PR #130, opened 2018-09-27
Feature: Quoted strings value rule

mgmgwi:
* taken from https://github.com/adrienverge/yamllint/pull/110 (submitted by @jurajseffer)
* small fix for generic strings
* fixes for comments from @adrienverge

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ __pycache__
 /docs/_build
 /dist
 /yamllint.egg-info
+/.vscode/
+/build/
--- a/tests/rules/test_quoted_strings.py
+++ b/tests/rules/test_quoted_strings.py
@@ -1,0 +1,125 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2018 ClearScore
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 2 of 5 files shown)


mgmgwi:
I don't know how to fix the `The command "pip install pyyaml flake8 flake8-import-order coveralls" failed and exited with 1 during` problem from the python 2.6 build.

mgmgwi:
@adrienverge: Any feedback would be very much appreciated.

adrienverge:
Hi @mgmgwi,

Thanks for all that good work.

I pushed a commit to your branch, with changes I wanted to see:
- list this rule in `default.yaml`,
- clarify the doc a bit,
- add a few more test cases.

In these new test cases, the rule currently fails on multi-line strings.
I think it's important that *explicit* multi-line strings (with `|` or `>`) pass the test, without a need to add quotes:
```yaml
multiline string 1: |
  line 1
  line 2
multiline string 2: >
  word 1
  word 2
```

What do you think? Maybe we could use other properties of `ScalarToken` (`plain`, `style`, `start_mark`, `end_mark`) to detect multiline `>` and `|` cases?

@jurajseffer any thoughts?

mgmgwi:
@adrienverge Thanks for your feedback. 

> I think it's important that explicit multi-line strings (with | or >) pass the test, without a need to add quotes.

I agree with you on that.


> What do you think? Maybe we could use other properties of `ScalarToken` (`plain`, `style`, `start_mark`, `end_mark`) to detect multiline `>` and `|` cases?

I added a patch which ignores multi-line strings. Do you think that's ok?

adrienverge:
> I don't know how to fix the `The command "pip install pyyaml flake8 flake8-import-order coveralls" failed and exited with 1 during` problem from the python 2.6 build.

Sorry about that, `coveralls` no longer supports Python 2.6 so builds started to fail recently. This is fixed on `master` now.

> Do you think that's ok?

It looks great! I just added an extra commit with another type of multi-line string (and it works as expected).
If this OK for you, can you squash all commits, rebase on `master` and push again? Tests should pass then.

mgmgwi:
> If this OK for you, can you squash all commits, rebase on master and push again? Tests should pass then.

@adrienverge: Thanks again for your feedback. Checks are looking good now.

adrienverge:
Thank you and @jurajseffer for contributing!

I'll make a new release that includes this new rule.

PR #139, opened 2018-10-18
Add -f colored to force colors

sedrubal:
Sometimes you want the output to be colored even if sys.stdout is not a tty. Example:

```bash
yamllint . | less -R
```

Therefore I added the argument `-f colored` or `--format colored` to force a colored output.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -351,7 +351,7 @@ class CommandLineTestCase(unittest.TestCase):
             '\n' % file))
         self.assertEqual(err, '')
 
-    def test_run_colored_output(self):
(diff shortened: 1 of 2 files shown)


adrienverge:
Nice idea!

However for symmetry it would be nice if `standard` would be fixed, forced too.

What about having these options:
- `parsable`
- `standard`  ← non-colored
- `colored`  ← colored
- `auto`   ← chooses `standard` or `colored` depending on terminal capabilities, like `standard` today, should be the new default for backwards-compatibility

?

Also, you need to add tests for this kind of modifications.

sedrubal:
yes, that sounds reasonable. I'll adapt the PR.

PR #149, opened 2018-11-24
Drop support for EOL Python 2.6

hugovk:
Summary: yamllint used to support Python 2.6 because it's a dependency of Ansible, which supported Python 2.6. Ansible has now dropped 2.6, so yamllint may also drop it.

---

Python 2.6 is EOL since 2013-10-29 and no longer receiving security updates (or any updates) from the core Python team.

Source: https://en.wikipedia.org/wiki/CPython#Version_history

Ansible 2.7.0 was released in October 2018 which drops Python 2.6.

* https://deleforterie.com/wordpress/index.php/2018/10/21/ansible-python-2-6-support-ends-with-ansible-2-7/
* https://github.com/ansible/ansible/releases/tag/v2.7.0
* https://github.com/ansible/ansible/pull/42971

> * Support for running an Ansible controller with Python-2.6 has been dropped. You can still manage machines which use Python-2.6 but you will have to manage them from a machine which has Python-2.7 or Python-3.5 or greater installed. See the porting guide if you need more information.

https://github.com/ansible/ansible/blob/stable-2.7/changelogs/CHANGELOG-v2.7.rst#v2-7-0

> The decision to drop Python-2.6 support on the controller was made because many dependent libraries are becoming unavailable there. In particular, python-cryptography is no longer available for Python-2.6 and the last release of pycrypto (the alternative to python-cryptography) has known security bugs which will never be fixed.

https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.7.html#python-compatibility

Yamllint's support of Python 2.6 is also preventing its dependencies from dropping it too, for example pathspec: https://github.com/cpburnz/python-path-specification/issues/16.

---

This PR drops support for 2.6.

It adds `python_requires` to help pip install the right version for people still running old Python versions, plus Trove version classifiers for PyPI. 

It also upgrades unit tests to use more useful asserts, and upgrades to modern Python syntax using `pyupgrade`.


adrienverge:
> Ansible has now dropped 2.6, so yamllint may also drop it.

yamllint is not only used by Ansible. What makes you say that Python 2.6 support can be dropped like that, without impacting any other project?

adrienverge:
Thanks a lot for this comprehensive review of dependent Python projects. You convinced me.

I'll do a slight reword of your last commit then merge.

hugovk:
You're welcome, and thanks for your work on yamllint!

PR #167, opened 2019-03-21
Add FreeBSD installation instructions

0mp:


adrienverge:
Thanks!

Please see the error on Travis, it looks like your change brings unwanted trailing chars.

One question: isn't there any way to package a simple `yamllint` package? (that would be an *alias* for `py36-yamllint` for instance)

0mp:
Ok, I've fixed the whitespace ;)

> One question: isn't there any way to package a simple `yamllint` package? (that would be an _alias_ for `py36-yamllint` for instance)

Well, not really. I'll have to ask the community about the details because I cannot remember exactly why it is not offered.

adrienverge:
OK. Please let us know if it can be changed to `yamllint` :+1:

0mp:
OK :smile:

vishwin:
https://wiki.freebsd.org/Python/PortsPolicy

tl;dr: no.

Although our Python ports policy may say CFP (because we're not the best at updating wikis), this is in fact enforced as written. There exists no compelling reason to omit the `py-` prefix, especially since this package exists in PyPi.

adrienverge:
@vishwin thanks for your feedback. But what does "CFP" mean?

vishwin:
Call for Proposal

PR #180, opened 2019-05-27
Sort import orders

xatier:
Sort import orders reported by [flake8-import-order](https://github.com/PyCQA/flake8-import-order).


xatier:
That was quick!

Issue #150, opened 2018-11-28
Feature suggestion: allow other truthy convention


roperto:
I'd like to use yes/no instead of true/false as this seem to be used more often in Ansible.

I can try to make a PR if the maintainer thinks it could be a nice addition and can review the code.

Any thoughs?

adrienverge:
This sounds reasonable.

Before writing any code, how would you think the configuration should look like? Currently it's just `truthy: disable` or `truthy: enable` or `truthy: {}`.

roperto:
What about a list of keywords that are valid?

Ie `['yes', 'no']` ?

That should not affect the currently used options.

adrienverge:
More precisely, what keyword would you use? `allowed-values`?
```yaml
truthy:
  allowed-values: [yes, no]
```
or

```yaml
truthy:
  allowed-values: [on, off]
```
or

```yaml
truthy: enable  # that defaults to [true, false]
```
?

roperto:
Sounds like a plan.

We probably need to quote the values as strings in the configuration do they don't evaluate to booleans, is that about right?

adrienverge:
You're totally right!

samdoran:
I would love to see this added. This would be very useful for controlling valid booleans values rather than just allowing all the flexible values or _only_ `true/false`.

evitalis:
I would also like to vote on being able to use flexible values and match what ansible allows to be valid. It won't hurt anything already using true/false but allow ansible playbooks to be linted with `yamllint` as well as `ansible-lint`.

I could see this not being changed though as YAML is not dictated by ansible itself but it does seem that many people use both linters for their playbooks.

Lirt:
I like key `allowed-values` as suggested by @adrienverge.

Having this option would make life of all `ansible` folks much easier. Did someone start to work on this?

evitalis:
@adrienverge Any chance of this being resolved resolved soon? It's making double checking ansible playbooks very noisy. Yes `ansible-lint` is tailored more for that project but a second review is not a bad thing.

adrienverge:
@evitalis this will be resolved soon, if someone starts working on it 🙂 Do you volunteer?

Lirt:
> @evitalis this will be resolved soon, if someone starts working on it slightly_smiling_face Do you volunteer?

I'm in, I will try to create MR today.

Lirt:
Please review https://github.com/adrienverge/yamllint/pull/181. 

[...]

PR #181, opened 2019-06-02
Enh/allowed truthy values/#150

Lirt:
This MR implements optional exceptions for truthy rule. 

Example usage (ansible):

```yaml
---
rules:
  line-length:
    max: 160
    level: warning
  truthy:
    allowed-values: ["yes", "no"]
```

Implemented in truthy rule by doing difference of sets of predefined banned values and configured allowed values.

Added 3 tests, all passed successfully.

Please feel free to review and suggest simplifications, or better variable names, which can be more standard and explanative.

Closes https://github.com/adrienverge/yamllint/issues/150

### DIFF
--- a/tests/rules/test_truthy.py
+++ b/tests/rules/test_truthy.py
@@ -49,6 +49,68 @@ class TruthyTestCase(RuleTestCase):
                    problem3=(7, 3), problem4=(7, 7),
                    problem5=(8, 3), problem6=(8, 7))
 
+    def test_different_allowed_values(self):
+        conf = ('truthy:\n'
+                '  allowed-values: ["yes", "no"]\n')
+        self.check('---\n'
+                   'key1: foo\n'
+                   'key2: yes\n'
+                   'key3: bar\n'
+                   'key4: no\n', conf)
+        self.check('---\n'
+                   'key1: true\n'
+                   'key2: Yes\n'
+                   'key3: false\n'
+                   'key4: no\n'
+                   'key5: yes\n',
+                   conf,
+                   problem1=(2, 7), problem2=(3, 7),
+                   problem3=(4, 7))
+
+    def test_combined_allowed_values(self):
+        conf = ('truthy:\n'
+                '  allowed-values: ["yes", "no", "true", "false"]\n')
+        self.check('---\n'
+                   'key1: foo\n'
+                   'key2: yes\n'
+                   'key3: bar\n'
+                   'key4: no\n', conf)
+        self.check('---\n'
+                   'key1: true\n'
+                   'key2: Yes\n'
+                   'key3: false\n'
+                   'key4: no\n'
+                   'key5: yes\n',
+                   conf, problem1=(3, 7))
+
+    def test_no_allowed_values(self):
+        conf = ('truthy:\n'
+                '  allowed-values: []\n')
+        self.check('---\n'
+                   'key1: foo\n'
+                   'key2: bar\n', conf)
+        self.check('---\n'
+                   'key1: true\n'
+                   'key2: yes\n'
+                   'key3: false\n'
+                   'key4: no\n', conf,
+                   problem1=(2, 7), problem2=(3, 7),
+                   problem3=(4, 7), problem4=(5, 7))
+
+    def test_empty_string_allowed_values(self):
+        conf = ('truthy:\n'
+                '  allowed-values: ["", ""]\n')
+        self.check('---\n'
+                   'key1: foo\n'
+                   'key2: bar\n', conf)
+        self.check('---\n'
+                   'key1: true\n'
+                   'key2: yes\n'
+                   'key3: false\n'
+                   'key4: no\n', conf,
+                   problem1=(2, 7), problem2=(3, 7),
+                   problem3=(4, 7), problem4=(5, 7))
+
     def test_explicit_types(self):
         conf = 'truthy: enable\n'
         self.check('---\n'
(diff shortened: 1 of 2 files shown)


Lirt:
Force-pushed 2 fixes for pep8 linting.

Lirt:
Updated documentation.

yamllint/rules/truthy.py
```diff
@@ -63,6 +71,23 @@
...
+
+#. With ``truthy: {allowed-values: ["yes", "no"]}``
+
+   the following code snippet would **PASS**:
+   ::
+
+    key1: yes
+    key2: no
+    key3: true
+    key4: false
```
adrienverge: `true` and `false` should not be allowed with `allowed-values: ["yes", "no"]`

Lirt:
Ah, I thought about this feature a little differently. My intention was, that allowed values will just add values, that will be ignored during checking, but true/false will stay used. My bad, I will rewrite it in a moment.

I didn't squash it, because it can be squashed by GitHub merge option. Is there any advantage in me squashing this manually?

yamllint/rules/truthy.py
```diff
@@ -84,8 +111,14 @@ def check(conf, token, prev, next, nextnext, context):
...
+        banned_values -= set(conf['allowed-values'])
+
     if isinstance(token, yaml.tokens.ScalarToken):
-        if token.value in TRUTHY and token.style is None:
+        if token.value in banned_values and token.style is None:
             yield LintProblem(token.start_mark.line + 1,
                               token.start_mark.column + 1,
-                              "truthy value should be true or false")
+                              "truthy value should be true, false or one of "
+                              "values configured in 'truthy: allowed-values'")
```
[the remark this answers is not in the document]
Lirt: Currently looks like:
```
truthy value should be one of [false, true]
truthy value should be one of [no, yes]
```

adrienverge:
> Ah, I thought about this feature a little differently. My intention was, that allowed values will just add values, that will be ignored during checking, but true/false will stay used. My bad, I will rewrite it in a moment.

Sure, I understand. But if one day a user asks for an option to disallow `true`/`false`, we won't be able to do it without breaking retrocompatibility. Let's take a leap forward 😉

> I didn't squash it, because it can be squashed by GitHub merge option. Is there any advantage in me squashing this manually?

I prefer having a commit ready to merge, so the commit message is fully written and ready. This commit message is important for browsing history, using `git blame`, `git bisect` etc.

yamllint/rules/truthy.py
```diff
@@ -84,8 +111,14 @@ def check(conf, token, prev, next, nextnext, context):
     if prev and isinstance(prev, yaml.tokens.TagToken):
         return
 
+    # Allow exceptions for truthy values
+    banned_values = set(TRUTHY)
+    if "allowed-values" in conf:
+        banned_values -= set(conf['allowed-values'])
```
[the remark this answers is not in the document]
Lirt: I was testing it without the `DEFAULT`, that's why I had the condition, but I removed it now. It is really not needed. 

I dropped this code and updated the token comparison.

yamllint/rules/truthy.py
Lirt: Updated the documentation example.

Lirt:
Yes it will be better. I squashed commits. Please recheck the reviews.

tests/rules/test_truthy.py
```diff
@@ -49,6 +49,68 @@ def test_enabled(self):
...
+                   'key4: no\n', conf)
+        self.check('---\n'
+                   'key1: true\n'
+                   'key2: Yes\n'
+                   'key3: false\n'
+                   'key4: no\n'
+                   'key5: yes\n',
+                   conf,
+                   problem1=(2, 7), problem2=(3, 7),
+                   problem3=(4, 7))
```
adrienverge: :+1:

tests/rules/test_truthy.py
```diff
@@ -49,6 +49,68 @@ def test_enabled(self):
...
+                   'key2: yes\n'
+                   'key3: bar\n'
+                   'key4: no\n', conf)
+        self.check('---\n'
+                   'key1: true\n'
+                   'key2: Yes\n'
+                   'key3: false\n'
+                   'key4: no\n'
+                   'key5: yes\n',
+                   conf, problem1=(3, 7))
```
adrienverge: :+1:

tests/rules/test_truthy.py
```diff
@@ -49,6 +49,68 @@ def test_enabled(self):
...
+                   'key1: true\n'
+                   'key2: yes\n'
+                   'key3: false\n'
+                   'key4: no\n', conf,
+                   problem1=(2, 7), problem2=(3, 7),
+                   problem3=(4, 7), problem4=(5, 7))
+
+    def test_empty_string_allowed_values(self):
+        conf = ('truthy:\n'
+                '  allowed-values: ["", ""]\n')
```
adrienverge: :+1:

yamllint/rules/truthy.py
```diff
@@ -63,6 +71,25 @@
...
+    key4: bar
+
+   the following code snippet would **FAIL**:
+   ::
+
+    key1: no
+    key2: true
+    key3: false
+    key4: yes
+    key5: foo
```
adrienverge: For clarity, I propose:
```
   the following code snippet would **PASS**:
   ::

    - yes
    - no
    - "true"
    - 'false'
    - foo
    - bar

   ::

    - true
    - false
    - on
    - off
```
What do you think?

tests/rules/test_truthy.py
Lirt: Should I update something here?

yamllint/rules/truthy.py
Lirt: Yes, it is easier to understand.

tests/rules/test_truthy.py
adrienverge: No, it's perfect.

adrienverge:
Thanks @Lirt!

Lirt:
@adrienverge Thank you for accepting and review!

PR #205, opened 2019-10-10
fix(default.yaml): disable `empty-values` & `octal-values` by default

### LINKED ISSUE #204: [BUG] Default setting (`enable`) of `empty-values` & `octal-values` does nothing
There appears to be a contradiction between the meaning of `enable` in the default settings and the actual end result.

#### `empty-values`

https://github.com/adrienverge/yamllint/blob/b62b424dd43acd4b0e08f46f49d88bc15461384e/yamllint/conf/default.yaml#L21

https://github.com/adrienverge/yamllint/blob/0f073f7a09d29ceb85a1aaf388bdbe5f4145f869/yamllint/rules/empty_values.py#L78-L79

#### `octal-values`

https://github.com/adrienverge/yamllint/blob/b62b424dd43acd4b0e08f46f49d88bc15461384e/yamllint/conf/default.yaml#L29

https://github.com/adrienverge/yamllint/blob/0f073f7a09d29ceb85a1aaf388bdbe5f4145f869/yamllint/rules/octal_values.py#L69-L70

---

In both cases, the only way to get these working is to specify them in the `.yamllint` file:

```yaml
rules:
  empty-values:
    forbid-in-block-mappings: true
    forbid-in-flow-mappings: true
  octal-values:
    forbid-implicit-octal: true
    forbid-explicit-octal: true
```

What is `enable` supposed to do?  Otherwise, isn't the default actually `disable` in both cases, with `default.yaml` needing updating (with the knock-on affect to the documentation)?

myii:
* Close #204 

---

@adrienverge This is the crux of the resolution.  What do you think about adding a note in the documentation, under each section (`empty-values` & `octal-values`)?  It could be used for the time being until existing users become informed that these values were in fact, not enabled.

### DIFF
--- a/tests/rules/test_octal_values.py
+++ b/tests/rules/test_octal_values.py
@@ -28,7 +28,9 @@ class OctalValuesTestCase(RuleTestCase):
         self.check('user-city: 0o10', conf)
 
     def test_implicit_octal_values(self):
-        conf = ('octal-values: {forbid-implicit-octal: true}\n'
+        conf = ('octal-values:\n'
+                '  forbid-implicit-octal: true\n'
+                '  forbid-explicit-octal: false\n'
                 'new-line-at-end-of-file: disable\n'
                 'document-start: disable\n')
         self.check('user-city: 010', conf, problem=(1, 15))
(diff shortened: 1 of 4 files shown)


myii:
You're welcome, thanks for the idea!

PR #210, opened 2019-11-14
Allow Disabling Non-Yaml Files

kadaan:
Allow disabling of a file, even if it is invalid yaml by including `yamllint disable-file` in the first line.

### DIFF
--- a/docs/disable_with_comments.rst
+++ b/docs/disable_with_comments.rst
@@ -73,3 +73,31 @@ It is possible, although not recommend, to disabled **all** rules:
 
 If you need to disable multiple rules, it is allowed to chain rules like this:
 ``# yamllint disable rule:hyphens rule:commas rule:indentation``.
+
+Disabling all checks for a file
+-------------------------------
+
+To prevent yamllint from reporting problems for a specific file, add the
+directive comment ``# yamllint disable-file`` as the first line of the file.
+For instance:
+
+.. code-block:: yaml
+
+ # yamllint disable-file
+ # The following mapping contains the same key twice, but I know what I'm doing:
+ key: value 1
+ key: value 2
+
+ - This line is waaaaaaaaaay too long but yamllint will not report anything about it.
+   This line will be checked by yamllint.
+
+or:
+
+.. code-block:: jinja
+
+ # yamllint disable-file
+ # This file is not valid YAML because it is a Jinja template
+ {% if extra_info %}
+ key1: value1
+ {% endif %}
+ key2: value2
--- a/tests/test_yamllint_directives.py
+++ b/tests/test_yamllint_directives.py
@@ -302,3 +302,104 @@ class YamllintDirectivesTestCase(RuleTestCase):
                    '    c: [x]\n',
                    conf,
                    problem=(6, 2, 'comments-indentation'))
+
+    def test_disable_file_directive(self):
+        conf = ('comments: {min-spaces-from-content: 2}\n'
+                'comments-indentation: {}\n')
+        self.check('# yamllint disable-file\n'
+                   '---\n'
+                   '- a: 1 # comment too close\n'
+                   '  b:\n'
+                   ' # wrong indentation\n'
+                   '    c: [x]\n',
+                   conf)
+        self.check('#    yamllint disable-file\n'
+                   '---\n'
+                   '- a: 1 # comment too close\n'
+                   '  b:\n'
+                   ' # wrong indentation\n'
+                   '    c: [x]\n',
+                   conf)
+        self.check('#yamllint disable-file\n'
+                   '---\n'
+                   '- a: 1 # comment too close\n'
+                   '  b:\n'
+                   ' # wrong indentation\n'
+                   '    c: [x]\n',
+                   conf)
+        self.check('#yamllint disable-file    \n'
+                   '---\n'
+                   '- a: 1 # comment too close\n'
+                   '  b:\n'
+                   ' # wrong indentation\n'
+                   '    c: [x]\n',
+                   conf)
+        self.check('---\n'
+                   '# yamllint disable-file\n'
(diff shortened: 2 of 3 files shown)


adrienverge:
Hi @kadaan, this is an interesting feature!

Some remarks:

- Checking the presence of the directive `yamllint disable-file` on *all* lines of each file (i.e. using `re.match(re.M)`) might cause a performance decrease.
  Your commit message says "in the first line", I propose to actually do that, what do you think?

- Could we check `^# yamllint disable-file` instead of `^.*\s+yamllint disable-file`, like for other directives (`yamllint disable` and `yamllint disable-line`)? Do you have a real use case with `//`?

- You need to update documentation too, please add a thrid section inside `docs/disable_with_comments.rst`.

- About tests:
  - Can you please place them in `tests/test_yamllint_directives.py`, where all other tests for `disable-line` are?
  - We need more tests, for example check correct behavior when `yamllint disable-file` is found elsewhere than on first line; or with similar strings like `aaayamllint disable-filezzz`; or with similar strings like `yamllint disable-file rule:cannot-do-this-but-users-might-try`... Can you think of other strange use-cases?

kadaan:
Hi @adrienverge.  

1.  Totally makes sense
2.  I had it that way initially, but then thought that maybe supporting any "comment" would be nice in case the file was not actually yaml and the comment char was different.  Might be better to start with the simple case and extend as needed
3. Good point
4. Sure

kadaan:
@adrienverge I think this should be ready to go.  What do you think?

adrienverge:
This is neat. @kadaan, thanks for your contribution!

I just allowed myself to add a more tests and some minor edits:
```diff
--- a/docs/disable_with_comments.rst
+++ b/docs/disable_with_comments.rst
@@ -43,34 +43,6 @@ specific line:
 
 Disabling all checks for a file
 -------------------------------
-To prevent yamllint from reporting problems for a specific file, add a
-directive comment (``# yamllint disable-file``) as the first line of the file.
+
+To prevent yamllint from reporting problems for a specific file, add the
+directive comment ``# yamllint disable-file`` as the first line of the file.
 For instance:
 
 .. code-block:: yaml
 
  # yamllint disable-file
- # The following mapping contains the same key twice,
- # but I know what I'm doing:
+ # The following mapping contains the same key twice, but I know what I'm doing:
  key: value 1
  key: value 2
 
@@ -20,9 +20,8 @@ or:
 .. code-block:: jinja
 
  # yamllint disable-file
- # This file is not valid yaml because it is a Jinja template
+ # This file is not valid YAML because it is a Jinja template
  {% if extra_info %}
  key1: value1
  {% endif %}
  key2: value2
diff --git a/tests/test_yamllint_directives.py b/tests/test_yamllint_directives.py
index 19b5af8..17bb69e 100644
--- a/tests/test_yamllint_directives.py
+++ b/tests/test_yamllint_directives.py
@@ -303,7 +303,7 @@ class YamllintDirectivesTestCase(RuleTestCase):
                    conf,
                    problem=(6, 2, 'comments-indentation'))
 
-    def test_disabled_file(self):
+    def test_disable_file_directive(self):
         conf = ('comments: {min-spaces-from-content: 2}\n'
                 'comments-indentation: {}\n')
         self.check('# yamllint disable-file\n'
@@ -362,3 +362,44 @@ class YamllintDirectivesTestCase(RuleTestCase):
                    problem1=(1, 1, 'document-start'),
                    problem2=(3, 8, 'comments'),
                    problem3=(5, 2, 'comments-indentation'))
+
+    def test_disable_file_directive_not_at_first_position(self):
+        self.check('# yamllint disable-file\n'
+                   '---\n'
+                   '- bad  : colon and spaces   \n',
+                   self.conf)
+        self.check('---\n'
+                   '# yamllint disable-file\n'
+                   '- bad  : colon and spaces   \n',
+                   self.conf,
+                   problem1=(3, 7, 'colons'),
+                   problem2=(3, 26, 'trailing-spaces'))
+
+    def test_disable_file_directive_with_syntax_error(self):
+        self.check('# This file is not valid YAML (it is a Jinja template)\n'
+                   '{% if extra_info %}\n'
+                   'key1: value1\n'
+                   '{% endif %}\n'
+                   'key2: value2\n',
+                   self.conf,
+                   problem=(2, 2, 'syntax'))
+        self.check('# yamllint disable-file\n'
+                   '# This file is not valid YAML (it is a Jinja template)\n'
+                   '{% if extra_info %}\n'
+                   'key1: value1\n'
+                   '{% endif %}\n'
+                   'key2: value2\n',
+                   self.conf)
+
+    def test_disable_file_directive_with_dos_lines(self):
+        self.check('# yamllint disable-file\r\n'
+                   '---\r\n'
+                   '- bad  : colon and spaces   \r\n',
+                   self.conf)
+        self.check('# yamllint disable-file\r\n'
+                   '# This file is not valid YAML (it is a Jinja template)\r\n'
+                   '{% if extra_info %}\r\n'
+                   'key1: value1\r\n'
+                   '{% endif %}\r\n'
+                   'key2: value2\r\n',
+                   self.conf)
--- a/yamllint/linter.py
+++ b/yamllint/linter.py
@@ -20,6 +20,7 @@ import yaml
 
 from yamllint import parser
 
+
 PROBLEM_LEVELS = {
     0: None,
     1: 'warning',
@@ -189,7 +190,7 @@ def _run(buffer, conf, filepath):
         '_run() argument must be a buffer, not a stream'
 
     first_line = next(parser.line_generator(buffer)).content
-    if re.match(r'^#\s*yamllint disable-file\s*$', first_line, re.M):
+    if re.match(r'^#\s*yamllint disable-file\s*$', first_line):
         return
 
     # If the document contains a syntax error, save it and yield it at the
```

adrienverge:
Released in v1.19.0 :+1:

Issue #86, opened 2017-11-10
[Feature] Flag to limit output to errors


JakeTheDev:
I'm working on a yaml-backed heiradata project with over 3600 yaml files that I want to auto-lint, and your tools is perfect for it, huge thanks for making this! :)

One thing that would add some sunshine to my life is being able to choose whether I want warnings to show or not. Maybe a rule could be set up with a default like "output-verbosity: normal", where we can tweak it to "output-verbosity: error # or warning" for an auto-build that lints our stuff. Or something smaller in scope like an "only-show-errors: false" that we can set to true.

adrienverge:
Hello,

yamllint is meant to be adapted to your needs using a [configuration file](https://yamllint.readthedocs.io/en/stable/configuration.html). In this simple file, you should use `level: error` or `level: warning` (similar to the `output-verbosity: error` you're proposing) for each rule.

Note:
- If you don't wish to use a file, you can also pass all conf through command line using `-d`, but I find it cumbersome.
- If you have two set of rules (show warnings / don't show warnings), you can also use 2 files and the `-c` argument.

jakethedev:
While i no longer work anywhere near yaml, being able to set a global level would be really nice. Setting the level per rule doesn't hide all warnings, for instance, it just makes those rules louder. Grep-filtering is an option for output filtering, but not for CICD exit code management, and having yamllint not print certain levels of violations would make use of the tool far more ergonomic. 

Just my two cents, i have no investment in this anymore though, that old job is old news 😎

PR #215, opened 2019-12-06
Filter output by problem level

ffapitalle:
Add configuration option `output-verbosity` to enable output filtering by problem level.

Aims to solve #86.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -116,8 +116,8 @@ return code will be:
  * ``1`` if one or more errors occur
  * ``2`` if no errors occur, but one or more warnings occur
 
-If the script is invoked with the ``-q`` (or ``--no-warning``) options, it will
-not output warning level problems, only error level.
+If the script is invoked with the ``--no-warnings`` option, it won't output
+warning level problems, only error level ones.
 
 YAML files extensions
 ---------------------
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -116,6 +116,9 @@ return code will be:
  * ``1`` if one or more errors occur
  * ``2`` if no errors occur, but one or more warnings occur
 
+If the script is invoked with the ``-q`` (or ``--no-warning``) options, it will
+not output warning level problems, only error level.
+
 YAML files extensions
 ---------------------
 
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -533,7 +533,7 @@ class CommandLineTestCase(unittest.TestCase):
             'mapping values are not allowed here\n'))
         self.assertEqual(err, '')
 
-    def test_run_supress_warnings(self):
+    def test_run_no_warnings(self):
         file = os.path.join(self.wd, 'a.yaml')
 
         sys.stdout, sys.stderr = StringIO(), StringIO()
@@ -558,3 +558,12 @@ class CommandLineTestCase(unittest.TestCase):
             cli.run((file, '--no-warnings', '-f', 'auto'))
 
         self.assertEqual(ctx.exception.code, 0)
+
+    def test_run_no_warnings_and_strict(self):
+        file = os.path.join(self.wd, 'warn.yaml')
+
+        sys.stdout, sys.stderr = StringIO(), StringIO()
+        with self.assertRaises(SystemExit) as ctx:
+            cli.run((file, '--no-warnings', '-s'))
+
+        self.assertEqual(ctx.exception.code, 2)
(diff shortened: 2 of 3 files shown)


adrienverge:
Hello @ffapitalle,

Can you explain the use case (indenpendently from #86)?

If the goal is to suppress warnings *using conf*, why not using `level: error` or `disable` *in conf*?

ffapitalle:
Hi @adrienverge,

Main goal is to suppress from the output warning level problems and show only errors.

My use case is exactly the same as #86, I've a proyect with a lot of files that are being validated automatically, and as a result I want a notification containing only errors. Rules marked as warnings should keep reporting as warnings.

I can use also a command line argument, like `-q` or `--no-warnings`, but as I found the former issue , I went that way.

adrienverge:
OK thanks, I understand the use case.

I find confusing and error-prone to have a conf that contradicts itself (*enable this rule as warning*, but *finally don't report it*).

A command line argument makes more sense, as it's temporary and explicitly passed when running the command. Also, an argument would be coherent with `-s` / `--strict`. Would you be OK to change your pull request to use `--no-warnings`?

ffapitalle:
> Would you be OK to change your pull request to use `--no-warnings`?

Agreed. I've have already another branch that adds a command line argument. Would you prefer open a new one ? Or add the changes to this request ?

adrienverge:
As you prefer!

ffapitalle:
> Add configuration option `output-verbosity` to enable output filtering by problem level.
> 
> Aims to solve #86.

Changed proposed solution: instead of adding a configuration option, I've added a command line option, which supress printing of warning level problems on script output.

docs/configuration.rst
```diff
@@ -116,6 +116,9 @@ return code will be:
  * ``1`` if one or more errors occur
  * ``2`` if no errors occur, but one or more warnings occur
 
+If the script is invoked with the ``-q`` (or ``--no-warning``) options, it will
```
adrienverge: - The commit names and previous comments talk about `--no-warnings` (with an `s`), I think it's better English (although I'm not a native English speaker) and more coherent with other options.

- I'd prefer reserve `-q` for later use, moreover the letter `q` would match `--quiet`, but no direct link with `--no-warning`. Can you use only `--no-warnings`? (Same for variable names `quiet` in other files)

tests/test_cli.py
```diff
@@ -532,3 +532,29 @@ def test_run_read_from_stdin(self):
             'stdin:2:10: [error] syntax error: '
             'mapping values are not allowed here\n'))
         self.assertEqual(err, '')
+
+    def test_run_supress_warnings(self):
```
adrienverge: `supress` → `suppress` 

or `test_run_no_warnings()`?

yamllint/cli.py
```diff
@@ -84,11 +84,14 @@ def standard_color(problem, filename):
...
 
 
-def show_problems(problems, file, args_format):
+def show_problems(problems, file, args_format, quiet):
     max_level = 0
     first = True
 
     for problem in problems:
+        max_level = max(max_level, PROBLEM_LEVELS[problem.level])
+        if quiet and (not problem.level == 'error'):
```
adrienverge:
```suggestion
        if quiet and problem.level != 'error':
```

yamllint/cli.py
```diff
@@ -84,11 +84,14 @@ def standard_color(problem, filename):
         return line
 
 
-def show_problems(problems, file, args_format):
+def show_problems(problems, file, args_format, quiet):
     max_level = 0
     first = True
 
     for problem in problems:
+        max_level = max(max_level, PROBLEM_LEVELS[problem.level])
```
adrienverge: I'm curious: why did you move this line?

Is it to keep a non-zero return code if `--strict` is used in the same time as `--no-warnings`? (I think it's a good thing). Can you add a `test_run_no_warnings_and_strict()`? for that?

docs/configuration.rst
ffapitalle: You're right, I've missed the `s` on the doc file, whereas everywhere else I've used `warnings`.

I've also removed `-q` and kept only `--no-warnings`.

tests/test_cli.py
ffapitalle: Another typo (I'm not native English speaker also). I've changed it as you suggests, it's more clear.

yamllint/cli.py
ffapitalle: I wanted to keep the value of the return code , even if I filter the output.

Added the extra test for checking both options given in the same run.

adrienverge: :+1:

PR #217, opened 2019-12-15
Add syntax-error as a code

sylvestre:


adrienverge:
Hey @sylvestre, why not, I'm curious about the use-case? And can you change `syntax-error` to `syntax`?

sylvestre:
Good question, I should have been more explicit! ;)

We are using your tool for Firefox YAML linting (merci btw)
We parse the results with this regexp
`YAMLLINT_FORMAT_REGEX = re.compile(r'(.*):(.*):(.*): \[(error|warning)\] (.*) \((.*)\)$')`
https://searchfox.org/mozilla-central/source/tools/lint/yamllint_/__init__.py#26

which expects the error code. Which doesn't exit with the syntax-error...
We could work around that in our code but I think it should be done upstream instead.

adrienverge:
De rien :) I'm glad to know that Mozilla uses yamllint.

Thanks for your contribution!

sylvestre:
@adrienverge any chance you could publish a release with this change?
thanks

adrienverge:
Sure, let's create the last release of 2019! :christmas_tree: 
https://github.com/adrienverge/yamllint/releases/tag/v1.20.0

Issue #72, opened 2017-09-20
Feature request: shorter syntax for "no rules"


sparr:
I want to run yamllint without any "style" rules, breaking *only* on syntax errors. The shortest way I've found to do that is `yamllint -d "{rules:{}}" file.yaml` which took me a few tries to get right and seems overly verbose.

It would be nice if an empty config file or `-d ""` would work, or even if there was a command line option to disable all rules.

adrienverge:
Hi,

What about `yamllint -d syntax-only file.yaml`?

It could be achieved by creating a config file (`yamllint/conf/syntax-only.yaml`) in this repo.

adrienverge:
I'm closing this because `yamllint -d "{rules:{}}"` is simple to use.

sparr:
> which took me a few tries to get right

`yamllint --help` says
```
  -d CONFIG_DATA, --config-data CONFIG_DATA
                        custom configuration (as YAML source)
```
but it wasn't obvious to me at the time what would make for a minimal valid set of configuration data to produce the desired result. I can't remember what I tried then, but today I tried...

* an empty string `""`, which was `invalid config: not a dict`
* an empty dict `"{}"`, which throws a TypeError, which is possibly issue-worthy itself, since I'd expect it to be caught and reported as another `invalid config`
* a dict with a useless key `"{a:}"`, which works and I must have missed previously in looking more directly at rules-specific behavior

PS: "CONFIG_DATA" doesn't seem to appear in the online docs at all.

PR #219, opened 2019-12-31
Provide a semantics-only configuration

dhutty:
* configure style rules as warnings
    * configure rules that can introduce semantic errors as errors

See https://github.com/adrienverge/yamllint/issues/72


docs/configuration.rst
```diff
@@ -43,6 +43,19 @@ It can be chosen using:
...
+As its name suggests, it is more tolerant:
+
+.. literalinclude:: ../yamllint/conf/semantics-only.yaml
+   :language: yaml
+
+It can be chosen using:
+
+.. code:: bash
+
+ yamllint -d semantics-only file.yml
```
adrienverge: I propose to simplify all this to something like:

```rst
If you want yamllint to only report syntax errors, use the pre-defined
configuration named ``syntax-only``:


.. code:: bash

 yamllint -d syntax-only file.yaml
```

dhutty:
The reason I chose the word 'semantics' was to emphasize that if yamllint exited with an error, it means that the YAML being tested might be semantically different from what the user intended as opposed merely being stylistically/conventionally problematic. 

In order words, my proposed `yamllint -d semantics-only foo.yaml` is a very weak/permissive configuration, which makes most of the common rules about the YAML itself get flagged only as warnings, but anything that could result in broken configuration should cause a failure (i.e. error exit).

PR #229, opened 2020-02-12
Fix new-lines rule on Python 3

bz2:
The fix is reasonably simple, to add the test that fails on Python 3 I also did some refactoring of `test_cli.py` setup. See the individual commits for full details.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -88,6 +88,9 @@ class CommandLineTestCase(unittest.TestCase):
                 u'# 19.99 €\n'
                 u'- お早う御座います。\n'
                 u'# الأَبْجَدِيَّة العَرَبِيَّة\n').encode('utf-8'),
+            # dos line endings yaml
+            'dos.yml': '---\r\n'
+                       'dos: true',
         })
 
     @classmethod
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -377,63 +350,46 @@ def test_run_non_ascii_file(self):
             locale.setlocale(locale.LC_ALL, 'C.UTF-8')
         except locale.Error:
             locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
+        self.addCleanup(locale.setlocale, locale.LC_ALL, loc)
```
adrienverge: Can you revert this?
For correct testing and for reproductibility, it's better if the original locale is restored *at the end of this test*, not *at the end of all tests*.

bz2: [`addCleanup()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup) happens after `tearDown()` - at the end of every test, not at the end of all.

bz2:
Thanks for the prompt review! I'll rearrange the commits as requested.

bz2:
@adrienverge Have rearranged and squashed to two commits as requested.

tests/test_cli.py
adrienverge: OK, thanks for the clarification!

Issue #233, opened 2020-03-08
Bug: quoted-strings rule not working for strings with values matching boolean or number scalars


ruipinge:
Hi,

Given the following `.yamllint` config:

```
extends: default
rules:
    quoted-strings:
        quote-type: single
```

and the yaml file:
```
---
string0: "true"
string1: 'false'
boolean0: true
```

the rule `quoted-strings` should fail on line 2. It doesn't because the string value matches a boolean scalar (although it's actually a string). It also happens for numbers, nulls, and any scalar values.

ruipinge:
I'm already working on a possible fix. Will open PR soon.

adrienverge:
Hi, I confirm the bug. Thanks for handling this!

adrienverge:
Fixed by https://github.com/adrienverge/yamllint/pull/234 and released in v1.21.0.

PR #234, opened 2020-03-08
Fix quoted-strings rules not working for string values matching scalars

ruipinge:
See #233

### DIFF
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -52,27 +52,30 @@ DEFAULT = {'quote-type': 'any'}
 def check(conf, token, prev, next, nextnext, context):
     quote_type = conf['quote-type']
 
-    if (isinstance(token, yaml.tokens.ScalarToken) and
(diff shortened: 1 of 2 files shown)


yamllint/rules/quoted_strings.py
```diff
@@ -52,27 +52,32 @@
...
-            isinstance(prev, (yaml.ValueToken, yaml.TagToken))):
-        # Ignore explicit types, e.g. !!str testtest or !!int 42
-        if (prev and isinstance(prev, yaml.tokens.TagToken) and
-                prev.value[0] == '!!'):
-            return
+    if not isinstance(token, yaml.tokens.ScalarToken):
+        return
+
+    if not isinstance(prev, (yaml.ValueToken, yaml.TagToken)):
+        return
```
adrienverge: For readability, could you simplify to:
```suggestion
    if not (isinstance(token, yaml.tokens.ScalarToken) and
            isinstance(prev, (yaml.ValueToken, yaml.TagToken))):
```

Issue #133, opened 2018-10-11
The rule for using quotes only when they are really needed


scherebedov:
Is it possible to make a rule to use quotes only when they are really needed?
Like a [this](https://stackoverflow.com/questions/19109912/do-i-need-quotes-for-strings-in-yaml?answertab=votes#tab-top) in answer.
Thank's.

adrienverge:
Hi,

Can you be more explicit on when quotes are "really needed"? The link you provide doesn't describe what you want.

For instance:
```yaml
boolean: yes          # I don't want quotes, it's a boolean
number: 10            # I don't want quotes, it's a number
dict: {}              # I don't want quotes, it's an object
string: abc 10 {}     # I don't need quotes, it's a string
```

Please provide explicit examples of what you would like this rule to do.

Also, for info, since version 1.12.0 the `quoted-strings` rule is available. But I don't think this is exactly what you're looking for: https://yamllint.readthedocs.io/en/latest/rules.html#module-yamllint.rules.quoted_strings

adrienverge:
@dctr that makes sense. I'll review contributions to add a new option.

tavin:
@adrienverge does that mean you'd like to see a PR?
I wonder if it should be handled by `quoted-strings` or by its own rule...

adrienverge:
Yes it means that if someones wants to implement this, I'll review the PR. In my opinion it makes most sense to be handled inside `quoted-strings`.

ruipinge:
Although I created a PR with the proposed solution, I believe there's a better one with more flexibility for this rule. Instead of `allow-redundant` something like `usage` with options like:
- `all`: any string value would need to be quoted using `quote-type` definition (currently the default, my suggestion would be to keep it)
- `whatever`: quotes are not mandatory, but when used must follow `quote-type` definition
- `no-redundant`: disallow unneeded quotes, but when needed must follow `quote-type` definition

Any suggestion on different naming are welcome.

If there's an agreement, I'll be glad to update the PR.

ruipinge:
Since they are related, this PR already includes the #233 fix.

@adrienverge any chance we can have your feedback on this?

adrienverge:
@ruipinge thanks for taking care of that. I agree your proposal gives more flexibility and is more future-proof.

About the naming, what do you think of:
```yaml
quoted-strings:
  quote-type: single | double | any
  required: true | false | only-if-needed
```
?

[...]

PR #235, opened 2020-03-09
Add support for redundant quotes in quoted-strings rule

ruipinge:
See #133

### DIFF
--- a/tests/rules/test_quoted_strings.py
+++ b/tests/rules/test_quoted_strings.py
@@ -22,6 +22,7 @@ class QuotedTestCase(RuleTestCase):
 
     def test_disabled(self):
         conf = 'quoted-strings: disable'
+
         self.check('---\n'
                    'foo: bar\n', conf)
         self.check('---\n'
@@ -35,18 +36,18 @@ class QuotedTestCase(RuleTestCase):
 
     def test_quote_type_any(self):
         conf = 'quoted-strings: {quote-type: any}\n'
+
         self.check('---\n'
                    'boolean1: true\n'
                    'number1: 123\n'
                    'string1: foo\n'                          # fails
-                   'string2: "true"\n'
-                   'string3: "123"\n'
-                   'string4: \'true\'\n'
-                   'string5: "foo"\n'
-                   'string6: \'bar\'\n'
-                   'string7: !!str genericstring\n'
-                   'string8: !!str 456\n'
-                   'string9: !!str "quotedgenericstring"\n'
+                   'string2: "foo"\n'
+                   'string3: "true"\n'
+                   'string4: "123"\n'
+                   'string5: \'bar\'\n'
+                   'string6: !!str genericstring\n'
+                   'string7: !!str 456\n'
+                   'string8: !!str "quotedgenericstring"\n'
                    'binary: !!binary binstring\n'
                    'integer: !!int intstring\n'
                    'boolean2: !!bool boolstring\n'
@@ -60,7 +61,7 @@ class QuotedTestCase(RuleTestCase):
                    '  word 1\n'
                    '  word 2\n'
                    'multiline string 3:\n'
-                   '  word 1\n'
+                   '  word 1\n'               # fails
                    '  word 2\n'
                    'multiline string 4:\n'
                    '  "word 1\\\n'
@@ -69,6 +70,7 @@ class QuotedTestCase(RuleTestCase):
 
     def test_quote_type_single(self):
         conf = 'quoted-strings: {quote-type: single}\n'
+
         self.check('---\n'
                    'boolean1: true\n'
                    'number1: 123\n'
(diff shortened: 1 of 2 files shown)


tests/rules/test_quoted_strings.py
```diff
@@ -17,11 +17,47 @@
...
+multiline string 3:
+  word 1
+  word 2
+multiline string 4:
+  "word 1\\
+   word 2"
+---
+"""
+
+
```
ruipinge: @adrienverge Extracted the two yaml strings so they can be reused and each test becomes cleaner.

yamllint/rules/quoted_strings.py
```diff
@@ -15,15 +15,23 @@
...
+quoted strings without needing it. You can also enforce the type of the quote
+used.
+
+.. rubric:: Options
+
+* ``quote-type`` defines allowed quotes: ``single``, ``double`` or ``any``
+  (default).
+* ``required`` defines whether using quotes in string values is required
+  (``true``, default) or not (``false``), or mandatory when really needed
+  (``only-when-needed``)
```
adrienverge:
```suggestion
  (``only-when-needed``).
```

yamllint/rules/quoted_strings.py
```diff
@@ -15,15 +15,23 @@
...
+Use this rule to forbid any string values that are not quoted, or to prevent
+quoted strings without needing it. You can also enforce the type of the quote
+used.
+
+.. rubric:: Options
+
+* ``quote-type`` defines allowed quotes: ``single``, ``double`` or ``any``
+  (default).
+* ``required`` defines whether using quotes in string values is required
+  (``true``, default) or not (``false``), or mandatory when really needed
```
adrienverge: `or mandatory when really needed` → `or allowed when really needed`?

Otherwise it could be interpreted the other way? (I may be wrong, I'm not a native English speaker.)

yamllint/rules/quoted_strings.py
```diff
@@ -37,6 +45,25 @@
    ::
 
     foo: bar
+
+
```
adrienverge: You can drop a useless line here.

yamllint/rules/quoted_strings.py
```diff
@@ -45,13 +72,23 @@
...
 TYPE = 'token'
-CONF = {'quote-type': ('any', 'single', 'double')}
-DEFAULT = {'quote-type': 'any'}
+CONF = {'quote-type': ('any', 'single', 'double'),
+        'required': (True, False, 'only-when-needed')}
+DEFAULT = {'quote-type': 'any',
+           'required': True}
 
+DEFAULT_SCALAR_TAG = u'tag:yaml.org,2002:str'
+START_TOKENS = {'#', '*', '!', '?', '@', '`', '&',
```
adrienverge: I wondered why you included `@` and `, but you're right:
> The “@” (#x40, at) and “`” (#x60, grave accent) are reserved for future use.

-- https://yaml.org/spec/1.2/spec.html#id2774058

tests/rules/test_quoted_strings.py
```diff
@@ -35,102 +71,59 @@ def test_disabled(self):
...
+        self.check(YAML, conf, problem2=(5, 10),
+                   problem3=(6, 10), problem4=(7, 10))
+        self.check(YAML_MULTILINE, conf, problem1=(12, 3))
+
+    def test_quotes_required(self):
+        conf = 'quoted-strings: {quote-type: any, required: true}\n'
+
+        self.check(YAML, conf, problem2=(4, 10))
+        self.check(YAML_MULTILINE, conf, problem1=(9, 3))
```
adrienverge: I would also have expected lines 10 and 11 to be reported:

```yaml
string6: !!str genericstring
string7: !!str 456
```
... but it's not currently the case on master with `{quote-type: any}`. I think that's acceptable!

tests/rules/test_quoted_strings.py
adrienverge: I understand you removed them to avoid code duplication, but having source YAML directly above expected errors really helps reading tests, and checking everything works as expected.

Unless you really really prefer like that, can you please revert this change?

tests/rules/test_quoted_strings.py
ruipinge: I didn't change the previous behavior and I actually agree with it: if, for some reason, you are explicitly defining the type, you should know what you're doing. Makes sense?

adrienverge: Agreed!

tests/rules/test_quoted_strings.py
ruipinge: I honestly was divided between the two approaches and ended up choosing the one with less code duplication because the testes seemed cleaner to me. But I understand it's something completely opinionated and... you should be the one playing the preference card. ;)

yamllint/rules/quoted_strings.py
ruipinge: I find this one better: `or only allowed when really needed`. Ok?

adrienverge: Perfect!

PR #240, opened 2020-03-25
Detect encoding per yaml spec (fix #238)

### LINKED ISSUE #238: Failure to read file in 1.21.0 with Python 2.7 and non-ascii yaml files.
Yamllint is now failing on files that linted successfully in previous versions. This happens in files that are not entirely ascii. With Python 3.x this issue does not occur.

I believe this is due to a [change in cli.py](https://github.com/adrienverge/yamllint/blame/542ae758f5de70a37c6c322c16b83b92fdc008bc/yamllint/cli.py#L180) where `io.open` is used instead of `open`.

Relevant Traceback:
```
 $ yamllint -c .lintConfig .
   File "/usr/local/bin/yamllint", line 11, in <module>
     sys.exit(run())
   File "/usr/local/lib/python2.7/dist-packages/yamllint/cli.py", line 181, in run
     problems = linter.run(f, conf, filepath)
   File "/usr/local/lib/python2.7/dist-packages/yamllint/linter.py", line 237, in run
     content = input.read()
   File "/usr/lib/python2.7/encodings/ascii.py", line 26, in decode
     return codecs.ascii_decode(input, self.errors)[0]
 UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 132: ordinal not in range(128)
```

spyoungtech:
Resolves #238

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -91,6 +91,10 @@ class CommandLineTestCase(unittest.TestCase):
             # dos line endings yaml
             'dos.yml': '---\r\n'
                        'dos: true',
+            # UTF-16 BOM
+            'non-ascii/utf16': b'\xff\xfe---\nutf16: true\n',
+            #UTF-8 BOM
+            'non-ascii/utf8': b'\xef\xbb\xbf---\nutf8: true\n',
         })
 
     @classmethod
@@ -171,6 +175,8 @@ class CommandLineTestCase(unittest.TestCase):
              os.path.join(self.wd, 'dos.yml'),
              os.path.join(self.wd, 'empty.yml'),
              os.path.join(self.wd, 'no-yaml.json'),
+             os.path.join(self.wd, 'non-ascii/utf16'),
+             os.path.join(self.wd, 'non-ascii/utf8'),
              os.path.join(self.wd, 'non-ascii/éçäγλνπ¥/utf-8'),
              os.path.join(self.wd, 's/s/s/s/s/s/s/s/s/s/s/s/s/s/s/file.yaml'),
              os.path.join(self.wd, 'sub/ok.yaml'),
@@ -188,6 +194,8 @@ class CommandLineTestCase(unittest.TestCase):
              os.path.join(self.wd, 'dos.yml'),
              os.path.join(self.wd, 'empty.yml'),
              os.path.join(self.wd, 'no-yaml.json'),
+             os.path.join(self.wd, 'non-ascii/utf16'),
+             os.path.join(self.wd, 'non-ascii/utf8'),
              os.path.join(self.wd, 'non-ascii/éçäγλνπ¥/utf-8'),
              os.path.join(self.wd, 's/s/s/s/s/s/s/s/s/s/s/s/s/s/s/file.yaml'),
              os.path.join(self.wd, 'sub/ok.yaml'),
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -91,8 +91,10 @@ class CommandLineTestCase(unittest.TestCase):
             # dos line endings yaml
             'dos.yml': '---\r\n'
                        'dos: true',
-            # UTF-16 BOM
-            'non-ascii/utf16': b'\xff\xfe---\nutf16: true\n',
+            # UTF-16 Little Endian BOM
+            'non-ascii/utf16le': b'\xff\xfe---\nutf16le: true\n',
+            #UTF-16 Big Endian
+            'non-ascii/utf16be': b'\xfe\xff---\nutf16be: true\n',
             #UTF-8 BOM
             'non-ascii/utf8': b'\xef\xbb\xbf---\nutf8: true\n',
         })
@@ -175,7 +177,8 @@ class CommandLineTestCase(unittest.TestCase):
              os.path.join(self.wd, 'dos.yml'),
              os.path.join(self.wd, 'empty.yml'),
              os.path.join(self.wd, 'no-yaml.json'),
-             os.path.join(self.wd, 'non-ascii/utf16'),
+             os.path.join(self.wd, 'non-ascii/utf16be'),
+             os.path.join(self.wd, 'non-ascii/utf16le'),
              os.path.join(self.wd, 'non-ascii/utf8'),
              os.path.join(self.wd, 'non-ascii/éçäγλνπ¥/utf-8'),
              os.path.join(self.wd, 's/s/s/s/s/s/s/s/s/s/s/s/s/s/s/file.yaml'),
@@ -194,7 +197,8 @@ class CommandLineTestCase(unittest.TestCase):
              os.path.join(self.wd, 'dos.yml'),
              os.path.join(self.wd, 'empty.yml'),
              os.path.join(self.wd, 'no-yaml.json'),
-             os.path.join(self.wd, 'non-ascii/utf16'),
+             os.path.join(self.wd, 'non-ascii/utf16be'),
+             os.path.join(self.wd, 'non-ascii/utf16le'),
              os.path.join(self.wd, 'non-ascii/utf8'),
              os.path.join(self.wd, 'non-ascii/éçäγλνπ¥/utf-8'),
              os.path.join(self.wd, 's/s/s/s/s/s/s/s/s/s/s/s/s/s/s/file.yaml'),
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -91,6 +91,12 @@ def setUpClass(cls):
             # dos line endings yaml
             'dos.yml': '---\r\n'
                        'dos: true',
+            # UTF-16 Little Endian BOM
+            'non-ascii/utf16le': b'\xff\xfe---\nutf16le: true\n',
+            # UTF-16 Big Endian
+            'non-ascii/utf16be': b'\xfe\xff---\nutf16be: true\n',
```
bz2: This test data is wrong.
```py
>>> b'\xff\xfe---\nutf16le: true\n'.decode('utf-16-le')
u'\ufeff\u2d2d\u0a2d\u7475\u3166\u6c36\u3a65\u7420\u7572\u0a65'
```
Better to have u'...'.encode('utf-16-be') - and also need a test that after decoding the files do contain the expected data?

spyoungtech: Ah, of course. My mistake. UTF-16 and ascii are not compatible, hence the garbage. This would only work correctly for UTF-8. I'll adjust as suggested.

adrienverge: I don't quite get why this is wrong. Anyway, does it still need to be adjusted?

tests/test_cli.py
```diff
@@ -171,6 +177,9 @@ def test_find_files_recursively(self):
              os.path.join(self.wd, 'dos.yml'),
              os.path.join(self.wd, 'empty.yml'),
              os.path.join(self.wd, 'no-yaml.json'),
+             os.path.join(self.wd, 'non-ascii/utf16be'),
+             os.path.join(self.wd, 'non-ascii/utf16le'),
+             os.path.join(self.wd, 'non-ascii/utf8'),
```
adrienverge: For consistency with the path just below, and for easier `grep`ping, can you change to `utf-8`, `utf-16-le`, etc.? Thanks!

tests/test_cli.py
```diff
@@ -517,3 +529,18 @@ def test_run_non_universal_newline(self):
             '\n' % path)
         self.assertEqual(
             (ctx.returncode, ctx.stdout, ctx.stderr), (1, expected_out, ''))
+
+    def test_encoding_detection_utf16le(self):
+        path = os.path.join(self.wd, 'non-ascii/utf16le')
+        encoding = cli.determine_encoding(path)
+        self.assertEqual(encoding, 'utf-16-le')
```
adrienverge: Can you also run `yamllint` on these files, just to make sure everything works fine?
```suggestion
        self.assertEqual(encoding, 'utf-16-le')

        with RunContext(self) as ctx:
            cli.run(())
        self.assertNotEqual(ctx.returncode, 0)
        self.assertEqual(ctx.stdout, '')
        self.assertEqual(ctx.stderr, '')
```
(same for other encodings)

spyoungtech:
Just wanted to say I'm still working on this, but have had some things come up. I should get around to fixing up the review items this weekend. Maintainers are also free to commit directly to my branch :-)

adrienverge:
Thanks @spyoungtech, no problem.

Sure, I could fix the few issues myself, but before merging I just want to be sure that:
- nothings needs to be changed regarding https://github.com/adrienverge/yamllint/pull/240/files#r397801149,
- > Do you know any solution to accept _custom user encodings_, depending on the system yamllint is run from? E.g. reading `LC_ALL` environment var?

  @bz2 maybe you have an opinion on that?

bz2:
@adrienverge I think the right thing on non-unicode encodings from a user perspective is add a new error type but otherwise parse as well as possible. Could implement that either as falling back to parse via locale encoding, and/or replacing non-utf8 byte sequences on parse.

tests/test_cli.py
bz2: It's wrong because utf-16 is a two-byte encoding, but the sample data is using `'-'` for instance, not `'u\x00'`.

spyoungtech: Another illustration of the issue:

```python
In [1]: data = b'\xff\xfe---\nutf16le: true\n'

In [2]: data.decode('utf-16-le')
Out[2]: '\ufeffⴭਭ瑵ㅦ氶㩥琠畲\u0a65'
```

This is because the bytestring uses ascii, which doesn't map back to utf-16.

The correct way looks more like this:

```python
In [4]: b'\xff\xfe' + '---\nutf16le: true'.encode('utf-16-le')
Out[4]: b'\xff\xfe-\x00-\x00-\x00\n\x00u\x00t\x00f\x001\x006\x00l\x00e\x00:\x00 \x00t\x00r\x00u\x00e\x00'

In [5]: _.decode('utf-16-le')
Out[5]: '\ufeff---\nutf16le: true'
```

spyoungtech:
> @adrienverge I think the right thing on non-unicode encodings from a user perspective is add a new error type but otherwise parse as well as possible. Could implement that either as falling back to parse via locale encoding, and/or replacing non-utf8 byte sequences on parse.

Agree. Perhaps it would be reasonable then to use something like `chardet` for encoding detection? It should handle the BOMs the same way, if present.

spyoungtech:
Okay, I went ahead and committed a new change to create a new function `yamlopen` that will use `chardet` for character detection if no encoding is provided explicitly (giving you the option to do this later, if you so choose!).  Under the hood `io.TextIOWrapper` is used (instead of StringIO, for example) so you can inspect the `encoding` attribute, allowing you guys to write rules based on the detected encoding, if you want.

yamllint/cli.py
```diff
@@ -177,7 +190,8 @@ def run(argv=None):
     for file in find_files_recursively(args.files, conf):
         filepath = file[2:] if file.startswith('./') else file
         try:
-            with io.open(file, newline='') as f:
+            encoding = determine_encoding(file)
+            with io.open(file, newline='', encoding=encoding) as f:
```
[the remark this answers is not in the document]
spyoungtech: done!

spyoungtech:
> do you know any easy solution to accept custom user encodings, depending on the system yamllint is run from? E.g. reading LC_ALL environment var?

Simply put, UTF-8 is the most sensible default; it's also the default of the spec. I had considered using `locale.getpreferredencoding()` but it would basically guarantee errors on Windows and would probably often return encodings that are not valid per YAML spec if it's not UTF-8. `sys.getdefaultencoding()` is another option, but has similar drawbacks.

The solution I put forward fixes the bug in #238 and also allows for an explicit encoding to be provided, should you ever decide on a mechanism to provide that encoding (perhaps command line argument? ). 

Another option would be to pass an appropriate argument into the `errors`  parameter for open. For example `replace` or `surrogateescape` to at least allow an attempt at linting, despite encoding errors.

tests/test_cli.py
```diff
@@ -517,3 +536,55 @@ def test_run_non_universal_newline(self):
...
+        path = os.path.join(self.wd, 'non-ascii/utf-16-be')
+        with cli.yamlopen(path) as yaml_file:
+            yaml_file.read()
+        self.assertEqual(yaml_file.encoding, 'UTF-16')
+
+    def test_explicit_encoding(self):
+        path = os.path.join(self.wd, 'a.yaml')
+        with cli.yamlopen(path, encoding='windows-1252') as yaml_file:
+            yaml_file.read()
+        self.assertEqual(yaml_file.encoding, 'windows-1252')
```
adrienverge, line 584: This does not test anything because it was already opened with `encoding='windows-1252'`, does it?

spyoungtech: It tests that `yamlopen` correctly accepts an explicit encoding, instead of detecting the encoding with chardet.

There are [tests that run the linter](https://github.com/adrienverge/yamllint/pull/240/files#diff-09f983a1db5a7aca6102f2e2a260c88cR540-R556) on all the new non-ascii files, except for the file that is made up of random bytes that cannot be decoded and exists just to be sure UTF-8 is used as a default if chardet can't detect any encoding.

I understand not wanting to add dependencies. If it's any consolation chardet itself is pure python with no external dependencies. I can also go back and revert to just checking the BOM, per the yaml spec and expecting failure or malformed data in any other case.

The motivation for using chardet was [the comment from bz2](https://github.com/adrienverge/yamllint/pull/240#issuecomment-608387745) about making a best-effort to parse files even if they're not an acceptable encoding.

The yamlopen context manager was also implemented [per request](https://github.com/adrienverge/yamllint/pull/240#discussion_r397795662) on review. I can remove that, too, if you prefer not to use it.

adrienverge: > It tests that `yamlopen` correctly accepts an explicit encoding, instead of detecting the encoding with chardet.

OK, thanks for the explanation!

> There are [tests that run the linter](https://github.com/adrienverge/yamllint/pull/240/files#diff-09f983a1db5a7aca6102f2e2a260c88cR540-R556) on all the new non-ascii files, except for the file that is made up of random bytes that cannot be decoded and exists just to be sure UTF-8 is used as a default if chardet can't detect any encoding.

Sure. But still no tests for ISO-8859-1-encoded file, unless I missed something?

> I understand not wanting to add dependencies. If it's any consolation chardet itself is pure python with no external dependencies. I can also go back and revert to just checking the BOM, per the yaml spec and expecting failure or malformed data in any other case.
> 
> The motivation for using chardet was [the comment from bz2](https://github.com/adrienverge/yamllint/pull/240#issuecomment-608387745) about making a best-effort to parse files even if they're not an acceptable encoding.

Sure, I understood that. But I'm very against adding new dependencies (pure Python or not).
Have you seen my proposal about a temporary Python-2-only solution? What do you think of it?

> The yamlopen context manager was also implemented [per request](https://github.com/adrienverge/yamllint/pull/240#discussion_r397795662) on review. I can remove that, too, if you prefer not to use it.

Yes I've seen it, but I don't think it helps readability.

PR #244, opened 2020-03-31
Add a "quote-percent" option to the "quoted-strings" rule

leofeyer:
I really like the new `required: only-when-needed` option that has been added to the `quoted-strings` rule in version 1.21.0. Thank you for that.

When linting Symfony configuration files, there is a kind of false-positive though: Symfony requires placeholders such as `%kernel.project_dir%` or `%env(DATABASE_URL)%` to be quoted. Now even though this deviates from the YAML standard, it would be nice if yamllint would support this somehow.

This PR outlines a possible solution by adding a `quote-percent` option to the `quoted-strings` rule. Tests are still missing, because I wanted to discuss the implementation first. WDYT?

### DIFF
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -72,9 +72,11 @@ from yamllint.linter import LintProblem
 ID = 'quoted-strings'
 TYPE = 'token'
 CONF = {'quote-type': ('any', 'single', 'double'),
-        'required': (True, False, 'only-when-needed')}
+        'required': (True, False, 'only-when-needed'),
+        'quote-percent': (True, False)}
 DEFAULT = {'quote-type': 'any',
-           'required': True}
+           'required': True,
+           'quote-percent': False}
 
 DEFAULT_SCALAR_TAG = u'tag:yaml.org,2002:str'
 START_TOKENS = {'#', '*', '!', '?', '@', '`', '&',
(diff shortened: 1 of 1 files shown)


adrienverge:
Hello @leofeyer, thanks for your proposal.

I'm against features specific to one program (including Symfony), and I think `quote-percent` usage is too narrow to be a good addition to yamllint.

However, we could imagine something more general, like a new option `needed-regex` (or any better wording)?

Then for Symfony you could use:
```yaml
quoted-strings:
  required: only-when-needed
  needed-regex: ^%.*%$
```

If you like the idea, feel free to open a new PR!

leofeyer:
What about making `START_TOKENS` configurable?

```py
CONF = {'quote-type': ('any', 'single', 'double'),
        'required': (True, False, 'only-when-needed'),
        'start-tokens': set()}
DEFAULT = {'quote-type': 'any',
           'required': True,
           'start-tokens': {'#', '*', '!', '?', '@', '`', '&', ',', '-', '{', '}', '[', ']', ':'}}
```

I am not familiar with Python, so no idea if that is the correct syntax.

adrienverge:
It wouldn't be a general solution. What if I want to quote strings that look like phone numbers, for example? A regex would be more configurable.

leofeyer:
I agree. 👍 As I said, I am not familiar with Python, but I will try my best to create a PR that implements the `needed-extra-regex` option.

adrienverge:
Cool!

I recommend starting by writing tests in `tests/rules/test_quoted_strings.py`. Write exactly what you expect from the rule (including corner cases), then it will be easy to check your feature while you write it (using `python3 -m unittest tests.rules.test_quoted_strings`).

leofeyer:
See #246. 🎉

PR #245, opened 2020-03-31
JSON output (yamllint -f json)

strowi:
Hi @adrienverge,

here is the pull-request regarding json-output. Hope i got the coverage-test right. The problem was generating a really valid json for the whole report, not just a single problem.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -101,6 +101,25 @@ machine (for instance for :doc:`syntax highlighting in text editors
  file.yml:57:1: [error] trailing spaces (trailing-spaces)
  file.yml:60:3: [error] wrong indentation: expected 4 but found 2 (indentation)
 
+Add the ``-f json`` arguments if you need an output formatted as json.
+The output will then look like:
+
+
+::
+
+ [
+    {
+      "path": "file.yml",
+      "line": 6,
+      "char": 2,
+      "description": "[warning] missing starting space in comment (comments)",
+      "code": "yamllint",
+      "name": "yamllint",
+      "severity": "warning"
+    }
+  ]
+
+
 If you have a custom linting configuration file (see :doc:`how to configure
 yamllint <configuration>`), it can be passed to yamllint using the ``-c``
 option:
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -101,9 +101,8 @@ machine (for instance for :doc:`syntax highlighting in text editors
  file.yml:57:1: [error] trailing spaces (trailing-spaces)
  file.yml:60:3: [error] wrong indentation: expected 4 but found 2 (indentation)
 
-Add the ``-f json`` arguments if you need an output formatted as json.
+Add the ``-f json`` arguments if you need an output formatted as JSON.
 The output will then look like:
-
 
 ::
 
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -112,8 +112,6 @@ The output will then look like:
       "line": 6,
       "char": 2,
       "description": "[warning] missing starting space in comment (comments)",
-      "code": "yamllint",
-      "name": "yamllint",
       "severity": "warning"
     }
   ]
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -25,6 +25,7 @@ import pty
 import shutil
 import sys
 import unittest
+import json
 
 from tests.common import build_temp_workspace
 
@@ -517,3 +518,11 @@ class CommandLineTestCase(unittest.TestCase):
             '\n' % path)
         self.assertEqual(
             (ctx.returncode, ctx.stdout, ctx.stderr), (1, expected_out, ''))
+
+    def test_json_output(self):
+        path = os.path.join(self.wd, 'a.yaml')
+
+        with RunContext(self) as ctx:
+            cli.run(('-f', 'json', path))
+        print (ctx.stdout)
+        self.assertTrue(json.loads(ctx.stdout))
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -92,8 +92,6 @@ class Format(object):
             "line": problem.line,
             "char": problem.column,
             "description": problem.message,
-            "code": "yamllint",
-            "name": "yamllint",
             "severity": problem.level,
         })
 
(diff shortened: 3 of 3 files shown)


docs/quickstart.rst
```diff
@@ -101,6 +101,25 @@ machine (for instance for :doc:`syntax highlighting in text editors
  file.yml:57:1: [error] trailing spaces (trailing-spaces)
  file.yml:60:3: [error] wrong indentation: expected 4 but found 2 (indentation)
 
+Add the ``-f json`` arguments if you need an output formatted as json.
```
adrienverge: as json → as JSON

docs/quickstart.rst
```diff
@@ -101,6 +101,25 @@ machine (for instance for :doc:`syntax highlighting in text editors
...
+The output will then look like:
+
+
```
adrienverge: Let's say consistent with the rest of the file, no need for 2 blank lines.

yamllint/cli.py
```diff
@@ -107,12 +122,14 @@ def show_problems(problems, file, args_format, no_warn):
...
 
+    if args_format == 'json':
+        print(json.dumps(problems_json))
+
     if not first and args_format != 'parsable':
         print('')
 
     return max_level
 
-
```
adrienverge: Why this?

yamllint/cli.py
```diff
@@ -130,7 +147,7 @@ def run(argv=None):
                               action='store',
                               help='custom configuration (as YAML source)')
     parser.add_argument('-f', '--format',
-                        choices=('parsable', 'standard', 'colored', 'auto'),
+                        choices=('parsable', 'standard', 'colored', 'auto', 'json'),
```
adrienverge: Can you keep `auto` in last position?

yamllint/cli.py
```diff
@@ -84,17 +85,31 @@ def standard_color(problem, filename):
...
 
+    @staticmethod
+    def json(problem, filename):
+        return json.dumps({
+            "path": filename,
+            "line": problem.line,
+            "char": problem.column,
+            "description": problem.message,
+            "code": "yamllint",
+            "name": "yamllint",
```
adrienverge: I don't the use for `code` and `name`. If there isn't, can you remove them?

yamllint/cli.py
```diff
@@ -84,17 +85,31 @@ def standard_color(problem, filename):
...
+    problems_json = []
 
     for problem in problems:
         max_level = max(max_level, PROBLEM_LEVELS[problem.level])
         if no_warn and (problem.level != 'error'):
             continue
         if args_format == 'parsable':
             print(Format.parsable(problem, file))
+        elif args_format == 'json':
+            problems_json.append(json.loads(Format.json(problem, file)))
```
adrienverge: The code does `json.dumps()` then `json.loads()` then `json.dumps()` again. Maybe it could be simplified.

strowi: Thx for taking the time. All the above are valid, maybe i was a bit hasty or this one got me too many headaches. (not a python pro...yet;)), and yes it's ugly.

But i am not sure how to make python return the json-object instead of the json-string and NOT break json-validity (there are descriptions containing double-quotes).

adrienverge: No problem, your code looks globally good!

The first thing you could do is remove `json.loads()` from here and `json.dumps()` from `Format`. It would change anything, except lightening the code.

Another approach would be to directly print json.dumped strings + `',\n'` in this line, and handle the starting `[` and ending `]` outside the loop.

strowi: Oh right;) I first did the `,\n`and `[]`-thing but that wasn't really json-compatible, because on the last array-element it breaks json, which would've needed another exception..
I just pushed another update with the first-mentioned changes.

docs/quickstart.rst
```diff
@@ -101,6 +101,24 @@ machine (for instance for :doc:`syntax highlighting in text editors
...
+::
+
+ [
+    {
+      "path": "file.yml",
+      "line": 6,
+      "char": 2,
+      "description": "[warning] missing starting space in comment (comments)",
+      "code": "yamllint",
+      "name": "yamllint",
```
adrienverge, line 116: There are still `code` and `name`.

tests/test_cli.py
```diff
@@ -25,6 +25,7 @@
 import shutil
 import sys
 import unittest
+import json
```
adrienverge, line 28: Please keep alphabetical order, some linters complain about this.

tests/test_cli.py
```diff
@@ -517,3 +518,11 @@ def test_run_non_universal_newline(self):
             '\n' % path)
         self.assertEqual(
             (ctx.returncode, ctx.stdout, ctx.stderr), (1, expected_out, ''))
+
+    def test_json_output(self):
+        path = os.path.join(self.wd, 'a.yaml')
+
+        with RunContext(self) as ctx:
+            cli.run(('-f', 'json', path))
+        print (ctx.stdout)
```
adrienverge, line 527: Why printing this line?

tests/test_cli.py
```diff
@@ -517,3 +518,11 @@ def test_run_non_universal_newline(self):
...
+        self.assertTrue(json.loads(ctx.stdout))
```
adrienverge, line 528: I don't understand what you test here.

Again, can you please make a real check? For example:
```python
        with RunContext(self) as ctx:
            cli.run(('-f', 'json', path))
        expected_out = (
            '[{"path":"%s",...' % path +
            '"line":1,"char":1,...},'
            '{"path":"file2",...'
            '"line":1,"char":1,...},]\n')
        self.assertEqual(
            (ctx.returncode, ctx.stdout, ctx.stderr), (0, expected_out, ''))
```

strowi:
> Thanks @strowi, it looks better but not perfect yet.

I think you missed my comment:

    At first glance it looks good, but lacks tests. For example, how does it behave when yamllint is run on multiple files, e.g. yamllint file.yaml other-file.yaml dir/? (I haven't tested, but it looks like it outputs incorrect JSON.)

Yes, i did. But looking more into this, it seems i am at a loss.. 
I should've looked more closely before, i missed the `for file in find_files_recursively(args.files, conf)`-part.
Right now it's generating an array for each file and just concats them.

Because of the way json is structured, the last line can't have a `,` at the end, and right now i have no idea how to do this by code for the file.. maybe i need to think about this some more..
```
[ {"1": "2"},
...
{"3":"4"}]

bluikko:
Please consider using the Codeclimate JSON syntax. Why come up with a proprietary JSON schema when a de-fact standard exists? It would also be supported in GitLab.

Issue #242, opened 2020-03-30
`quoted-strings` only-when-needed incorrectly detecting strings with redundant quotes


RealOrangeOne:
Since upgrading to 1.21.0, I enabled the following config:

```yaml
  quoted-strings:
    quote-type: double
    required: only-when-needed
```

Unfortunately, seems there's some strings which it believes don't need quoting, when in fact they do:

- `aur_builder ALL=(ALL) NOPASSWD: /usr/bin/pacman`
- `%H:%M:%S`
- `'Mac' in ansible_facts.product_name`

It seems to be something around when a string either contains or starts with a special character, but isn't entirely a special character?

adrienverge:
Hi @RealOrangeOne (cc @ruipinge), thanks for the report and sorry about that.

There are actually 2 problems:
1. Strings that actually need quotes, but `only-when-needed` reports them as "redundantly quoted":
   \- `"aur_builder ALL=(ALL) NOPASSWD: /usr/bin/pacman"`
   \- `"%H:%M:%S"`
   \- `"'Mac' in ansible_facts.product_name"`
2. Strings that don't need quotes, but undetected by `only-when-needed`:
   \- `"-----"`
   \- `":char"`

I believe using
https://github.com/adrienverge/yamllint/blob/1a13837e84431f0fe6823cb655886daec0211eb9/yamllint/rules/quoted_strings.py#L80-L81
and checking `token.value[0] in START_TOKENS` is not a good way to do, because we can't ensure the need for quotes using only the first character.

Could an alternative solution be to drop `START_TOKENS`, and try parsing the value to check whether it would be seen as a string or something else?
Do you have any other idea?

RealOrangeOne:
Perhaps, although there are complexities. For example, given `foo: 6`, should it be quoted or not?

Perhaps attempting to parse the value, and if it results in a complex type (ie list, object, parse error, whatever), then raise? 

I'd be interested in knowing if there's a solution for the simpler data types (int, bool etc), but I can't think of one.

adrienverge:
From a user perspective, `foo: 6` isn't a problem: either it's quoted (therefore it's a string), or it's not (so it's a mapping). In either case, yamllint shouldn't complain.

I'm not sure I understand what you mean by "then raise".

To be clearer about my idea on `required: only-when-needed`: if a *quoted* string is found (e.g. `"foo: bar"` or `"-----"`), try to parse its contents as YAML.
[...]

PR #246, opened 2020-04-01
Add the "needed-extra-regex" option

leofeyer:
This is the follow-up PR to #244, which adds the `needed-extra-regex` option to the `quoted-strings` rule.

```yaml
quoted-strings:
  required: only-when-needed
  needed-extra-regex: ^%.*$
```

This will suppress the "string value is redundantly quoted" warning for strings starting with a `%` character.

### DIFF
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -26,6 +26,8 @@ used.
 * ``required`` defines whether using quotes in string values is required
   (``true``, default) or not (``false``), or only allowed when really needed
   (``only-when-needed``).
+* ``needed-extra-regex`` allows strings matching the given PCRE regex to be
+  quoted even when quoting is only allowed when really needed
 
 **Note**: Multi-line strings (with ``|`` or ``>``) will not be checked.
 
@@ -63,9 +65,22 @@ used.
    ::
 
     foo: 'bar'
+
+#. With ``quoted-strings: {quote-type: single, required: only-when-needed, needed-extra-regex: ^%.*%$ }``
+
+   the following code snippet would **PASS**:
+   ::
+
+    foo: '%bar%'
+
+   the following code snippet would **FAIL**:
+   ::
+
+    foo: 'bar'
 """
 
 import re
+
 import yaml
 
 from yamllint.linter import LintProblem
@@ -137,7 +152,7 @@ def check(conf, token, prev, next, nextnext, context):
                 and token.value[0] not in START_TOKENS):
             extra_regex = conf['needed-extra-regex']
 
-            if extra_regex == '' or not re.match(extra_regex, token.value):
+            if extra_regex == '' or not re.search(extra_regex, token.value):
                 msg = "string value is redundantly quoted with %s quotes" % (
                     quote_type)
 
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -128,6 +128,7 @@ def check(conf, token, prev, next, nextnext, context):
 
     quote_type = conf['quote-type']
     required = conf['required']
+    extra_regex = conf['needed-extra-regex']
 
     # Completely relaxed about quotes (same as the rule being disabled)
     if required is False and quote_type == 'any':
@@ -150,12 +151,10 @@ def check(conf, token, prev, next, nextnext, context):
 
         # Quotes are disallowed when not needed
         if (tag == DEFAULT_SCALAR_TAG and token.value
-                and token.value[0] not in START_TOKENS):
-            extra_regex = conf['needed-extra-regex']
-
-            if extra_regex == '' or not re.search(extra_regex, token.value):
-                msg = "string value is redundantly quoted with %s quotes" % (
-                    quote_type)
+                and token.value[0] not in START_TOKENS and (extra_regex == ''
+                or not re.search(extra_regex, token.value))):
+            msg = "string value is redundantly quoted with %s quotes" % (
+                quote_type)
 
         # But when used need to match config
         elif token.style and not quote_match(quote_type, token.style):
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -26,8 +26,8 @@ used.
 * ``required`` defines whether using quotes in string values is required
   (``true``, default) or not (``false``), or only allowed when really needed
   (``only-when-needed``).
-* ``needed-extra-regex`` allows strings matching the given PCRE regex to be
-  quoted even when quoting is only allowed when really needed
+* ``needed-extra-regex`` requires strings matching the given PCRE regex to be
+  quoted even when quoting is only allowed when really needed.
 
 **Note**: Multi-line strings (with ``|`` or ``>``) will not be checked.
 
@@ -77,6 +77,7 @@ used.
    ::
 
     foo: 'bar'
+    foo: %bar%
 """
 
 import re
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -65,6 +65,7 @@ used.
     foo: 'bar'
 """
 
+import re
 import yaml
 
 from yamllint.linter import LintProblem
@@ -72,9 +73,11 @@ from yamllint.linter import LintProblem
 ID = 'quoted-strings'
 TYPE = 'token'
 CONF = {'quote-type': ('any', 'single', 'double'),
-        'required': (True, False, 'only-when-needed')}
+        'required': (True, False, 'only-when-needed'),
+        'needed-extra-regex': str}
 DEFAULT = {'quote-type': 'any',
-           'required': True}
+           'required': True,
+           'needed-extra-regex': ''}
 
 DEFAULT_SCALAR_TAG = u'tag:yaml.org,2002:str'
 START_TOKENS = {'#', '*', '!', '?', '@', '`', '&',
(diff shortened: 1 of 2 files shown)


leofeyer:
Neither the failing tests nor the coverage decrease seem related to my changes. 🤷‍♂

yamllint/rules/quoted_strings.py
```diff
@@ -65,16 +65,19 @@
     foo: 'bar'
 """
 
+import re
```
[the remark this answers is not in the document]
leofeyer: Fixed in ec8d24dd585d5d24dd81c120c177ca9c12384e42.

leofeyer:
I have addressed all requested changes in ec8d24dd585d5d24dd81c120c177ca9c12384e42.

yamllint/rules/quoted_strings.py
```diff
@@ -26,6 +26,8 @@
 * ``required`` defines whether using quotes in string values is required
   (``true``, default) or not (``false``), or only allowed when really needed
   (``only-when-needed``).
+* ``needed-extra-regex`` allows strings matching the given PCRE regex to be
+  quoted even when quoting is only allowed when really needed
```
adrienverge, line 30:
```suggestion
* ``needed-extra-regex`` requires strings matching the given PCRE regex to be
  quoted even when quoting is only allowed when really needed.
```

yamllint/rules/quoted_strings.py
```diff
@@ -63,18 +65,34 @@
...
+
+   the following code snippet would **PASS**:
+   ::
+
+    foo: '%bar%'
+
+   the following code snippet would **FAIL**:
+   ::
+
+    foo: 'bar'
```
adrienverge, line 79: Can you also add:
```
   ::

    foo: %bar%
```
?

leofeyer: Added in dd1db1c12c54a64baa349afed48770ec1f42c6f2.

leofeyer:
> Also in tests, can you make sure _quotes are **required**_ when the regex matches? For example:
> 
> ```python
>         self.check('---\n'
>                    'string1: \'$foo$\'\n'                    # fails
>                    'string2: \'%foo%\'\n'
>                    'string3: $foo$\n'
>                    'string4: %foo%\n',                       # fails
>                    conf, problem1=(2, 10), problem2=(5, 10))
> ```

No, this cannot be tested, as it throws a syntax error: `found character '%' that cannot start any token`

adrienverge:
> No, this cannot be tested, as it throws a syntax error: `found character '%' that cannot start any token`

Sure, but you get the idea, right? The goal is to test that the new option does its job, for example using another regex like `^-.*-$` with `string4: -foo-`.

By the way, if `%` cannot start any token (and creates a syntax error), then checking `needed-extra-regex: ^%.*$` is a bit useless because yamllit would already report this starting `%` as an error? Or maybe I'm missing something?

leofeyer:
> Sure, but you get the idea, right? The goal is to test that the new option does its job, for example using another regex like `^-.*-$` with `string4: -foo-`.

I can try.

> By the way, if `%` cannot start any token (and creates a syntax error), then checking `needed-extra-regex: ^%.*$` is a bit useless because yamllit would already report this starting `%` as an error? Or maybe I'm missing something?

Maybe `%` should be in `START_TOKENS` then? At least the specs list `%` as a directive indicator, which can never start an unquoted string:

- https://yaml.org/spec/1.2/spec.html#id2774058
- https://yaml.org/spec/1.2/spec.html#id2781147

adrienverge:
> Maybe `%` should be in `START_TOKENS` then? At least the specs list `%` as a directive indicator, which can never start an unquoted string:
> 
> * https://yaml.org/spec/1.2/spec.html#id2774058
> * https://yaml.org/spec/1.2/spec.html#id2781147

Interesting! Since `%` is only for directive indicators (at the very beginning of the document), I think adding it to `START_TOKENS` would be valid, but wouldn't have any positive effect.

Anyway, since `%` can never start an unquoted string, maybe it's better to change tests to use another regex? For example `^-.*-$` would be an easy change.

leofeyer:
> For example `^-.*-$` would be an easy change.

Isn't `-` in `START_TOKENS` and will be quoted anyway?

leofeyer:
Also, if `%` can never start an unquoted string, isn't it a bug that yamllint allows it and even says "is incorrectly quoted" if a quoted string starts with `%`? IMHO, this is a bug that should be fixed independently from this PR, which adds a new feature.

adrienverge:
> Isn't `-` in `START_TOKENS` and will be quoted anyway?

Actually `-` can start a string, I think it's valid to write `key: -val` (where `"-val"` is a string). So, `START_TOKENS` is a simplification that doesn't work perfectly. Thanks for pointing this out.
@ruipinge any thoughts? To summarize: with `quoted-strings: {required: only-when-needed}`, yamllint shouldn't report `key: -val` as a problem, because `"-val"` is a valid string that doesn't need quotes.

> Also, if `%` can never start an unquoted string, isn't it a bug that yamllint allows it and even says "is incorrectly quoted" if a quoted string starts with `%`? IMHO, this is a bug that should be fixed independently from this PR (which adds a new feature).

I'm not sure whether it's strictly forbidden to have something like `key: %val`, it would need to read the spec carefully. Anyway I agree with you: let's not mess with `%` in this PR. Let's use something else in tests, for example `^http://`?

leofeyer:
I have adjusted the tests in eb3c857742875d515e13bac29ffd2fbf4bf5218f to use `^http://` instead of `^%`.

adrienverge:
Great, thanks! :+1: 

What about my comment https://github.com/adrienverge/yamllint/pull/246#pullrequestreview-389830596?

leofeyer:
> What about my comment [#246 (review)](https://github.com/adrienverge/yamllint/pull/246#pullrequestreview-389830596)?

This would be a new feature that has nothing to do with my PR, because right now `required: only-when-needed` does not enforce quotes, either.

You can test this yourself by changing `test_quoted_strings.py` line 181 `'string1: foo\n'` to `'string1: *foo\n'` and running the tests. They will pass just fine and there will be no `string value is not quoted with %s quotes` error.

Therefore I suggest to merge the PR as is and to create a new PR to adjust the checks to enforce quotes if a string matches `START_TOKENS` or the new extra regex.

adrienverge:
OK so this feature has 2 parts:
1. Allow quoted strings if they match regex, while `require: only-when-needed` is enforced.
2. Require strings to be quoted if they match regex.

In my opinion part 2 *has* "something to do with your PR", as users who read `require` and `extra-regex` may expect strings that match to be quoted (that was my case, for example).
Moreover, I prefer adding the whole feature at once than just half of it (merging this PR now), because it helps choosing adapted keywords (e.g. `needed-extra-regex` may not be ideal given the 2 points above, maybe `extra-allowed` and `extra-required` would be better?) and detect problems early.


Other things I don't agree with:

- Your example `string1: *foo` isn't valid because `*foo` is a YAML alias (see https://yaml.org/spec/1.2/spec.html#id2786196), not a string.

- Tests you added don't check `needed-extra-regex` with other "regular" value types (see the rest of `test_quoted_strings.py`), and it's a problem because it could create unexpected behaviors (it's frequent).

- Linting don't pass on your PR (check out CI results).

- A user could need to give multiple regexes, e.g. `^http://` and `^ftp://`: maybe a list of regexes would be better?

For these reasons (and to save time) I will implement this in another PR.

ruipinge:
> Actually `-` can start a string, I think it's valid to write `key: -val` (where `"-val"` is a string). So, `START_TOKENS` is a simplification that doesn't work perfectly. Thanks for pointing this out.
> @ruipinge any thoughts? To summarize: with `quoted-strings: {required: only-when-needed}`, yamllint shouldn't report `key: -val` as a problem, because `"-val"` is a valid string that doesn't need quotes.

My hunch goes with adding *any* possible indicator character to the list of `START_TOKENS`. Actually, I missed not only `%`, but also `| `, `'`, `"`, and `>`. All of these serve, in their own way, as control characters, so we would make the yaml files safer by "imposing" any string starting with it be quoted.

ruipinge:
I know that development is already ongoing, so I'm sorry to be late for the party.

Although, after reading the two threads and noticed that what motivated this PR and #246 was actually the fact that 

> Symfony requires placeholders such as %kernel.project_dir% or %env(DATABASE_URL)% to be quoted

I wanted to give my two cents. If we came to agree that the `%` was missing as a `START_TOKEN` (and actually Symfony parser was doing good), would this feature continue to make sense? Wouldn't it become a feature without a use case?

adrienverge:
@ruipinge thanks for your feedback.

I'm still OK to add `%` to `START_TOKENS`, I'd accept a PR for that.

But `START_TOKENS` has proven to be partially broken: looking at the first character of a string is not enough. For example:
- `-----` doesn't need quotes, but redundant quotes would not be detected,
- `:char` doesn't need quotes, but redundant quotes would not be detected,
- `"'foo' == 'bar'"` needs quotes, but `only-when-needed` complains if it has,
- `"hello: world"` needs quotes, but `only-when-needed` complains if it has.

By the way this is already a problem for users: an issue was open at https://github.com/adrienverge/yamllint/issues/242. Let's continue this topic there, but we may have to drop `START_TOKENS`.

> I wanted to give my two cents. If we came to agree that the `%` was missing as a `START_TOKEN` (and actually Symfony parser was doing good), would this feature continue to make sense? Wouldn't it become a feature without a use case?

Good point. You're probably right *today*, but since [the feature is already developed](https://github.com/adrienverge/yamllint/pull/252) and the use case may arrive in the future (some users may want some types of strings to be quoted, for visual comfort?), I'd be for adding it. What do you think?
If we do, I want these new options to be future-proof and correctly labeled, that's why your feedback would be very welcome :)

adrienverge:
Closing because https://github.com/adrienverge/yamllint/pull/252 and https://github.com/adrienverge/yamllint/pull/254 fixed the original problem.

Issue #158, opened 2019-01-14
false positive truthy failure on travis files


ssbarnea:
It seems that yamllint chokes complaining about not using truthy values for travis **keys**.

Just to be clear: this is about **keys not values with name "on".**. When used as a key, on have a total different sense than the "on" as a value, which I find acceptable to trigger a warning.

ssbarnea:
Thanks for the update, now we need to find a way to fix this because having it documented does not make the user experience better. So far the only place I found to be conflicting is travis.yaml and to be honest I cannot find any reasom for blaming its syntax here.

Maybe we need an option to look for truthy only for leaf-keys or to disable it for travis files. I am open to suggestions. 

Fixing this is not an urgency as I disabled the truthy feature completely for the moment on that small project. Still, we should look for a way to address this in the default tool behaviour, so the user experience will be good and he will not encounter this false-positive.

ssbarnea:
My mistake to not give an example, https://github.com/pycontribs/tendo/blob/master/.travis.yml#L78

We are lucky the the "on" condition (which is similar to `when` on ansible) is not very common and is not present in default travis templates. Still, is a core feature which is very old and established so I doubt they will change the file format to pass yamllint-ing.

I don't know what is the right approach to take here but I know how to describe the desired user experience: when user enables yamllint they should not get a false-positive on a travis.yaml file.

adrienverge:
What about a new option for the `truthy` rule? Something like `check-keys: true | false` or similar?

@pdericson any thoughts?

jayvdb:
What about a special case for `.travis.yml` , or having schemas which define key names which could then be whitelisted if they appear in the correct place in the yaml.

jb-sean:
This issue is not limited to `.travis.yml` configs. The new GitHub Actions yaml spec has a required `on` key: https://help.github.com/en/articles/workflow-syntax-for-github-actions#on
E.g.,
```yaml
name: Pull request based on master
on:
  pull_request:
    branches:
      - master
# ...
```
yamllint is throwing `2:1       warning  truthy value should be one of [false, true]  (truthy)`.

adrienverge:
Hi,

While waiting for someone to contribute a new `check-keys` option or something equivalent, an option is to disable the `truthy` rule in yamllint config:
```yaml
---
extends: default
rules:
  truthy: disable
```
... or allow `on`:
```yaml
---
extends: default
rules:
  truthy:
    allowed-values: ['true', 'false', 'on']
[...]

PR #247, opened 2020-04-04
truthy rule: add `check-keys` option

ilyam8:
Fixes: #158

This PR adds `check-keys` option to the truthy rule.

If `check-keys` option is disabled the rule applies only to the values. Default is enabled.

### DIFF
--- a/tests/rules/test_truthy.py
+++ b/tests/rules/test_truthy.py
@@ -114,3 +114,29 @@ class TruthyTestCase(RuleTestCase):
                    'boolean5: !!bool off\n'
                    'boolean6: !!bool NO\n',
                    conf)
+
+    def test_check_keys_disabled(self):
+        conf = ('truthy:\n'
+                '  allowed-values: []\n'
+                '  check-keys: false\n'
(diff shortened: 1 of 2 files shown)


ilyam8:
> Илья!

😲 ☺️

tests/rules/test_truthy.py
```diff
@@ -114,3 +114,29 @@ def test_explicit_types(self):
...
+                   'true: 0\n'
+                   'FALSE: 0\n'
+                   'False: 0\n'
+                   'false: 0\n'
+                   'ON: 0\n'
+                   'On: 0\n'
+                   'on: 0\n'
+                   'OFF: 0\n'
+                   'Off: 0\n'
+                   'off: 0\n',
```
ilyam8, line 141: Are you sure nested case is not needed?

```yaml
YES:
  Yes:
    yes:
      on: 0
```

adrienverge: Oops, пожалуйста, I thought it was a duplicate so I removed it. I'll push a commit to restore that. Thanks!

PR #252, opened 2020-04-10
quoted-strings: Add extra-required and extra-allowed options

adrienverge:
### config: Allow rules to validate their configuration

---

### config: Allow generic types inside lists

For example it's possible to define a conf like:
```yaml
rule:
  foo: [str],
  bar: [int, bool, 'magic'],
```

---

### quoted-strings: Add options extra-required and extra-allowed

Add ability to:
- require strings to be quoted if they match a pattern (PCRE regex)
- allow quoted strings if they match a pattern, while `require:
  only-when-needed` is enforced.

Co-Authored-By: Leo Feyer (https://github.com/adrienverge/yamllint/pull/246)

### DIFF
--- a/tests/rules/test_quoted_strings.py
+++ b/tests/rules/test_quoted_strings.py
@@ -16,6 +16,8 @@
 
 from tests.common import RuleTestCase
 
+from yamllint import config
+
 
 class QuotedTestCase(RuleTestCase):
     rule_id = 'quoted-strings'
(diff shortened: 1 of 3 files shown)


adrienverge:
@leofeyer @ruipinge feel free to comment on the implementation. Are there any use-cases uncovered?

To summarize:
- With ``quoted-strings: {required: false, extra-required: [^http://, ^ftp://]}``

  - the following code snippet would **PASS**:

    ```yaml
    - localhost
    - "localhost"
    - "http://localhost"
    - "ftp://localhost"
    ```

  - the following code snippet would **FAIL**:

    ```yaml
    - http://localhost
    - ftp://localhost
    ```

- With ``quoted-strings: {required: only-when-needed, extra-allowed: [^http://, ^ftp://], extra-required: [QUOTED]}``

  - the following code snippet would **PASS**:

    ```yaml
    - localhost
    - "http://localhost"
    - "ftp://localhost"
    - "this is a string that needs to be QUOTED"
    ```

  - the following code snippet would **FAIL**:

    ```yaml
    - "localhost"
    - this is a string that needs to be QUOTED

leofeyer:
🎉

Thank you @adrienverge.

adrienverge:
You're welcome, and thanks for your help on this.

For info, if your use case is just strings starting with `%`, it should be fixed by https://github.com/adrienverge/yamllint/pull/254. You don't have to change your configuration.

PR #255, opened 2020-04-22
Option to specify config file with environment variable

wbadart:
Hey team-

This PR instructs the CLI to check the `YAMLLINTRC` environment variable for a global config path. For _most_ intents and purposes, this can be achieved with the `-c` flag, but I found a corner case where the environment variable is useful: I'd like to point my editor and my CI server at the same config file, and it's much easier to specify the variable once in something like a `.env` file than tweaking how both systems invoke `yamllint` individually.

I'm not married to the name "`YAMLLINTRC`"; please let me know if there's a better name (I've been working with pylint lately, so I think I subconsciously borrowed their convention). Also, please let me know if this is PR is the right place to make the corresponding update to the docs.

Thanks!

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -24,6 +24,7 @@ import os
 import pty
 import shutil
 import sys
+import tempfile
 import unittest
 
 from tests.common import build_temp_workspace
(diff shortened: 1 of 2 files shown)


wbadart:
Oh hold up, just realized this might be redundant with #119....

tests/test_cli.py
```diff
@@ -285,6 +285,26 @@ def test_run_with_user_global_config_file(self):
             cli.run((os.path.join(self.wd, 'a.yaml'), ))
         self.assertEqual(ctx.returncode, 1)
 
+    def test_run_with_user_yamllintrc_config_file(self):
+        config = os.path.join(self.wd, 'fake-local-config')
+        self.addCleanup(os.remove, config)
```
adrienverge, line 290: In other tests we use:
https://github.com/adrienverge/yamllint/blob/a54cbce1b62a37b76eb07479a16dd761d476990c/tests/test_config.py#L216

Maybe it would be better to stay consistent?

wbadart: 100%

wbadart:
I took your advice and grepped through some of the other tests to see how they're structured. 8657838 should make the new test here much more consistent with those!

Issue #171, opened 2019-03-25
Empty lines in comments not handled sensibly by require-starting-space


g1ps:
**Problem**
Comments like this:
```
[Line 32 of yml file]
      # - name: some task
      #   import_role:
      #     name: some-role
      #
      # - name: some task
      #   import_role:
      #     name: some-role
      #
      # - name: some task
      #   import_role:
      #     name: some-role
      #
      # - name: some task
      #   import_role:
      #     name: some-role
```
are flagged as needing a leading space after the hash.

```
$ yamllint ./zz.yml
./zz.yml
  36:8      warning  missing starting space in comment  (comments)
  40:8      warning  missing starting space in comment  (comments)
  44:8      warning  missing starting space in comment  (comments)
```

**Request**
That makes no sense as it's likely to flag various valid comments. The rule should only be applied to lines with text following the hash. It should not be applied to cases where a newline follows the hash.

Thanks.

adrienverge:
The problems you describe don't happen on my side.

I suspect interference with potential strange line-endings on Windows. Do you have the `new-lines` rule enabled? Do your files end with universal (`\n`) newlines?

g1ps:
Note that neither:
```
  # yamllint: disable
...
  # yamllint: enable
```
nor
```
  # yamllint: disable rule:comments
...
  # yamllint: enable rule:comments
```
nor
```
# yamllint: disable rule:require-starting-space
...
# yamllint: enable rule:require-starting-space
```
work. In fact I can't disable rules at all for any I've tested so far. `yamllint` even picks up hte indentation of the disabling comments if they're not indented like the content, which seems dubious. Again the docs aren't detailed enough to tell me whether I can disable a rule parameter this way (I should certainly be able to) or not.

g1ps:
> I suspect interference with potential strange line-endings on Windows. Do you have the `new-lines` rule enabled? Do your files end with universal (`\n`) newlines?

No, it's disabled since we might have either. They end with Windows line endings here.

Thanks.

g1ps:
You're right. I ran dos2unix on a file which showed these problems and they went away. This will be a problem for a lot of users trying to run this way to avoid pain with Ansible. Even Linux users might validly have files with Windows line endings on their machines so I hope you can add checking for both types.

Disabling rules works now, too, except that indentation applies to the comments so it expects this:
```
  # yamllint disable rule:indentation
  - name: 'run some task'
  import_role:
    name: some-role
[...]

PR #257, opened 2020-04-28
Bugfix: GH Issue #171 / CRLF wih require-starting-space.

bsolomon1124:
Pound-signs followed by a lone CRLF should not
raise if require-starting-space is specified and newlines are dos or disabled.

Adds three new passing test cases in addition to the fix:

```
$ coverage run --source=yamllint setup.py test -s tests.rules.test_comments.CommentsTestCase
test_empty_comment_crlf_disabled_newlines (tests.rules.test_comments.CommentsTestCase) ... ok
test_empty_comment_crlf_dos_newlines (tests.rules.test_comments.CommentsTestCase) ... ok
```

### DIFF
--- a/yamllint/rules/comments.py
+++ b/yamllint/rules/comments.py
@@ -97,7 +97,9 @@ def check(conf, comment):
                     comment.column_no == 1 and
                     re.match(r'^!\S', comment.buffer[text_start:])):
                 return
-            elif comment.buffer[text_start] not in (' ', '\n', '\0'):
+            # We can test for both \r and \r\n just by checking first char
+            # \r itself is a valid newline on some older OS.
+            elif comment.buffer[text_start] not in {' ', '\n', '\r', '\x00'}:
                 column = comment.column_no + text_start - comment.pointer
                 yield LintProblem(comment.line_no,
                                   column,
--- a/tests/rules/test_comments.py
+++ b/tests/rules/test_comments.py
@@ -186,6 +186,27 @@ class CommentsTestCase(RuleTestCase):
                    'inline: comment  #\n'
                    'foo: bar\n', conf)
 
+    def test_empty_comment_crlf_dos_newlines(self):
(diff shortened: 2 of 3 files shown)


bsolomon1124:
This PR is ready for review.

bsolomon1124:
I also tested this forked/branch on the real-world repo where ` missing starting space in comment` was previously being raised and the issue was resolved there also.

yamllint/rules/comments.py
```diff
@@ -97,7 +97,9 @@ def check(conf, comment):
                     comment.column_no == 1 and
                     re.match(r'^!\S', comment.buffer[text_start:])):
                 return
-            elif comment.buffer[text_start] not in (' ', '\n', '\0'):
+            # We can test for both \r and \r\n just by checking first char
+            # \r itself is a valid newline on some older OS.
```
adrienverge, line 101: OK, the problem becomes simpler then.

This does no handle examples like:

    ---\n
    #\rcomment\n

and:

    ---\r\n
    #\r[end of file]

... but I guess it's acceptable to ignore those.

yamllint/rules/comments.py
```diff
@@ -97,7 +97,9 @@ def check(conf, comment):
...
+            elif comment.buffer[text_start] not in {' ', '\n', '\r', '\x00'}:
```
adrienverge, line 102: I'm not sure the O(1) search for 4 one-byte elements compensate the overhead of building a set + hashings required, but I admit it's nicer-looking now :wink:

bsolomon1124: You might very well be right on that one 😏

yamllint/rules/comments.py
bsolomon1124: > but I guess it's acceptable to ignore those.

Agreed. I don't think mixing of newline types in those corner cases would be something you'd want yamllint to be okay with.

Issue #261, opened 2020-05-02
ReadError exception fails to mention the file generating the error.


ssbarnea:
While the error causing the exception seems to be caused by https://github.com/adrienverge/yamllint/issues/218 what makes it much worse is the it produces an output which makes impossible to identify the filename causing the problem:

```
$ yamllint -f parsable .                                                                                                                                                  Traceback (most recent call last):
File "/Users/ssbarnea/.pyenv/versions/3.7.6/bin/yamllint", line 10, in <module>
sys.exit(run())
File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/site-packages/yamllint/cli.py", line 185, in run
no_warn=args.no_warnings)
File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/site-packages/yamllint/cli.py", line 91, in show_problems
for problem in problems:
File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/site-packages/yamllint/linter.py", line 198, in _run
syntax_error = get_syntax_error(buffer)
File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/site-packages/yamllint/linter.py", line 179, in get_syntax_error
list(yaml.parse(buffer, Loader=yaml.BaseLoader))
File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/site-packages/yaml/__init__.py", line 73, in parse
loader = Loader(stream)
File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/site-packages/yaml/loader.py", line 14, in __init__
Reader.__init__(self, stream)
File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/site-packages/yaml/reader.py", line 74, in __init__
self.check_printable(stream)
File "/Users/ssbarnea/.pyenv/versions/3.7.6/lib/python3.7/site-packages/yaml/reader.py", line 144, in check_printable
'unicode', "special characters are not allowed")
yaml.reader.ReaderError: unacceptable character #x001b: special characters are not allowed
in "<unicode string>", position 60
FAIL: 1
```

At minimum, the filename must be mentioned.

PR #262, opened 2020-05-02
Fix confusing ReadError traceback

ssbarnea:
Avoids confusing tracebacks when ReadErrors exceptions are raised
from the yaml reader.

Now we do print the filename that failed to load and stop processing
in order to avoid additional exceptions.

Also adds tests to prevent future regressions.

Example output after this change:
```
bad-file.yaml:4:36: [error] syntax error: special characters are not allowed (syntax)
```

Fixes: #261


ssbarnea:
@adrienverge Does this look ok?

adrienverge:
Wow, reviving a PR after one day is a bit too much. I have concerns about this commit, I'll take the time to think about it. Please be patient.

yamllint/linter.py
```diff
@@ -183,6 +183,15 @@ def get_syntax_error(buffer):
                               'syntax error: ' + e.problem + ' (syntax)')
         problem.level = 'error'
         return problem
+    except yaml.reader.ReaderError as e:
+        # we need to convert position into line+column
+        lines = buffer[:e.position].splitlines()
```
webknjaz, line 188: What if the buffer is big and the error is at the end?
I think something like `line_num = buffer.count('\n', 0, e.position)` and `col_num = e.position - buffer.rindex('\n', 0, e.position)` would be more memory-efficient.

tests/test_syntax_errors.py
```diff
@@ -92,3 +92,7 @@ def test_sets(self):
                    '    a:\n'
                    '      set\n'
                    '...\n', None)
+
+    def test_non_printable_characters(self):
+        self.check('output: \x1b\n',
+                   None, problem=(1, 9))
```
adrienverge, line 98: Please add more tests. At the least I can think of:
- invalid character in the first byte,
- invalid character in the last byte,
- invalid character not on line 1,
- multiple invalid characters,
- test with other rules enabled.

yamllint/linter.py
adrienverge: @webknjaz very good point, and your solution looks good. The column one needs a small adaptation because it wouldn't work if the invalid char is on the first line.
```suggestion
        line = buffer.count('\n', 0, e.position)
        col = e.position - (buffer.rindex('\n', 0, e.position) if line else 0)
        problem = LintProblem(line + 1, col + 1,

```

tests/test_syntax_errors.py
ssbarnea: Sorry for the long delay, somehow this patch got lost and only rediscovered it yesterday. I added the tests.

Still, I am unsure if the coverage failure is a bug in coverage because these new tests would cover that piece of code. Also there are no instructions about running coverage locally for the project so I hope you may be able to provide few hints here.

tests/test_syntax_errors.py
```diff
@@ -92,3 +92,28 @@ def test_sets(self):
                    '    a:\n'
                    '      set\n'
                    '...\n', None)
+
+    def test_invalid_char_last(self):
+        self.check('---\nkey: value\t\n',
```
adrienverge, line 97: Can you use the more-readable format that is used everywhere is tests?
```suggestion
        self.check('---\n'
                   'key: value\t\n',
```

tests/test_syntax_errors.py
```diff
@@ -92,3 +92,28 @@ def test_sets(self):
...
+        self.check('---\n\tkey: value\n',
+                   None, problem=(2, 1))
+
+    def test_invalid_char_on_second_line(self):
+        self.check('---\nfoo: bar\n\toutput: foo\n',
+                   None, problem=(3, 1))
+
+    def test_invalid_chars(self):
+        # We expect to identify only the first syntax error as this kind of
+        # error is fatal (stops parsing)
```
adrienverge, line 110: :+1:

adrienverge (commented):
Why did you remove this test? Can you reintroduce it?
```python
    def test_non_printable_characters(self):
        self.check('output: \x1b\n',
                   None, problem=(1, 9))
```

PR #264, opened 2020-05-02
Added flake8 ignores

ssbarnea:
Avoid W503/W504 with current code as the current code not compliant
and they are contradictory.


ssbarnea:
@adrienverge ^ thanks.

PR #280, opened 2020-07-08
Add "locale" config to key-ordering rule

wolfgangwalther:
Support sorting by locale with strcoll(). Properly handle case and accents.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -189,3 +189,22 @@ Here is a more complex example:
      ignore: |
        *.ignore-trailing-spaces.yaml
        ascii-art/*
+
+Setting the locale
+------------------
+
+It is possible to set the ``locale`` option globally. This is passed to Python's
+`locale.setlocale
+<https://docs.python.org/3/library/locale.html#locale.setlocale>`_,
+so an empty string ``""`` will use the system default locale, while e.g.
+``"en_US.UTF-8"`` will use that. If unset, the default is ``"C.UTF-8"``.
+
+Currently this only affects the ``key-ordering`` rule. The default will order
+by Unicode code point number, while other locales will sort case and accents
+properly as well.
+
+.. code-block:: yaml
+
+ extends: default
+
+ locale: en_US.UTF-8
--- a/tests/rules/test_key_ordering.py
+++ b/tests/rules/test_key_ordering.py
@@ -14,6 +14,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import locale
+
 from tests.common import RuleTestCase
 
 
@@ -103,10 +105,6 @@ class KeyOrderingTestCase(RuleTestCase):
                    'haïr: true\n'
                    'hais: true\n', conf,
                    problem=(3, 1))
-        self.check('---\n'
-                   'haïr: true\n'
-                   'hais: true\n', conf,
-                   problem=(3, 1))
 
     def test_key_tokens_in_flow_sequences(self):
         conf = 'key-ordering: enable'
@@ -114,3 +112,39 @@ class KeyOrderingTestCase(RuleTestCase):
                    '[\n'
                    '  key: value, mappings, in, flow: sequence\n'
                    ']\n', conf)
+
+    def test_locale_case(self):
+        self.addCleanup(locale.setlocale, locale.LC_ALL, 'C.UTF-8')
+        try:
+            locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
+        except locale.Error:
+            self.skipTest('locale en_US.UTF-8 not available')
+        conf = ('key-ordering: enable')
+        self.check('---\n'
+                   't-shirt: 1\n'
+                   'T-shirt: 2\n'
+                   't-shirts: 3\n'
+                   'T-shirts: 4\n', conf)
+        self.check('---\n'
+                   't-shirt: 1\n'
+                   't-shirts: 2\n'
+                   'T-shirt: 3\n'
+                   'T-shirts: 4\n', conf,
+                   problem=(4, 1))
+
+    def test_locale_accents(self):
+        self.addCleanup(locale.setlocale, locale.LC_ALL, 'C.UTF-8')
+        try:
+            locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
+        except locale.Error:
+            self.skipTest('locale en_US.UTF-8 not available')
+        conf = ('key-ordering: enable')
+        self.check('---\n'
+                   'hair: true\n'
+                   'haïr: true\n'
+                   'hais: true\n'
+                   'haïssable: true\n', conf)
+        self.check('---\n'
+                   'hais: true\n'
+                   'haïr: true\n', conf,
+                   problem=(3, 1))
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -95,6 +95,13 @@ class CommandLineTestCase(unittest.TestCase):
             # dos line endings yaml
             'dos.yml': '---\r\n'
                        'dos: true',
+            # different key-ordering by locale
+            'c.yaml': '---\n'
+                      'A: true\n'
+                      'a: true',
+            'en.yaml': '---\n'
+                       'a: true\n'
+                       'A: true'
         })
 
     @classmethod
@@ -108,8 +115,10 @@ class CommandLineTestCase(unittest.TestCase):
         self.assertEqual(
             sorted(cli.find_files_recursively([self.wd], conf)),
             [os.path.join(self.wd, 'a.yaml'),
+             os.path.join(self.wd, 'c.yaml'),
              os.path.join(self.wd, 'dos.yml'),
              os.path.join(self.wd, 'empty.yml'),
+             os.path.join(self.wd, 'en.yaml'),
              os.path.join(self.wd, 's/s/s/s/s/s/s/s/s/s/s/s/s/s/s/file.yaml'),
              os.path.join(self.wd, 'sub/directory.yaml/empty.yml'),
              os.path.join(self.wd, 'sub/ok.yaml'),
@@ -146,6 +155,8 @@ class CommandLineTestCase(unittest.TestCase):
         self.assertEqual(
             sorted(cli.find_files_recursively([self.wd], conf)),
             [os.path.join(self.wd, 'a.yaml'),
+             os.path.join(self.wd, 'c.yaml'),
+             os.path.join(self.wd, 'en.yaml'),
              os.path.join(self.wd, 's/s/s/s/s/s/s/s/s/s/s/s/s/s/s/file.yaml'),
              os.path.join(self.wd, 'sub/ok.yaml'),
              os.path.join(self.wd, 'warn.yaml')]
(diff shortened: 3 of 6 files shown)


wolfgangwalther:
I don't think the "breaking" tests are related to my changes. The linter seems to be complaining about `docs/conf.py` - haven't touched that.

tests/rules/test_key_ordering.py
```diff
@@ -103,14 +103,38 @@ def test_accents(self):
                    'haïr: true\n'
                    'hais: true\n', conf,
                    problem=(3, 1))
-        self.check('---\n'
-                   'haïr: true\n'
-                   'hais: true\n', conf,
-                   problem=(3, 1))
```
adrienverge, line 109: Could you keep this one? It makes sure an error is reported when `locale` isn't provided.

yamllint/rules/key_ordering.py
```diff
@@ -16,8 +16,12 @@
 
 """
 Use this rule to enforce alphabetical ordering of keys in mappings. The sorting
-order uses the Unicode code point number. As a result, the ordering is
-case-sensitive and not accent-friendly (see examples below).
+order uses the Unicode code point number as a default. As a result, the
+ordering is case-sensitive and not accent-friendly (see examples below).
+This can be changed by setting "locale". This is passed to pythons
```
adrienverge, line 21: For clarity I propose

    This can be changed by setting the ``locale`` option.

yamllint/rules/key_ordering.py
```diff
@@ -86,6 +134,8 @@ def check(conf, token, prev, next, nextnext, context):
     if 'stack' not in context:
         context['stack'] = []
 
+    setlocale(LC_COLLATE, conf['locale'])
```
adrienverge, line 137: With this, running `yamllint` on a directory with 4 or 5 YAML files calls `setlocale(LC_COLLATE, ...)` 657 times. This is a lot.

Moreover, calling `setlocale()` could theoretically affect other parts of yamllint.

→ Is there a way to call an equivalent of `strcoll()` by passing the `conf['locale']` directly? (I quickly looked at [the `locale` docs](https://docs.python.org/3/library/locale.html#locale.setlocale), haven't found.)
→ If calling `setlocale()` is really needed, can we do it only once?

yamllint/rules/key_ordering.py
```diff
@@ -63,15 +67,59 @@
...
+    - t-shirt: 1
+      t-shirts: 2
+      T-shirt: 3
+      T-shirts: 4
+
+   the following code snippet would **FAIL**:
+   ::
+
+    - hais: true
+      haïr: true
```
adrienverge, line 109: I'm not sure we need all these, it could burden the documentation for users. What about just keeping the accented examples?

yamllint/rules/key_ordering.py
```diff
@@ -16,8 +16,12 @@
...
 """
 Use this rule to enforce alphabetical ordering of keys in mappings. The sorting
-order uses the Unicode code point number. As a result, the ordering is
-case-sensitive and not accent-friendly (see examples below).
+order uses the Unicode code point number as a default. As a result, the
+ordering is case-sensitive and not accent-friendly (see examples below).
+This can be changed by setting "locale". This is passed to pythons
+`locale.setlocale<https://docs.python.org/3/library/locale.html#locale.setlocale>`,
+so an empty string "" will use the system default locale, while e.g.
+"en_US.UTF-8" will use that. This allows to sort case and accents properly.
```
adrienverge, line 24: Will look nicer with

    "" → ``""``
    "en_US.UTF-8" → ``"en_US.UTF-8"``

:+1:

tests/rules/test_key_ordering.py
wolfgangwalther: Those 4 lines were identically repeated. So that test is still in there, just once now and not twice. Should have left a comment here, sorry ;)

yamllint/rules/key_ordering.py
wolfgangwalther: I think the case example should be in there somehow. That's what initially motivated my changes - and the result is different from before now in two ways, because capital letters are now sorted **after** their lower counterparts. They were sorted **before** all lower letters before. So not only case-sensitive, but also "the other way around".

I think we can just keep the first sample (reduced to the t- and h- lines) that passes. Everyone can guess the failing lines themselves with more extended examples above.

What do you think?

yamllint/rules/key_ordering.py
wolfgangwalther: 657 times is a lot - I hope you did not count that ;)

We can't avoid calling `setlocale()`, because `strcoll()` only works with that. The only other way I found to do that kind of sorting would be to use `PyICU`. But adding an external dependency with c++ libs as dependencies just for that part seems like overkill to me.
 
In theory this setting could affect other parts of yamllint, but the docs say:

>locale.LC_COLLATE
>   Locale category for sorting strings. The functions strcoll() and strxfrm() of the locale module are affected.

If only those two functions rely on this setting - and they are not used anywhere else (they are not right now), then nothing should be affected.

Calling setlocale only once seems plausible. I have not digged in deeper in yamllints source code, but just copy-pasted my way through so far ;), so maybe you can give me a hint: I guess there is some place where we can set global options instead of rule-specific?

We would just need to make `locale` global and call it only when reading the config file. Which seems better in regard to the "what if we use strcoll somewhere else as well - will that not affect us there as well" problem.

wolfgangwalther:
I think making `locale` a global option is the right way to go. I identified three possible ways to do it:

1. make it take the system default from the environment variables all the time and only use that for configuration. I decided not to do that, because that would be a breaking change.

2. have a global `locale` option and use `setlocale()` in `linter.run`. This has two drawbacks:

- would need to change the `RuleTestCase.check()` to support a global config as well - I guess this would require to change all test-cases that use it.

- if somebody imports `linter` into their own project to use yamllint as a lib, `setlocale` would change the locale of the whole project to `C.UTF-8`. This might not be desired for the remaining code - in this case, yamllint as a lib should just respect the locale setting of the project it is imported to.

3. have a global `locale` option and use `setlocale()` in `cli.run`. This is what I went for.

To be able to test this, I added another test case for cli.run, whether that sets the locale properly. In the `key-ordering` tests, I just set the locale manually then, to test whether the rule respects that setting.

I think this solution is better then the one before.

yamllint/config.py
```diff
@@ -111,6 +113,9 @@ def parse(self, raw_content):
             self.yaml_files = pathspec.PathSpec.from_lines('gitwildmatch',
                                                            conf['yaml-files'])
 
+        if 'locale' in conf:
+            self.locale = conf['locale']
```
adrienverge, line 117: Can you add a small check to `raise YamlLintConfigError(...)` if `conf['locale']` is not a string?

yamllint/rules/key_ordering.py
```diff
@@ -63,8 +65,24 @@
...
+   ::
+
+    - t-shirt: 1
+      T-shirt: 2
+      t-shirts: 3
+      T-shirts: 4
+    - hair: true
+      haïr: true
+      hais: true
+      haïssable: true
```
adrienverge, line 81: :+1: 

(agree with your previous message about `t`/`T`)

docs/configuration.rst
```diff
@@ -189,3 +189,22 @@ Here is a more complex example:
      ignore: |
        *.ignore-trailing-spaces.yaml
        ascii-art/*
+
+Setting the locale
+------------------
+
+It is possible to set the ``locale`` option globally. This is passed to pythons
```
adrienverge, line 196: `pythons` → `Python's`?

tests/test_cli.py
```diff
@@ -305,6 +305,16 @@ def test_run_with_user_yamllint_config_file_in_env(self):
                 cli.run((os.path.join(self.wd, 'a.yaml'), ))
             self.assertEqual(ctx.returncode, 1)
 
+    def test_run_with_locale(self):
+        loc = locale.getlocale(locale.LC_ALL)
+        self.assertNotEqual(loc, 'en_US.UTF-8')
```
adrienverge, line 310: My understanding of this line is that you want to make sure `cli.run(('-d', 'locale: en_US.UTF-8',` actually changes something. But for our American friends, it may result in crashes when trying to package yamllint (if their system default is `en_US.UTF-8`)...

Instead, maybe we could drop this line, but repeat the test code with a second locale, e.g. `de_DE`?

wolfgangwalther: I'm not sure that is actually going to be a problem (because the system default is only used, when explicitly setting `locale: ""` and the default should be `C.UTF-8` as set in config.py), but that line is faulty anyways (`getlocale` returns a tuple instead of a string, see the `assertEqual` below).

Your suggestion to repeat the test is good, I guess we would be on the safe side. I choose `en_US.UTF-8` under the assumption that this locale should be installed in most systems. I guess changing back to `C.UTF-8` instead of `de_DE` for the 2nd test is better, because the tests would not depend on `de_DE` being available?

wolfgangwalther: So I played around a little bit - and that whole test-case is completely broken right now. `getlocale` just does not return expected values at all.

I'll probably just add a more implicit test here - by linting some basic yaml file with the key-ordering rule and a locale setting to see it having an effect.

adrienverge: Interesting! OK, looking forward to see that.

adrienverge: PS: For the record, I remember having packaging problems because `en_US.UTF-8` or `C.UTF-8` were not available on some systems (CentOS 6 or 7 I think).

wolfgangwalther: So I changed the test to actually run some key-ordering rules against files and look at the return code to see whether the `locale` option had any effect.

Those tests are skipped if the `en_US.UTF-8` locale is not present on the system - same for the key-ordering locale tests.

But since `C.UTF-8` is set as the default, I did not do any checks here. Tests will fail (probably everywhere?) when `C.UTF-8` is not available.

tests/test_cli.py
```diff
@@ -363,15 +411,6 @@ def test_run_empty_file(self):
...
 
-        # Make sure the default localization conditions on this "system"
-        # support UTF-8 encoding.
-        loc = locale.getlocale()
-        try:
-            locale.setlocale(locale.LC_ALL, 'C.UTF-8')
-        except locale.Error:
-            locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
-        self.addCleanup(locale.setlocale, locale.LC_ALL, loc)
-
```
wolfgangwalther, line 374: So because we are using `setlocale` in `cli.run` with `C.UTF-8` as the default, those lines are not needed anymore. The command will fail if `C.UTF-8` is not available - and UTF-8 is enabled for sure. So this test should be good without those lines.

Also note that `getlocale()` + the cleanup don't do what you expect them to do. `getlocale()` does not return `LC_ALL`, but `LC_CTYPE`, I think, and the cleanup could end up setting the locale to something different as it was before.

tests/test_cli.py
adrienverge: That seems reasonable. I hope there aren't any systems where `C.UTF-8` is not available.

adrienverge:
Released in version 1.24.0.

adrienverge:
I reverted this commit on `master` and released 1.24.1, because it broke yamllint (and continuous integrations) for different users and companies. I hope it will just be temporary and we can integrate this again!

wolfgangwalther:
Well, I guess that answered your "question":

> I hope there aren't any systems where C.UTF-8 is not available.

Should be fixed in #288

PR #283, opened 2020-07-10
Do no match directories that look like YAML files

### LINKED ISSUE #279: Bug: a directory with name *.yml makes yamllint to lint all containing files despite their type
Give the following tree structure:
/test.yml
/generated/products.yml/products.sql

When executing yamllint, it scans the file products.sql, which is not yaml, raising an error.

jsok:
Fixes #279

### DIFF
--- a/yamllint/config.py
+++ b/yamllint/config.py
@@ -46,7 +46,7 @@ class YamlLintConfig(object):
         return self.ignore and self.ignore.match_file(filepath)
 
     def is_yaml_file(self, filepath):
-        return self.yaml_files.match_file(filepath)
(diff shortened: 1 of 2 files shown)


yamllint/config.py
```diff
@@ -46,7 +46,7 @@ def is_file_ignored(self, filepath):
         return self.ignore and self.ignore.match_file(filepath)
 
     def is_yaml_file(self, filepath):
-        return self.yaml_files.match_file(filepath)
+        return self.yaml_files.match_file(os.path.basename(filepath))
```
adrienverge, line 49: A 18-char fix. Nice.

Issue #285, opened 2020-07-15
running yamllint throws unsupported locale error


GerritJanYourhosting:
Just ran into an issue with the 1.24 release of yamllint with the error below

https://github.com/adrienverge/yamllint/pull/280#discussion_r454926661
` Traceback (most recent call last):
   File "/usr/bin/yamllint", line 11, in <module>
     sys.exit(run())
   File "/usr/lib/python2.7/site-packages/yamllint/cli.py", line 179, in run
     locale.setlocale(locale.LC_ALL, conf.locale)
   File "/usr/lib64/python2.7/locale.py", line 547, in setlocale
     return _setlocale(category, locale)
 locale.Error: unsupported locale setting`

This happends within a docker container running `centos:7`

adrienverge:
@GerritJanYourhosting thanks for reporting this.
@wolfgangwalther this is probably due to https://github.com/adrienverge/yamllint/pull/280. What do you think?

madrover:
Hi
I'm getting the same issue, see:

```
$ uname -a
Darwin miquel-mac.binihome.local 19.5.0 Darwin Kernel Version 19.5.0: Tue May 26 20:41:44 PDT 2020; root:xnu-6153.121.2~2/RELEASE_X86_64 x86_64
$ python3 --version
Python 3.8.3
$ export | grep LC_
LC_ALL=en_US.UTF-8
LC_CTYPE=en_US.UTF-8
$ pip3 install yamllint==1.23.0
Requirement already satisfied: yamllint==1.23.0 in /usr/local/lib/python3.8/site-packages (1.23.0)
Requirement already satisfied: pyyaml in /usr/local/lib/python3.8/site-packages (from yamllint==1.23.0) (5.3.1)
Requirement already satisfied: pathspec>=0.5.3 in /usr/local/lib/python3.8/site-packages (from yamllint==1.23.0) (0.8.0)
$ yamllint .
$ pip3 install yamllint==1.24.0
Collecting yamllint==1.24.0
  Using cached yamllint-1.24.0-py2.py3-none-any.whl (59 kB)
Requirement already satisfied: pathspec>=0.5.3 in /usr/local/lib/python3.8/site-packages (from yamllint==1.24.0) (0.8.0)
Requirement already satisfied: pyyaml in /usr/local/lib/python3.8/site-packages (from yamllint==1.24.0) (5.3.1)
Installing collected packages: yamllint
  Attempting uninstall: yamllint
[...]

PR #289, opened 2020-07-16
Add build environment without UTF-8 locales to travis-ci

wolfgangwalther:
Preventing regressions like #285

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ install:
   - pip install pyyaml coveralls flake8 flake8-import-order doc8
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then pip install sphinx; fi
   - pip install .
-  - if [[ $REMOVE_LOCALES ]]; then rm -rf /usr/lib/locale/* ; fi
+  - if [[ $REMOVE_LOCALES ]]; then sudo rm -rf /usr/lib/locale/* ; fi
 script:
   - if [[ $TRAVIS_PYTHON_VERSION != nightly ]]; then flake8 .; fi
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then doc8 $(git ls-files '*.rst'); fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ install:
   - pip install pyyaml coveralls flake8 flake8-import-order doc8
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then pip install sphinx; fi
   - pip install .
-  - if [[ $REMOVE_LOCALES ]]; then sudo rm -rf /usr/lib/locale/* ; fi
+  - if [[ $REMOVE_LOCALES = "true" ]]; then sudo rm -rf /usr/lib/locale/* ; fi
 script:
   - if [[ $TRAVIS_PYTHON_VERSION != nightly ]]; then flake8 .; fi
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then doc8 $(git ls-files '*.rst'); fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ install:
   - pip install pyyaml coveralls flake8 flake8-import-order doc8
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then pip install sphinx; fi
   - pip install .
-  - if [[ $REMOVE_LOCALES = "true" ]]; then sudo rm -rf /usr/lib/locale/* ; fi
+  - if [[ $REMOVE_LOCALES = "true" ]]; then sudo rm -rf /usr/lib/locale/*; fi
 script:
   - if [[ $TRAVIS_PYTHON_VERSION != nightly ]]; then flake8 .; fi
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then doc8 $(git ls-files '*.rst'); fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,14 @@ python:
   - 3.7
   - 3.8
   - nightly
+env:
+  - REMOVE_LOCALES=false
+  - REMOVE_LOCALES=true
 install:
   - pip install pyyaml coveralls flake8 flake8-import-order doc8
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then pip install sphinx; fi
   - pip install .
+  - if [[ $REMOVE_LOCALES ]]; then rm -rf /usr/lib/locale/* ; fi
 script:
   - if [[ $TRAVIS_PYTHON_VERSION != nightly ]]; then flake8 .; fi
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then doc8 $(git ls-files '*.rst'); fi
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -120,7 +120,8 @@ class CommandLineTestCase(unittest.TestCase):
 
         shutil.rmtree(cls.wd)
 
-    @unittest.skipIf(not utf8_available() and sys.version_info < (3, 7), 'UTF-8 paths with python < 3.7 and without locale not supported')
+    @unittest.skipIf(not utf8_available() and sys.version_info < (3, 7),
+                     'UTF-8 paths not supported')
     def test_find_files_recursively(self):
         conf = config.YamlLintConfig('extends: default')
         self.assertEqual(
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -56,11 +56,8 @@ class RunContext(object):
         return self._raises_ctx.exception.code
 
 
-# Check system's UTF-8 availability, because without it using UTF-8 paths
-# like 'éçäγλνπ¥' will break on Python ⩽ 3.6
-def utf8_paths_supported():
-    if sys.version_info >= (3, 7):
-        return True
+# Check system's UTF-8 availability
+def utf8_available():
     try:
         locale.setlocale(locale.LC_ALL, 'C.UTF-8')
         locale.setlocale(locale.LC_ALL, (None, None))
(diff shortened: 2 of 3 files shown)


adrienverge:
Thanks, that's a good idea.

It shows interesting problems on Python >= 3, <= 3.6:
```
======================================================================
ERROR: setUpClass (tests.test_cli.CommandLineTestCase)
  File "/home/travis/build/adrienverge/yamllint/tests/test_cli.py", line 102, in setUpClass
    'en.yaml': '---\n'
  File "/home/travis/build/adrienverge/yamllint/tests/common.py", line 61, in build_temp_workspace
    if not os.path.exists(os.path.dirname(path)):
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/genericpath.py", line 19, in exists
    os.stat(path)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 39-46: ordinal not in range(128)
```

wolfgangwalther:
Yes, I am looking at that right now as well. Not sure why it says line 102, but I would guess this is more about 89-94 - what do you think?

adrienverge:
> Yes, I am looking at that right now as well. Not sure why it says line 102, but I would guess this is more about 89-94 - what do you think?

I agree, line 102 looks OK. I guess line counting is altered by the right-to-left characters (`الأَبْجَدِيَّة العَرَبِيَّة`) on line 94, or by other high-unicode ones (`お早う御座`).

wolfgangwalther:
Yes, that makes sense.

I assume we have to adjust `setUpClass` and `test_find_files_recursively` accordingly, if no UTF-8 locale is present. Not sure what the smartest way of doing that would be, yet. Any ideas?

tests/test_cli.py
```diff
@@ -185,18 +198,23 @@ def test_find_files_recursively(self):
...
+                 os.path.join(self.wd, 'en.yaml'),
+                 os.path.join(self.wd, 'no-yaml.json'),
+                 os.path.join(self.wd,
+                              's/s/s/s/s/s/s/s/s/s/s/s/s/s/s/file.yaml'),
+                 os.path.join(self.wd, 'sub/directory.yaml/empty.yml'),
+                 os.path.join(self.wd, 'sub/directory.yaml/not-yaml.txt'),
+                 os.path.join(self.wd, 'sub/ok.yaml'),
+                 os.path.join(self.wd, 'warn.yaml')]
+                +
+                ([] if self.utf8_missing else
```
wolfgangwalther, line 215: hard to get this part looking good and passing the linter as well. any suggestions welcome.

wolfgangwalther:
So, I constructed those lists in `test_find_files_recursively` conditionally, depending on utf8 locale availability. I guess that's the only way to do it. Formatting that nicely is hard, though ;)

wolfgangwalther:
Still not clear, though, why this is just a problem on some python versions and not on all of them, when no UTF-8 locale is available.

.travis.yml
```diff
@@ -8,10 +8,14 @@ python:
...
   - 3.8
   - nightly
+env:
+  - REMOVE_LOCALES=false
+  - REMOVE_LOCALES=true
 install:
   - pip install pyyaml coveralls flake8 flake8-import-order doc8
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then pip install sphinx; fi
   - pip install .
+  - if [[ $REMOVE_LOCALES = "true" ]]; then sudo rm -rf /usr/lib/locale/* ; fi
```
adrienverge, line 18: Can you remove the extra space before `;`?

tests/test_cli.py
```diff
@@ -61,7 +61,16 @@ class CommandLineTestCase(unittest.TestCase):
     def setUpClass(cls):
         super(CommandLineTestCase, cls).setUpClass()
 
-        cls.wd = build_temp_workspace({
+        # Check system's UTF-8 availability, because without it
+        # using UTF-8 paths will break
```
adrienverge, line 65: I propose:
```suggestion
        # using UTF-8 paths like 'éçäγλνπ¥' will break on Python ⩽ 3.6
```

wolfgangwalther: I like how you're using a unicode character in the `<= 3.6` in the comment, where we blame python's unicode support - because we can! :)

wolfgangwalther:
Skipping tests instead would be fine with me - if it was only the tests that are using UTF-8 paths.

```python
@unittest.skipIf(not utf8_paths_supported(), 'UTF-8 paths not supported')
class CommandLineTestCase(unittest.TestCase):
```

Doesn't that basically skip the whole `test_cli.py` file? That would kind of defeat the purpose of this PR in the first place - because we would not test `cli.run()` at all for a system without UTF-8 locale (at least for some supported python versions). So this could in theory again lead to a similiar regression we had the first time with the locale stuff.

So I guess I would put the `@unittest.skipIf` decorator just on the relevant tests.

tests/test_cli.py
```diff
@@ -102,14 +108,26 @@ def setUpClass(cls):
             'en.yaml': '---\n'
                        'a: true\n'
                        'A: true'
-        })
+        }
+
+        if utf8_paths_supported():
+            # non-ASCII chars
+            workspace_def['non-ascii/éçäγλνπ¥/utf-8'] = (
```
wolfgangwalther, line 115: I still kept this part in, because this way we only need to disable two out of ~25 tests

tests/test_cli.py
```diff
@@ -425,17 +443,13 @@ def test_run_empty_file(self):
...
-        path = os.path.join(self.wd, 'non-ascii', 'éçäγλνπ¥', 'utf-8')
-
-        # Make sure the default localization conditions on this "system"
-        # support UTF-8 encoding.
-        try:
-            locale.setlocale(locale.LC_ALL, (None, 'UTF-8'))
-        except locale.Error:
-            self.skipTest('no UTF-8 locale available')
-        self.addCleanup(locale.setlocale, locale.LC_ALL, (None, None))
+        if sys.version_info < (3, 7):
```
wolfgangwalther, line 448: added a check for python version here - because we basically identified through the previous test results, that we only need to manually set the locale to UTF-8 to support those paths for those earlier versions.

wolfgangwalther:
Are you sure the "bytes" stuff is actually going to work?

The original error message didn't complain about the `.encode()` stuff in `setUpClass` - but about `os.path.exists` in `build_temp_workspace`:

```shell
  File "/home/travis/build/adrienverge/yamllint/tests/test_cli.py", line 102, in setUpClass
    'en.yaml': '---\n'
  File "/home/travis/build/adrienverge/yamllint/tests/common.py", line 61, in build_temp_workspace
    if not os.path.exists(os.path.dirname(path)):
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/genericpath.py", line 19, in exists
    os.stat(path)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 39-46: ordinal not in range(128)
```

If I'm not mistaken we are passing a UTF-8 encoded string (which is perfectly supported by python, I'm pretty sure) to `os.stat` - where it probably get's encoded as `ascii` (otherwise I don't know where that mentioning of `ascii` comes from - but this part is probably dependent on the systems locales). Why would the `bytes` string you are putting in right now be any different?

Now, if that patch indeed works - can't we just convert the `UTF-8` string to a byte string in `build_temp_workspace`? I don't think we need to touch `setUpClass` at all with all the single bytes in there.

Because let's be honest: `setUpClass` is in no way easier to read or understand in your suggestion. Probably a lot harder than before - at least for me. And it doesn't look nice as well ;)

wolfgangwalther:
Maybe I'm misunderstanding what `.encode('utf-8')` does - but is there actually any difference between the following two?

```python
            (
                u'---\n'
                u'- hétérogénéité\n'
                u'# 19.99 €\n'
                u'- お早う御座います。\n'
                u'# الأَبْجَدِيَّة العَرَبِيَّة\n').encode('utf-8')
```

and

```
            (
                b'---\n'
                b'- h\xc3\xa9t\xc3\xa9rog\xc3\xa9n\xc3\xa9it\xc3\xa9\n'
                b'# 19.99 \xe2\x82\xac\n'
                b'- \xe3\x81\x8a\xe6\x97\xa9\xe3\x81\x86\xe5\xbe\xa1\xe5\xba'
                b'\xa7\xe3\x81\x84\xe3\x81\xbe\xe3\x81\x99\xe3\x80\x82\n'
                b'# \xd8\xa7\xd9\x84\xd8\xa3\xd9\x8e\xd8\xa8\xd9\x92\xd8\xac'
                b'\xd9\x8e\xd8\xaf\xd9\x90\xd9\x8a\xd9\x8e\xd9\x91\xd8\xa9 '
                b'\xd8\xa7\xd9\x84\xd8\xb9\xd9\x8e\xd8\xb1\xd9\x8e\xd8\xa8\xd9'
                b'\x90\xd9\x8a\xd9\x8e\xd9\x91\xd8\xa9\n')
```

I think this is exactly what `.encode` does, right?

wolfgangwalther:
Your patch also has this: `u'non-ascii/éçäγλνπ¥/utf-8'.encode('utf-8')` - so a `.encode()` on the **key/path**.

I put this into `build_temp_workspace` and was able to get it working locally. Let's see what the CI builds say, in all the different flavours.

wolfgangwalther:
And - I had to revert the python version check in `test_run_non_ascii_file`. Setting the locale here seems to be important for the body of the file - not the path. And that seems to be independent of python versions.

wolfgangwalther:
Putting both the version and locale check into `def utf8_paths_supported()` didn't work out. I changed that again. One of the two tests we are skipping now is skipped only depending on locale availability (`test_run_non_ascii_file` - we need the locale in any case) and the other test depends on either an available UTF-8 locale OR python >= 3.7 to run right now.

The change to `build_temp_workspace` seems to have broken the python 2.7 tests now, though.

adrienverge:
I like this new version :+1: 


> Are you sure the "bytes" stuff is actually going to work?

I tried and it seemed to work; maybe passing an already-encoded string to `os.path` avoids any encoding/decoding at the Python level?


> The change to `build_temp_workspace` seems to have broken the python 2.7 tests now, though.

It's because `test_cli.py` passes `"éçäγλνπ¥"` to `build_temp_workspace()` as bytes instead of an Unicode string. With the following change, it should work:
```diff
             # non-ASCII chars
-            'non-ascii/éçäγλνπ¥/utf-8': (
+            u'non-ascii/éçäγλνπ¥/utf-8': (
                 u'---\n'
```
We're very close! :)

wolfgangwalther:
> I tried and it seemed to work; maybe passing an already-encoded string to os.path avoids any encoding/decoding at the Python level?

I think it worked - but not because of all the manually encoded bytes (as they were the content of the file), but only the `.encode('utf-8')` on the key you had in there as well! That's essentially what I am doing now as well.

Your suggestion about adding `u'...'` makes a lot of sense. Let's see whether that turns all of them green :)

If it does - I think we have found very good solution!

PR #301, opened 2020-08-18
Update commas.py

kdeyko:
Error in the example snippet, it would NOT pass otherwise actually:
```
$ cat test.yml
strange var:
  [10, 20,30, {x: 1, y: 2}]

$ yamllint -d "{extends: default, rules: {commas: {min-spaces-after: 1, max-spaces-after: 1}}}" test.yml
test.yml
  2:11      error    too few spaces after comma  (commas)
```


adrienverge (approved):
Thanks!

PR #314, opened 2020-09-03
Add new rules to forbid flow collection styles

ssato:
It's desirable for some use cases such like ansible playbooks to forbid flow collection styles as needed and this pair of new rules accomplish that.

Of course, these should not be enabled for most use cases so that I intended to disable them by default.


adrienverge:
Instead of creating new rules, what about `braces: {forbid: true}` and `brackets: {forbid: true}`? The rules `braces` and `brackets` already exist, I think it would be less confusing for users.

ssato:
> Instead of creating new rules, what about `braces: {forbid: true}` and `brackets: {forbid: true}`? The rules `braces` and `brackets` already exist, I think it would be less confusing for users.

Thanks a lot for your suggestion!

I fully agree with you and will open another PR because the way is different from this.

ssato:
I created another PR #319 to cover this.

PR #316, opened 2020-09-07
Fix detecting strings with hashtag as requiring quotes

julienfalque:
Fixes #271.

This is my first Python contribution ever, please let me know if there's anything wrong :)

PR #317, opened 2020-09-07
Add .eggs to ignored files

julienfalque:
When running tests for #316, the directory `.eggs` was created. I don't know the Python world yet, maybe I did something wrong 😅️. Otherwise this will prevent commiting the directory accidentally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 __pycache__
 *.py[cod]
+/.eggs
 /docs/_build
 /dist
 /yamllint.egg-info


adrienverge:
I'm not sure what happened, but I cannot reproduce (not `.eggs` directory created when I run `python setup.py test`).

I'll close but please reopen if you think I'm wrong!

julienfalque:
I ran the tests in a Docker container (`python:3.8-alpine3.12`). Maybe that's related?

PR #319, opened 2020-09-08
Add 'forbid' configuration parameters to the braces and brackets rules

ssato:
This is another try to accomplish the same thing in the PR #314 without adding new rules.

It's desirable for some use cases such like ansible playbooks to forbid flow collection styles as needed and this changs make that happen.

Of course, these should not be enabled for most use cases so that these configurations are disable by default.

### DIFF
--- a/tests/rules/test_braces.py
+++ b/tests/rules/test_braces.py
@@ -31,6 +31,36 @@ class ColonTestCase(RuleTestCase):
                    'dict6: {  a: 1, b, c: 3 }\n'
                    'dict7: {   a: 1, b, c: 3 }\n', conf)
 
+    def test_forbid(self):
+        conf = ('braces:\n'
+                '  forbid: false\n')
+        self.check('---\n'
+                   'dict: {}\n', conf)
+        self.check('---\n'
+                   'dict: {a}\n', conf)
+        self.check('---\n'
+                   'dict: {a: 1}\n', conf)
+        self.check('---\n'
+                   'dict: {\n'
+                   '  a: 1\n'
+                   '}\n', conf)
+
+        conf = ('braces:\n'
+                '  forbid: true\n')
+        self.check('---\n'
+                   'dict:\n'
+                   '  a: 1\n', conf)
+        self.check('---\n'
+                   'dict: {}\n', conf, problem=(2, 8))
+        self.check('---\n'
+                   'dict: {a}\n', conf, problem=(2, 8))
+        self.check('---\n'
+                   'dict: {a: 1}\n', conf, problem=(2, 8))
+        self.check('---\n'
+                   'dict: {\n'
+                   '  a: 1\n'
+                   '}\n', conf, problem=(2, 8))
+
     def test_min_spaces(self):
         conf = ('braces:\n'
                 '  max-spaces-inside: -1\n'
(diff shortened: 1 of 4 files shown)


tests/rules/test_braces.py
```diff
@@ -31,6 +31,24 @@ def test_disabled(self):
                    'dict6: {  a: 1, b, c: 3 }\n'
                    'dict7: {   a: 1, b, c: 3 }\n', conf)
 
+    def test_forbid(self):
+        conf = ('braces:\n'
+                '  forbid: false\n')
+        self.check('---\n'
+                   'dict: {}\n', conf)
+        self.check('---\n'
+                   'dict: {a: 1}\n', conf)
```
adrienverge, line 40: Can you add these examples please?
```yaml
dict: {a}
```
```yaml
dict: {
  a: 1
}
```
(same with `forbid: true` and with flow sequences)

yamllint/rules/braces.py
```diff
@@ -103,23 +122,31 @@
...
            'min-spaces-inside-empty': -1,
            'max-spaces-inside-empty': -1}
 
 
 def check(conf, token, prev, next, nextnext, context):
-    if (isinstance(token, yaml.FlowMappingStartToken) and
+    if conf['forbid'] and isinstance(token, yaml.FlowMappingStartToken):
+        yield LintProblem(token.start_mark.line + 1,
+                          token.end_mark.column + 1,
+                          'Flow mappings are forbidden to use')
```
adrienverge, line 147: I propose
```suggestion
                          'forbidden flow mapping')
```
to be consistent with other messages (`forbidden implicit octal value`, `empty value in block mapping`, `comment not indented like content`, `found forbidden document end`...)

adrienverge (commented):
Looks good, thanks @ssato!

tests/rules/test_braces.py
ssato: Thanks for your suggestion! I applied it, squashed and force-pushed.

yamllint/rules/braces.py
ssato: Thanks for your suggestion! I applied it along with similar change in yamllint/rules/brackets.py, squashed and force-pushed.

tests/rules/test_braces.py
```diff
@@ -31,6 +31,30 @@ def test_disabled(self):
...
+        self.check('---\n'
+                   'dict: {\n'
+                   '  a: 1\n'
+                   '}\n', conf)
+
+        conf = ('braces:\n'
+                '  forbid: true\n')
+        self.check('---\n'
+                   'dict:\n'
+                   '  a: 1\n', conf)
```
adrienverge, line 52: :+1:

tests/rules/test_braces.py
adrienverge: Thanks, but you didn't add the same to `forbid: true` and for flow sequences. It would be useful to check these:
```python
        self.check('---\n'
                   'dict: {a}\n', conf, problem=(2, 8))
```
and:
```yaml
array: [
  a,
  b,
]
```
↑ both with `forbid: true` and `false`

ssato: Thanks! I added these test cases also.

adrienverge (approved):
Thanks a lot @ssato!

PR #321, opened 2020-09-12
fix: add runtime dependency to setuptools

ssato:
Because yamllint depends on pkg_resources.load_entry_point from
setuptools to make it worked, this runtime dependency to setuptools is
necessary to list.


adrienverge (approved):
Thanks!

PR #322, opened 2020-09-23
Add support for GitHub annotations

andrewimeson:
This adds support for the output format that GitHub Actions can use to automatically annotate pull requests with linter failures.

I'm not very good with Python, so there are two ugly things I need recommendations for improving:
1. The two tests define the same magical output twice. I image I could go to one variable outside of the function definition, but I don't know where to put it
2. In the test for auto-detection I don't think the way I unset variables is as clean as it could be

I tested this by adding a bad YAML file (`a.yaml`) and then creating a GitHub Actions definition in `.github/workflows/test.yml` that looks like this

```yaml
---
name: yamllint test

# yamllint disable-line rule:truthy
on: push

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2

      - name: Set up Python
        uses: actions/setup-python@v2
        with:
          python-version: 3.8

      - name: Setup Python venv
        run: |
          python -m venv ./.venv && . ./.venv/bin/activate &&
          pip install pyyaml coveralls flake8 flake8-import-order doc8
          pip install .
#      - name: Lint bad yaml file by explicitly telling it the format
#        run: |
#          source ./.venv/bin/activate &&
#          yamllint -f github a.yaml

      - name: Lint bad yaml file by autodetection
        run: |
          source ./.venv/bin/activate &&
          yamllint a.yaml
```
and it successfully outputs the annotations
image

### DIFF
--- a/docs/integration.rst
+++ b/docs/integration.rst
@@ -17,3 +17,37 @@ Here is an example, to add in your .pre-commit-config.yaml
     hooks:
       - id: yamllint
         args: [-c=/path/to/.yamllint]
+
+Integration with GitHub Actions
+-------------------------------
+
+yamllint auto-detects when it's running inside of `GitHub
+Actions<https://github.com/features/actions>` and automatically uses the suited
+output format to decorate code with linting errors automatically. You can also
+force the GitHub Actions output with ``yamllint --format github``.
+
+An example workflow using GitHub Actions:
+
+.. code:: yaml
+
+   ---
+   name: yamllint test
+
+   on: push
+
+   jobs:
+     test:
+       runs-on: ubuntu-latest
+       steps:
+         - uses: actions/checkout@v2
+
+         - name: Set up Python
+           uses: actions/setup-python@v2
+           with:
+             python-version: 3.8
+
+         - name: Install yamllint
+           run: pip install yamllint
+
+         - name: Lint YAML files
+           run: yamllint .
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -549,6 +549,20 @@ class CommandLineTestCase(unittest.TestCase):
         self.assertEqual(
             (ctx.returncode, ctx.stdout, ctx.stderr), (1, expected_out, ''))
 
+    def test_run_format_github(self):
+        path = os.path.join(self.wd, 'a.yaml')
+
+        with RunContext(self) as ctx:
+            cli.run((path, '--format', 'github'))
+        expected_out = (
+            '::error file=%s,line=2,col=4::[trailing-spaces] trailing'
+            ' spaces\n'
+            '::error file=%s,line=3,col=4::[new-line-at-end-of-file] no'
+            ' new line character at the end of file\n'
+            % (path, path))
+        self.assertEqual(
+            (ctx.returncode, ctx.stdout, ctx.stderr), (1, expected_out, ''))
+
     def test_run_read_from_stdin(self):
         # prepares stdin with an invalid yaml string so that we can check
         # for its specific error, and be assured that stdin was read
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -563,6 +563,24 @@ class CommandLineTestCase(unittest.TestCase):
         self.assertEqual(
             (ctx.returncode, ctx.stdout, ctx.stderr), (1, expected_out, ''))
 
+    def test_github_actions_detection(self):
+        path = os.path.join(self.wd, 'a.yaml')
+        self.addCleanup(os.environ.__delitem__, 'GITHUB_ACTIONS')
+        self.addCleanup(os.environ.__delitem__, 'GITHUB_WORKFLOW')
+
+        with RunContext(self) as ctx:
+            os.environ['GITHUB_ACTIONS'] = 'something'
+            os.environ['GITHUB_WORKFLOW'] = 'something'
+            cli.run((path, ))
+        expected_out = (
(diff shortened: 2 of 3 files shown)


docs/integration.rst
```diff
@@ -17,3 +17,38 @@ Here is an example, to add in your .pre-commit-config.yaml
     hooks:
       - id: yamllint
         args: [-c=/path/to/.yamllint]
+
+Integration with GitHub Actions
+-------------------------------
+
+Starting with version XXX of yamllint, it will automatically use the
```
andrewimeson, line 24: This needs a version number or to be rephrased to not specify when the feature was introduced

yamllint/cli.py
```diff
@@ -50,6 +50,11 @@ def supports_color():
             hasattr(sys.stdout, 'isatty') and sys.stdout.isatty())
 
 
+def is_github_actions():
+    if 'GITHUB_ACTIONS' in os.environ and 'GITHUB_WORKFLOW' in os.environ:
+        return True
```
adrienverge, line 55:
```suggestion
    return 'GITHUB_ACTIONS' in os.environ and 'GITHUB_WORKFLOW' in os.environ
```

yamllint/cli.py
```diff
@@ -96,6 +114,9 @@ def show_problems(problems, file, args_format, no_warn):
             continue
         if args_format == 'parsable':
             print(Format.parsable(problem, file))
+        elif args_format == 'github' or \
+                (args_format == 'auto' and is_github_actions()):
```
adrienverge, line 118: Can we drop `is_github_actions()` and simplify to:
```suggestion
        elif args_format == 'github' or (args_format == 'auto' and 
            'GITHUB_ACTIONS' in os.environ and 'GITHUB_WORKFLOW' in os.environ):
```
?

docs/integration.rst
adrienverge: Including a version number will make maintenance harder, and won't be much useful once the version is released and installed. I prefer not including it.

About the wording, I propose a small simplification, what do you think?
```
yamllint auto-detects when it's running inside of `GitHub
Actions<https://github.com/features/actions>` and automatically uses the suited
output format to decorate code with linting errors automatically. You can also
force the GitHub Actions output with ``yamllint --format github``.
```

docs/integration.rst
```diff
@@ -17,3 +17,39 @@ Here is an example, to add in your .pre-commit-config.yaml
...
+`annotations API <https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message>`
+to decorate code with lint errors automatically. It should auto-detect that it
+is running inside of `GitHub Actions<https://github.com/features/actions>` and automatically use the correct output
+format, but you can also force it with ``yamllint --format github``
+
+An example workflow using GitHub Actions
```
adrienverge, line 30:
```suggestion
An example workflow using GitHub Actions:
```

docs/integration.rst
```diff
@@ -17,3 +17,39 @@ Here is an example, to add in your .pre-commit-config.yaml
...
+
+.. code:: yaml
+
+   ---
+   name: yamllint test
+
+   # yamllint disable-line rule:truthy
```
adrienverge, line 37: If this line is not mandatory here, can you remove it, for the clarity of the example?

docs/integration.rst
```diff
@@ -17,3 +17,39 @@ Here is an example, to add in your .pre-commit-config.yaml
...
+
+         - name: Set up Python
+           uses: actions/setup-python@v2
+           with:
+             python-version: 3.8
+
+         - name: Install yamllint
+           run: pip install yamllint
+
+         - name: Lint bad yaml file by autodetection
```
adrienverge, line 54:
```suggestion
         - name: Lint YAML files
```

tests/test_cli.py
```diff
@@ -549,6 +549,39 @@ def test_run_format_colored(self):
...
+            % (path, path))
+        self.assertEqual(
+            (ctx.returncode, ctx.stdout, ctx.stderr), (1, expected_out, ''))
+
+    def test_github_actions_detection(self):
+        path = os.path.join(self.wd, 'a.yaml')
+
+        with RunContext(self) as ctx:
+            os.environ['GITHUB_ACTIONS'] = "something"
+            os.environ['GITHUB_WORKFLOW'] = "something"
```
adrienverge, line 571:
```suggestion
            os.environ['GITHUB_ACTIONS'] = 'something'
            os.environ['GITHUB_WORKFLOW'] = 'something'
```

adrienverge (commented):
Hello @andrewimeson,

> I'm not very good with Python

Honestly this is a neat contribution, with docs and tests!

> * The two tests define the same magical output twice. I image I could go to one variable outside of the function definition, but I don't know where to put it

I would keep them like that (and remove the `IMPROVE` comment). These strings are short, having them inlined like this helps to read the test.

> * In the test for auto-detection I don't think the way I unset variables is as clean as it could be

You can grep `"os.environ"` in `test_cli.py`, you'll see examples of how to perform a good cleanup :wink:

yamllint/cli.py
andrewimeson: dropped this function in favor of your suggestion to move it to the format selection block

adrienverge (approved):
That's perfect (in two commits :+1:)

Thank you for your contribution @andrewimeson.

PR #326, opened 2020-09-26
directives: Fix DOS lines messing with rule IDs

### LINKED ISSUE #325: DOS lines mess with rule IDs in directives
Hi!

I encountered this issue on Windows. When using a directive with rules in a file with DOS line endings (`\r\n`), the last rule in the list is ignored. For example:

```yaml
# yamllint disable-line rule:colons rule:trailing-spaces
- bad  : colon and spaces   
```

In this example, the `trailing-spaces` rule would not be disabled.

I will open a PR with a fix as soon as I can.

MathieuCouette:
Fixes #325 

The linter allows a directive to contain trailing whitespace characters like `\r`, but does not trim them before iterating on the rules. As a result, the last rule in the list contains the trailing whitespace characters and never matches any existing rule.

I added the necessary trimming, as well as a test with 2 checks to go along with it.

### DIFF
--- a/tests/test_yamllint_directives.py
+++ b/tests/test_yamllint_directives.py
@@ -232,6 +232,34 @@ class YamllintDirectivesTestCase(RuleTestCase):
                    problem1=(3, 18, 'trailing-spaces'),
                    problem2=(4, 8, 'colons'))
 
+    def test_disable_directive_with_rules_and_dos_lines(self):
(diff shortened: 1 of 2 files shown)


adrienverge (approved):
@MathieuCouette thanks for your contribution! A perfect commit with a clear message and tests :+1:

PR #327, opened 2020-09-26
gitignore: Add /.eggs

MathieuCouette:
Quick PR to ignore the `/.eggs` folder, which appears to be generated every time the `python setup.py test` command is run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ __pycache__
 /dist
 /yamllint.egg-info
 /build
+/.eggs


adrienverge:
Hello,

When I run `python setup.py test`, there is no `.eggs` folder created (instead, a directory `yamllint.egg-info` is generated, but it's already in `.gitignore`). What Python version are you using?

MathieuCouette:
```bash
root@964ae9d82fd6:/workspaces/yamllint# python --version
Python 3.8.5
root@964ae9d82fd6:/workspaces/yamllint# easy_install --version
setuptools 50.3.0 from /usr/local/lib/python3.8/site-packages (Python 3.8)
```

According to [the changelogs](https://setuptools.readthedocs.io/en/latest/history.html#id591), this feature was introduced in 2014 with the 7.0 version of `setuptools`:

> Eggs that are downloaded for `setup_requires`, `test_requires`, etc. are now placed in a `./.eggs` directory instead of directly in the current directory.

I couldn't find much more info, but it was added to [the `.gitignore` of the `setuptools` project](https://github.com/pypa/setuptools/commit/d59415c6921bcb1282feffcf104d18949040b404) a while ago:

> Directory used by setuptools to auto-install dependencies

In my case, it contains the packages listed in `install_requires`. It could be that the directory is not generated for you because you already have those packages installed globally.

MathieuCouette:
The content of the `./.eggs/README.txt` file:

```
This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.

This directory caches those eggs to prevent repeated downloads.

However, it is safe to delete this directory.


```

adrienverge (approved):
Thanks for finding the info!

Issue #328, opened 2020-09-26
'python setup.py test' is deprecated


MathieuCouette:
When I run the `python setup.py test` command as recommended in the `CONTRIBUTING.rst` file, I get the following deprecation warning:

```bash
root@964ae9d82fd6:/workspaces/yamllint# python setup.py test
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

It appears the command was deprecated last year (pypa/setuptools#1684) and might soon be completely removed (https://github.com/pypa/setuptools/issues/931#issuecomment-694482786). It may be safe to investigate an alternative like `tox` (#320).

adrienverge:
Hello @MathieuCouette, thanks for the notice.

The problem of pull request #320 is that `tox` itself calls `coverage run --source=yamllint setup.py test`, which calls `python setup.py test` and triggers the deprecation warning. @ssato @MathieuCouette what would you recommend?

MathieuCouette:
Replacing `python setup.py test` with `pytest` seems to work, and looks like the easiest fix to me. This requires `pytest` to be installed, as well as the content of `install_requires` (which can be installed with `pip install .`).

In this situation, the move to `tox` addresses a different problem, which is that the test (development) dependencies currently have to be manually installed. Moving to `pytest` adds yet another bunch of these dependencies.

I don't think `tox` is necessary, but it seems nice. I don't see it as adding another test (development) dependency to be manually installed, but rather as replacing all of them (e.g., `doc8`, `flake8`, `flake8-import-order`, `pathspec`, `pytest`, `pyyaml`, `sphinx`). There are of course other advantages, but I think @ssato already touched on those.

I can leave my comments on the PR, if you want.

P.S.: I come from the Node.js world, and while I've used the language itself before, I am fairly new to Python development. I hadn't heard of `tox` before yesterday. 🙂

ssato:
> The problem of pull request #320 is that tox itself calls coverage run --source=yamllint setup.py test, which calls python setup.py test and triggers the deprecation warning. @ssato @MathieuCouette what would you recommend?

Maybe candidates are nose or nose2 or pytest and pytest is the strongest candidate, I think, because it seems that pytest is much popular than others and used in many oss projects already.

adrienverge:
Thanks for your feedback @ssato and @MathieuCouette. I have created #329 which is, I think, the easiest change to get this fixed.

PR #329, opened 2020-09-29
tests: Stop using deprecated 'python setup.py test'

adrienverge:
Using `python setup.py test` is now deprecated [1], users are encouraged
to be explicit about the test command.

Running yamllint tests using the Python standard library (`unittest`)
can be done using:

    python -m unittest discover

Why not nose, tox or pytest? Because they would add a dependency, make
tests running more complicated and verbose for new users, and their
benefit is not worth for this simple project (only 2 runtime
dependencies: PyYAML and pathspec).

Resolves https://github.com/adrienverge/yamllint/issues/328.

\[1]: https://github.com/pypa/setuptools/pull/1878

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ script:
   - if [[ $TRAVIS_PYTHON_VERSION != nightly ]]; then flake8 .; fi
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then doc8 $(git ls-files '*.rst'); fi
   - yamllint --strict $(git ls-files '*.yaml' '*.yml')
-  - coverage run --source=yamllint setup.py test
+  - coverage run --source=yamllint -m unittest discover
   - if [[ $TRAVIS_PYTHON_VERSION != 2* ]]; then
       python setup.py build_sphinx;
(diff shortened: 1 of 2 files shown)


MathieuCouette:
I would consider adding a `pip install .` step to the `CONTRIBUTING.rst` file, as is done in the Travis CI file. `python setup.py test` would install missing dependencies in `./eggs`, `python -m unittest discover` doesn't.

Mind you, I was able to figure out that `pyyaml` needed to be installed globally from the `ModuleNotFoundError: No module named 'yaml'` logs.

adrienverge:
Thanks @MathieuCouette! I've updated the pull request with:
```diff
@@ -13,6 +13,7 @@ Pull Request Process
 
    .. code:: bash
 
+    pip install --user .
     python -m unittest discover  # all tests...
     python -m unittest tests/rules/test_commas.py  # or just some tests (faster)
 
```

MathieuCouette (approved):
Works perfectly in a clean Docker container. 👍 The `unittest` deprecation warnings are a bit more intrusive, though, so I created a PR to fix those: #330

PR #330, opened 2020-09-29
tests: Replace deprecated aliases

MathieuCouette:
This replaces the `unittest` deprecated aliases to avoid warnings when running the tests. For more info: https://docs.python.org/3/library/unittest.html#deprecated-aliases

### DIFF
--- a/tests/test_module.py
+++ b/tests/test_module.py
@@ -47,7 +47,7 @@ class ModuleTestCase(unittest.TestCase):
             subprocess.check_output([PYTHON, '-m', 'yamllint'],
                                     stderr=subprocess.STDOUT)
         self.assertEqual(ctx.exception.returncode, 2)
-        self.assertRegexpMatches(ctx.exception.output.decode(),
+        self.assertRegex(ctx.exception.output.decode(),
                                  r'^usage: yamllint')
 
     def test_run_module_on_bad_dir(self):
@@ -55,7 +55,7 @@ class ModuleTestCase(unittest.TestCase):
             subprocess.check_output([PYTHON, '-m', 'yamllint',
                                      '/does/not/exist'],
                                     stderr=subprocess.STDOUT)
-        self.assertRegexpMatches(ctx.exception.output.decode(),
+        self.assertRegex(ctx.exception.output.decode(),
                                  r'No such file or directory')
 
     def test_run_module_on_file(self):
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -246,19 +246,19 @@ class CommandLineTestCase(unittest.TestCase):
             cli.run(())
         self.assertNotEqual(ctx.returncode, 0)
         self.assertEqual(ctx.stdout, '')
-        self.assertRegexpMatches(ctx.stderr, r'^usage')
+        self.assertRegex(ctx.stderr, r'^usage')
 
         with RunContext(self) as ctx:
             cli.run(('--unknown-arg', ))
         self.assertNotEqual(ctx.returncode, 0)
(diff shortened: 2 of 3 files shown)


MathieuCouette:
If #329 does not fix the issue with older Python versions, we might just have to run the tests with `python -W ignore::DeprecationWarning -m unittest discover` to suppress the warnings instead.

adrienverge:
This is a known issue, but cannot be fixed easily because of Python 2.7 support...

- On 2021-01-01 Python 2 support will be dropped forever, so I planned to apply this change (same as this PR) then.
- I prefer waiting 3 more months with the warning (it has been here for a while), rather than hiding potential *new* warnings with `-W ignore::DeprecationWarning`.

What do you think?

MathieuCouette:
@adrienverge Agreed!

tests/test_module.py
```diff
@@ -47,15 +47,15 @@ def test_run_module_no_args(self):
             subprocess.check_output([PYTHON, '-m', 'yamllint'],
                                     stderr=subprocess.STDOUT)
         self.assertEqual(ctx.exception.returncode, 2)
-        self.assertRegexpMatches(ctx.exception.output.decode(),
+        self.assertRegex(ctx.exception.output.decode(),
                                  r'^usage: yamllint')
```
adrienverge, line 51: There are some indentation issues in this file.

adrienverge (commented):
Or we can apply this change now, and use `@unittest.skipIf(sys.version_info < (3, 0))` to avoid running these tests?

MathieuCouette:
> 
> 
> Or we can apply this change now, and use `@unittest.skipIf(sys.version_info < (3, 0))` to avoid running these tests?

The risk is low, but it's possible that a bug with Python 2 would be missed because some tests were skipped. If that's a risk you're willing to take, then sure.

Alternatively, we could add the necessary aliases for `assertRegex` and `assertRaisesRegex` to work with Python 2:

```python
        # https://docs.python.org/3/library/unittest.html#deprecated-aliases
        if sys.version_info < (3, 2):
            cls.assertRegex = cls.assertRegexpMatches
            cls.assertRaisesRegex = cls.assertRaisesRegexp
```

I've added this snippet to the `setUpClass` method of the 3 test classes that use `assertRegex` and `assertRaisesRegex`, and it seems to do the trick. It might also be easier to clean up once you remove Python 2 support.

What do you think?

adrienverge:
These are good points!

I have to say I don't care much about regressions for Python 2, since all legacy code will be dropped in 3 months.

Declaring aliases is smart, and can be a good solution (I'm just afraid that we forget them in 3 months when the oldest supported version will be Python 3.5).

MathieuCouette:
Both solutions appear as likely to be forgotten about to me. I suggest creating an issue on the abandonment of the Python 2.7 support in order to keep track of what needs to be done, like the removal of the aliases, and to indicate to users and contributors that the Python 2.7 support will soon be abandoned.

adrienverge:
> Both solutions appear as likely to be forgotten about to me.

You're probably right. Anyway I plan to `grep` things like `sys.version_info` everywhere, so no need to create an issue, in my opinion.

> and to indicate to users and contributors that the Python 2.7 support will soon be abandoned.

There's already a message on top of the README file / home page, it has been here for a while!

MathieuCouette:
> Anyway I plan to `grep` things like `sys.version_info` everywhere, so no need to create an issue, in my opinion.

Sounds good.

> There's already a message on top of the README file / home page, it has been here for a while!

Right, totally my bad!

I'll open a new PR with the changes, since I rebased my branch.

PR #335, opened 2020-10-11
tests: Replace deprecated aliases

MathieuCouette:
Rebasing of PR #330. This replaces the `unittest` deprecated aliases to avoid warnings when running the tests.

For more info: https://docs.python.org/3/library/unittest.html#deprecated-aliases


adrienverge (approved):
Perfect!

Thanks @MathieuCouette :+1:

PR #339, opened 2020-10-29
Use package resources instead of __file__

silverjam:
In order for yamllint to be packaged by [PyOxidizer](https://pyoxidizer.readthedocs.io)-- it needs to avoid the use of `__file__` (see the [PyOxidizer docs][1] for more details).  Instead use `pkg_resources` so that the default config can be bundled within the (single executable) package that PyOxidizer produces.

See https://importlib-resources.readthedocs.io/en/latest/ for details on the `pkg_resources` module.

[1]: https://pyoxidizer.readthedocs.io/en/stable/oxidized_importer_behavior_and_compliance.html?highlight=__file__#file-and-cached-module-attributes


yamllint/config.py
```diff
@@ -20,6 +20,13 @@
 import yaml
 
 import yamllint.rules
+from yamllint import conf
+
+try:
+    import importlib.resources as pkg_resources
+except ImportError:
+    # Try backported to PY<37 `importlib_resources`.
```
silverjam, line 28: Following advice from here: https://stackoverflow.com/a/20885799/749342

setup.cfg
```diff
@@ -48,13 +48,14 @@ project_urls =
 [options]
 packages = find:
 
-python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
+python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
```
adrienverge, line 51: This change doesn't seem directly related, and may have other implications. I'm not against dropping support for Python 3.4, but it should be advertized in another PR with a proper title (that could add official support for 3.9 in the same time?)

yamllint/config.py
```diff
@@ -20,6 +20,13 @@
...
+    import importlib_resources as pkg_resources
```
adrienverge, line 29: New dependencies cause other problems. Even if this one is part of standard lib after Python 3.9, I would prefer not adding a new one as long as we support Python ⩽ 3.8.

Is it considerable to do something like this?
```python
try:
    import importlib.resources
    _PATH = importlib.resources.resource_filename(...)
except ImportError:
    _PATH = os.path.dirname(os.path.realpath(__file__)```

**Review by adrienverge — COMMENTED** [review]
Hello,

I didn't know about PyOxidizer, cool use of Rust to package Python :+1: 

I see it's quite a young project and doesn't have a non-beta release yet: do you know how much this tool is used and by whom?

**Comment by silverjam** [comment]
> I see it's quite a young project and doesn't have a non-beta release yet: do you know how much this tool is used and by whom?

In terms of usage, I've successfully packaged yamllint with this change...  the project is still fairly young for sure, and I'm not sure who's using it in an official/production capacity, but it's made a lot of progress recently.  I'm personally excited about this project though, it's a "revolution" in how Python projects (and utilities like yamllint in particular) can be packaged (it brings the ease of a single binary deployment to Python essentially).  I know my excitement is not a good reason to accept changes that are detrimental to other users, so hopefully we can find a compromise.

File: setup.cfg (hunk 2 above)
Comment (silverjam) [reply setup.cfg:51]: importlib_resources is not supported by Python 3.4, which is why this was introduced, happy to drop this though

File: yamllint/config.py (hunk 3 above)
Comment (silverjam) [reply yamllint/config.py:29]: It would be a bit ugly, but something like the following would work:

```python
try:
    import importlib.resources
    _PATH = importlib.resources.resource_filename(...)
except ImportError:
    try:
        import importlib_resources
        _PATH = importlib_resources.resource_filename(...)
    except ImportError:
        _PATH = os.path.dirname(os.path.realpath(__file__)
```

This way importlib_resources could be manually installed in order to make packaging successful for Python <3.9.  Or importlib_resources could be added as a "extra" dependency.

silverjam:
Here's an example of a fully statically linked binary of yamllint, which will run on any Linux system: [yamllint-linux.zip](https://github.com/adrienverge/yamllint/files/5484760/yamllint-linux.zip).

```
❯ sha1sum yamllint
23df57ac69f308b6de3cded25810e8d374466e58  yamllint
❯ ldd yamllint-linux                                                                                                                                                                                                           (base)
        not a dynamic executable
❯ ./yamllint-linux
usage: yamllint [-h] [-] [-c CONFIG_FILE | -d CONFIG_DATA] [-f {parsable,standard,colored,github,auto}] [-s] [--no-warnings] [-v] [FILE_OR_DIR [FILE_OR_DIR ...]]
yamllint: error: one of the arguments FILE_OR_DIR - is required

File: setup.cfg (hunk 2 above)
Comment (adrienverge) [reply setup.cfg:51]: Please don't drop it if it's needed. My proposal was to make it official (to other people reviewing yamllint) in a dedicated PR (and why not add official support for 3.9 in the same time).

File: yamllint/config.py (hunk 3 above)
Comment (adrienverge) [reply yamllint/config.py:29]: I really really prefer to avoid new dependencies. Good news is that `importlib.resources` is available for recent Python. For old versions, well, too bad...

**Review by adrienverge — COMMENTED** [review]
> I know my excitement is not a good reason to accept changes that are detrimental to other users, so hopefully we can find a compromise.

Sure. I have no problem with this change, as long as it does not affect anything else.

File: setup.cfg (hunk 2 above)
Comment (silverjam) [reply setup.cfg]: Ok, I added a new PR here to add support for Python 3.9 and drop support for Python 3.4: https://github.com/adrienverge/yamllint/pull/345

**Comment by silverjam** [comment]
@adrienverge I think I've addressed your concerns, but there's a slight drop in code coverage, and Travis tests seem to be succeeding, however the overall Travis run reports a failure.  Looking at the build output I don't see what's failing on Travis, any ideas there?

File: yamllint/config.py (hunk 3 above)
Comment (silverjam) [reply yamllint/config.py]: Ok, dropped the new dep entirely

**Comment by silverjam** [comment]
> @adrienverge I think I've addressed your concerns, but there's a slight drop in code coverage, and Travis tests seem to be succeeding, however the overall Travis run reports a failure. Looking at the build output I don't see what's failing on Travis, any ideas there?

I found the issue, it was just linter error I didn't see... fixed now.

**Comment by adrienverge** [comment]
Hello,

Can you please squash all your commit into one (with a nice description and explanation) and rebase on `master`? GitHub reports conflicts; when I try to solve them manually I encounter problems on `print("using importlib")`...

Your PR creates an empty file `yamllint/conf/__init__.py`: is there a use?

About `open_embedded_config()`: I'd prefer not creating yet another function, if it's only used by `get_extended_config_file()`. You can put the logic into `get_extended_config_file()`, testing whether `pkg_resources` is defined or not. And inside imports, for example (notice how imports are always ordered in yamllint):
```python

try:
    import importlib.resources as pkg_resources
except ImportError:
    pkg_resources = None  # for Python < 3.9
import os.path

import pathspec
import yaml
```


`get_extended_config_file()` used to return a filename, it now returns a file descriptor, and passes it direclty to `YamlLintConfig(content=...)` which expects a string. Can you rename the function to `get_extended_config()` and return a string directly (unless the file descriptor has a use elsewhere)?

silverjam:
> Can you please squash all your commit into one (with a nice description and explanation) and rebase on `master`? GitHub reports conflicts; when I try to solve them manually I encounter problems on `print("using importlib")`...
> 

Will do.

> 
> 
> Your PR creates an empty file `yamllint/conf/__init__.py`: is there a use?
> 

This is required for the pkg_resources library to work.  The `conf` dir needs to be a "real" module for the library to be able to locate it, from what I understand.

> 
> About `open_embedded_config()`: I'd prefer not creating yet another function, if it's only used by `get_extended_config_file()`. You can put the logic into `get_extended_config_file()`, testing whether `pkg_resources` is defined or not. And inside imports, for example (notice how imports are always ordered in yamllint):
> 
> ```python
> 
> # You should have received a copy of the GNU General Public License
> 
> # along with this program.  If not, see <http://www.gnu.org/licenses/>.
> 
> 
> 
> try:
> 
>     import importlib.resources as pkg_resources
> 
> except ImportError:
> 
>     pkg_resources = None  # for Python < 3.9
> 
> import os.path
> 
> 
> 
> import pathspec
> 
> import yaml
> 
> ```
> 
> 

Everything above here sounds doable, will do.


> 
> 
> 
> `get_extended_config_file()` used to return a filename, it now returns a file descriptor, and passes it direclty to `YamlLintConfig(content=...)` which expects a string. Can you rename the function to `get_extended_config()` and return a string directly (unless the file descriptor has a use elsewhere)?
> 
> 

I don't think this is possible.  The pkg_resources lib doesn't expose something returns a filename, since the resource may not have one (if it's embedded in a zip file for example).  I thought `YamlLintConfig(content=...)` was a form that expected a file descriptor? And the `file=` arg is the one that expects a filename string?

adrienverge:
> I don't think this is possible. The pkg_resources lib doesn't expose something that returns a filename

Sorry if I was unclear, but I was talking about the *content* as a string. Using `pkg_resources.read_text()` instead of `open_text()`, for instance.

By the way, looking at the docs, it seems that the importlib.resources lib *does* expose something that returns a path: https://docs.python.org/3/library/importlib.html#importlib.resources.path.

> This is required for the pkg_resources library to work. The `conf` dir needs to be a "real" module for the library to be able to locate it, from what I understand.

Same: after some research, there seems to be a one-line alternative that doesn't need another import nor `__init__.py` file:

```python
print(pkg_resources.files('yamllint') / 'conf' / 'default.yaml')
# .../lib/python3.9/site-packages/yamllint/conf/default.yaml
# (haven't tested with PyOxidizer but I guess it would return some kind of temp path?)
```

I see a last problem with this PR: when developing yamllint / running it from its source repo (e.g. with `python -m yamllint file.yaml`, while being inside of the Git repo), `pkg_resources.read_text()` fetches config from any installed yamllint package, but not `default.yaml` from the local code repo. This is a problem because when hacking/prototyping with the local repo code, for example playing with conf options, you'd expect *all* files to be taken from it (not just Python files, and config from elsewhere).
It seems that my `pkg_resources.files('yamllint') / 'conf' / name` proposal solves this problem, but please make sure, it's important.

silverjam:
Using `importlib.resources.files` doesn't work within PyOxidizer, since it seems `open_text` doesn't work for development mode (not sure why) it seems the pull request is in an unsolvable state (and doesn't provide any benefit for the intended use case).

silverjam:
@adrienverge OK, one last attempt here... code is a bit more complicated now, but changed things around so that __file__ is preferred over improtlib.resources, which should fix the development scenario and addressed the concerns over open_text vs read_text.

silverjam:
@adrienverge Any feedback?

adrienverge:
Hello @silverjam, thanks for the update. It seems to me that it complexifies the code a lot for the benefit it brings, so that the config-reading code is now harder to dive in, to understand and to debug. In the same time, PyOxidizer is still in beta, and the last release went out one year ago. Maybe it´s better to wait for it to be more "mature"?

silverjam:
> Hello @silverjam, thanks for the update. It seems to me that it complexifies the code a lot for the benefit it brings, so that the config-reading code is now harder to dive in, to understand and to debug. In the same time, PyOxidizer is still in beta, and the last release went out one year ago. Maybe it´s better to wait for it to be more "mature"?

OK, your call, but the complexity was added to minimize the risk to people not using PyOxidizer.  PyOxidizer is under active development, releases are done via cargo though (as far as I understand).  The latest release was a month ago: https://crates.io/crates/pyoxidizer.

silverjam:
@adrienverge Last call on this, if your interest has changed I can try to update the PR, I'll leave closed if I don't hear from  you.

adrienverge:
Hello @silverjam, thanks for the notice. My point of view is that PyOxidizer is still in beta, while this change would complexify the code a lot compared to the benefit.

PR #346, opened 2020-12-07
Allow only non-empty brackets/braces

rustygeldmacher:
We'd like to disallow brackets and braces in our YAML, but there's a
catch: the only way to describe an empty array or hash in YAML is to
supply an empty one (`[]` or `{}`). Otherwise, the value will be null.

This commit adds a `non-empty` option to `forbid` for brackets and
braces. When it is set, all flow and sequence mappings will cause errors
_except_ for empty ones.

### DIFF
--- a/tests/rules/test_braces.py
+++ b/tests/rules/test_braces.py
@@ -69,6 +69,14 @@ class ColonTestCase(RuleTestCase):
         self.check('---\n'
                    'dict: {}\n', conf)
         self.check('---\n'
+                   'dict: {\n'
+                   '}\n', conf)
+        self.check('---\n'
+                   'dict: {\n'
+                   '# commented: value\n'
+                   '# another: value2\n'
+                   '}\n', conf)
+        self.check('---\n'
                    'dict: {a}\n', conf, problem=(2, 8))
         self.check('---\n'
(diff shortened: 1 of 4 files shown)


rustygeldmacher:
@adrienverge Any idea why Travis hasn't run on this yet?

adrienverge:
Thanks @rustygeldmacher, a clean PR with tests and documentation is rare :+1:

Could add one or two tests with corner cases? For example I'm thinking about multiple lines:
```yaml
dict: {
}
dict: {
  # commented: value
}
```

> Any idea why Travis hasn't run on this yet?

Hum, no idea. I suggest repushing on GitHub to trigger a new build.

rustygeldmacher:
@adrienverge Thanks for taking a look! I'll add the test cases you suggest and push another commit. Just to double check -- the cases you point out should _pass_ if `forbid: non-empty` is specified, correct?

adrienverge:
> Just to double check -- the cases you point out should _pass_ if `forbid: non-empty` is specified, correct?

Yes! :+1:

rustygeldmacher:
@adrienverge -- test cases added, flake8 fixed. Does it look like everything's good?

adrienverge (approved):
Perfect!

PR #348, opened 2020-12-21
Support no warnings in config

kadaan:
Add support for including `no-warnings: true` or `no-warnings: false` in the config file.  If either the config file `no-warnings: true` or the cli option `-no-warnings` are set, then no-warnings mode will be enabled.


kadaan:
@adrienverge What do you think?

adrienverge:
Hello,

I'm not sure what to think because I dont' know what "the no-warnings mode" is. Your commit doesn't say it, neither does it add documentation to explain this option.

If it's somehow comparable to `--strict` (but with a different effect), the two options should appear consistent to the user. And I'm not favorable to adding another key-value option inside configuration files, if their effect can already be achieved within those files (by changing rules `level`).

kadaan:
`—no-warnings` is already a command line option that yamllint accepts. This simply allows it to be specified in the configuration file, rather than only on the command line.

adrienverge:
> `—no-warnings` is already a command line option that yamllint accepts.

Ah yes, sorry I forgot that.

I'm not favorable to adding redundant options inside configuration. Conf already allows removing rules if their output is not wanted. Command-line arguments `--strict` and `--no-warnings` are only *runtime switches* to temporarily alter yamllint output.

kadaan:
Hmmmm.  That is unfortunate. We are trying to use this as part of our build and need no warnings enabled. If that was all, we wouldn’t need this feature. But, we are also trying to have Atom perform the same linking.  Because this cannot be controlled via the config file, it is not possible to get the same behavior in both scenarios without manual configuration.

Issue #351, opened 2021-01-05
Bug: Confict between required quoted strings and explicit octal


scolby33:
There appears to be a conflict between `octal-values.forbid-explicit-octal = false` and `quoted-strings.required = true`.

## Setup

.yamllint.yaml:
```yaml
---
rules:
  octal-values:
    forbid-explicit-octal: false
  quoted-strings:
    required: true
...
```

test.yaml:
```yaml
---
foo: 0o100
...
```

```console
$ yamllint test.yaml
test.yaml
  2:6       error    string value is not quoted with any quotes  (quoted-strings)
```

## What Happens

yamllint reports an error because it thinks the value of the `foo` key is a string.

## What I Expect to Happen

yamllint should accept the value of the `foo` key as an explicit octal integer equal to decimal 64.

adrienverge:
Hello, thanks for the report. The problem impacts `quoted-strings` only (not `octal-values`).

It's due to the lib we're using (PyYAML) that implements the YAML specification version 1.1, not 1.2. Hence, values starting with `0o` are not considered as numbers: they are just strings, so they need quotes when `quoted-strings: {required: true}`.

See:
```python
>>> import yaml
>>> yaml.resolver.Resolver().resolve(yaml.nodes.ScalarNode, '100', (True, False))
'tag:yaml.org,2002:int'
>>> yaml.resolver.Resolver().resolve(yaml.nodes.ScalarNode, '0100', (True, False))
'tag:yaml.org,2002:int'
>>> yaml.resolver.Resolver().resolve(yaml.nodes.ScalarNode, '0o100', (True, False))
'tag:yaml.org,2002:str'
```
However there seems to be a solution, which I implement in https://github.com/adrienverge/yamllint/pull/354. Could you please try the pull request and report back?

scolby33:
That makes sense! I actually played around with a similar REPL session where I noticed that pyyaml does indeed deserialize `0o100` to a string, and that ruamel.yaml deserialized it to an integer, but I thought that since the `explicit-octal` setting exists that it should work.

Thanks for your quick response; I tested locally both with the example I posted and the (more complicated) YAML document that first brought this to my attention and the fix works for me!

adrienverge:
Thanks for testing @scolby33 :+1:

PR #354, opened 2021-01-05
quoted-strings: Fix explicit octal recognition

adrienverge:
PyYAML implements YAML spec version 1.1, not 1.2. Hence, values starting
with `0o` are not considered as numbers: they are just strings, so they
need quotes when `quoted-strings: {required: true}`.
```python
>>> import yaml
>>> yaml.resolver.Resolver().resolve(yaml.nodes.ScalarNode, '100', (True, False))
'tag:yaml.org,2002:int'
>>> yaml.resolver.Resolver().resolve(yaml.nodes.ScalarNode, '0100', (True, False))
'tag:yaml.org,2002:int'
>>> yaml.resolver.Resolver().resolve(yaml.nodes.ScalarNode, '0o100', (True, False))
'tag:yaml.org,2002:str'
```
Let's try to prevent that.

Fixes https://github.com/adrienverge/yamllint/issues/351.

### DIFF
--- a/tests/rules/test_quoted_strings.py
+++ b/tests/rules/test_quoted_strings.py
@@ -436,3 +436,21 @@ class QuotedTestCase(RuleTestCase):
                    '- foo bar\n'
                    '- "foo bar"\n',
                    conf, problem1=(3, 3), problem2=(7, 3), problem3=(11, 3))
+
+    def test_octal_values(self):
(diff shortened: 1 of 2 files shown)


DvalCaruso (approved):
awesome

DvalCaruso (approved):

PR #358, opened 2021-02-09
setuptools is a runtime dep only for Python < 3.8

polyzen:
> In recent versions of setuptools and Python, console-script entry
points are using stdlib importlib by default, thus setuptools is no
longer a runtime dependency.

pypa/setuptools#2197
https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v4730
https://docs.python.org/3/library/importlib.metadata.html

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,7 +52,7 @@ include_package_data = True
 install_requires =
   pathspec >= 0.5.3
   pyyaml
-  setuptools
+  setuptools; python_version < "3.8"
 
 test_suite = tests
 


adrienverge:
Do you have a pointer to this statement in Python docs or other "official" docs?

polyzen:
Was curious myself, thanks to @felixonmars the commit has been updated with the information.

adrienverge:
Thanks for this extra information @polyzen.

I have read the changelog entry, overviewed the PR at setuptools and the `importlib.metadata` docs. However I still don't understand why `setuptools` is no longer needed as a dependency? Especially since we explicitly import it here: https://github.com/adrienverge/yamllint/blob/0fff4e2/setup.py#L17
A few more details would be welcome :)

polyzen:
On February 10, 2021 3:11:33 AM EST, "Adrien Vergé" <notifications@github.com> wrote:
> Thanks for this extra information @polyzen.
> 
> I have read the changelog entry, overviewed the PR at setuptools and
> the `importlib.metadata` docs. However I still don't understand why
> `setuptools` is no longer needed as a dependency? Especially since we
> explicitly import it here:
> https://github.com/adrienverge/yamllint/blob/0fff4e2/setup.py#L17
> A few more details would be welcome :)

`install_requires` is for runtime deps. setuptools is only needed for building now, unless you have a setuptools older than that release or python < 3.8.

--
Best,
Daniel <https://danielcapella.com>

adrienverge:
> `install_requires` is for runtime deps. setuptools is only needed for building now, unless you have a setuptools older than that release or python < 3.8.

Yamllint supports Python versions < 3.8. Will it work to run a packaged yamllint, without setuptools installed, on Python 3.5?

I'm sorry to ask details again, but I've re-reviewed the 3 links you posted, and don't find how they relate to this. Maybe you could copy-paste the relevant part here?

polyzen:
Updated the commit to keep setuptools as a runtime dep for Python < 3.8.

Entry points used to require the pkg_resources module, which is provided by setuptools. This was slow, so now importlib.metadata is used if possible:
https://github.com/pypa/setuptools/blob/e1ffc2abbae4f2aa78dd09ee9827d754b7702b7b/setuptools/command/easy_install.py#L2045
https://github.com/pypa/setuptools/pull/2194

adrienverge (approved):
Thanks @polyzen!

PR #370, opened 2021-03-22
line_length: skip all hash signs starting comment

pmalek-sumo:
Whenever a comment start with more than 1 `#` then using the following config

```yaml
rules:
  line-length:
    max: 20
    allow-non-breakable-words: true
```

to lint the following document:

```yaml
## http://www.verylongurlurlurlurlurlurlurlurl.com
key:
  subkey: "value"
```

will disregard the `max` property and issue an error/warning anyway.

This PR addresses that by jumping not 2 but `length_of_hash_signs_at_line_begining + 1`

### DIFF
--- a/tests/rules/test_line_length.py
+++ b/tests/rules/test_line_length.py
@@ -120,11 +120,22 @@ class LineLengthTestCase(RuleTestCase):
         self.check('---\n'
                    '# http://www.verylongurlurlurlurlurlurlurlurl.com\n'
                    'key:\n'
-                   '  value:\n', conf)
+                   '  subkey: value\n', conf)
         self.check('---\n'
                    '## http://www.verylongurlurlurlurlurlurlurlurl.com\n'
                    'key:\n'
-                   '  value:\n', conf)
+                   '  subkey: value\n', conf)
+        self.check('---\n'
+                   '# # http://www.verylongurlurlurlurlurlurlurlurl.com\n'
+                   'key:\n'
+                   '  subkey: value\n', conf,
+                   problem=(2, 21))
+        self.check('---\n'
+                   '#A http://www.verylongurlurlurlurlurlurlurlurl.com\n'
+                   'key:\n'
+                   '  subkey: value\n', conf,
+                   problem1=(2, 2, 'comments'),
+                   problem2=(2, 21, 'line-length'))
 
         conf = ('line-length: {max: 20, allow-non-breakable-words: true}\n'
                 'trailing-spaces: disable')
--- a/tests/rules/test_line_length.py
+++ b/tests/rules/test_line_length.py
@@ -116,6 +116,16 @@ class LineLengthTestCase(RuleTestCase):
                    'long_line: http://localhost/very/very/long/url\n'
                    '...\n', conf, problem=(2, 21))
 
+        conf = 'line-length: {max: 20, allow-non-breakable-words: true}'
+        self.check('---\n'
+                   '# http://www.verylongurlurlurlurlurlurlurlurl.com\n'
+                   'key:\n'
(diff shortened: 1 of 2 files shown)


tests/rules/test_line_length.py
```diff
@@ -116,6 +116,16 @@ def test_non_breakable_word(self):
...
 
+        conf = 'line-length: {max: 20, allow-non-breakable-words: true}'
+        self.check('---\n'
+                   '# http://www.verylongurlurlurlurlurlurlurlurl.com\n'
+                   'key:\n'
+                   '  value:\n', conf)
+        self.check('---\n'
+                   '## http://www.verylongurlurlurlurlurlurlurlurl.com\n'
+                   'key:\n'
+                   '  value:\n', conf)
```
adrienverge, line 127: Can you add these positive counterparts?
```python
        self.check('---\n'
                   '# # http://www.verylongurlurlurlurlurlurlurlurl.com\n'
                   'key:\n'
                   '  value:\n', conf, problem=(2, 21))
        self.check('---\n'
                   '#A http://www.verylongurlurlurlurlurlurlurlurl.com\n'
                   'key:\n'
                   '  value:\n', conf, problem=(2, 21))
```

yamllint/rules/line_length.py
```diff
@@ -140,7 +140,11 @@ def check(conf, line):
                 start += 1
 
             if start != line.end:
-                if line.buffer[start] in ('#', '-'):
+                if line.buffer[start] in ('#'):
```
adrienverge, line 143: In Python, `in ('#')` is equal to `in '#'`, which works but doesn't make sense here.
Instead you can use `== '#'`.

Same for `in ('-')`.

yamllint/rules/line_length.py
```diff
@@ -140,7 +140,11 @@ def check(conf, line):
...
+                    idx = line.buffer.find(' ', start, line.end)
```
adrienverge, line 144: I don't like this approach because it doesn't do what your commit title says: `skip all hash signs starting comment`. The current implementation allows **any** non-space character after `#`, which includes chars like `A` or `\n` or `🤙`.

Maybe something like:
```python
if line.buffer[start] == '#':
    while line.buffer[start] == '#':
         start += 1
    start += 1
elif line.buffer[start] == '-':
```
?

adrienverge (commented):
Thanks for this clean and tested commit!

Please see my remarks above.

tests/rules/test_line_length.py
pmalek-sumo: Done, please check

yamllint/rules/line_length.py
pmalek-sumo: Done, please check

yamllint/rules/line_length.py
pmalek-sumo: Done, please check

pmalek-sumo:
For some reason the tests are failing. Seems unrelated to this PR

adrienverge (approved):
> For some reason the tests are failing. Seems unrelated to this PR

Indeed. I solved this in https://github.com/adrienverge/yamllint/pull/372, so if you rebase on `master` it should work again.

pmalek-sumo:
Rebased it. Thanks @adrienverge !

adrienverge:
Thanks for your contribution @pmalek-sumo!

pmalek-sumo:
Any chance to release a patch version to include this fix?

adrienverge:
Sure, released in [version 1.26.1](https://github.com/adrienverge/yamllint/releases/tag/v1.26.1). Sorry about the delay!

PR #378, opened 2021-04-29
Use cleaned file path in logged output

### LINKED ISSUE #377: `github` output format on Github hides filenames and lineno
Using `github` or `auto` output format under GH Actions CI will not show file names or line numbers.
Screen Shot 2021-04-27 at 3 44 20 PM

Forcing `standard` will yield something usable:
image

Using yamllint 1.26.1

raphael-riel:
Fixes #377 

Shown file's path should use the cleaned version, without the leading `./`.
Otherwise it confuses Github which cannot fin the file on which to put the annotation.

Every method using the parameter are actually expecting a string. The file pointer, when casted to string was yielding an actual filename, but with the leading `./`.

### DIFF
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -207,7 +207,7 @@ def run(argv=None):
         except EnvironmentError as e:
             print(e, file=sys.stderr)
             sys.exit(-1)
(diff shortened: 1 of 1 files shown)


raphael-riel:
No bug. Nothing to see here.

PR #388, opened 2021-06-10
setup: update python_requires to comply with PEP 345/440

### LINKED ISSUE #387: value of `python_requires` does not comply with  PEP 345/440
https://github.com/adrienverge/yamllint/blob/85ccd625a382423edd988fcb796c4dd5897886fb/setup.cfg#L49

According to [PEP 345 Requires-Python](https://www.python.org/dev/peps/pep-0345/#requires-python), the value of this field must be a valid [Version Specifier](https://www.python.org/dev/peps/pep-0345/#version-specifiers). Which in turn expects this to comply with [PEP 440](https://www.python.org/dev/peps/pep-0440/).

While not an issue for those that directly use `pip`, this will cause issues for `poetry` users in the next release (if their current stance is maintained). Discussion of the issue and there stance can be found here: https://github.com/python-poetry/poetry/issues/4095.

ITProKyle:
According to [PEP 345 Requires-Python](https://www.python.org/dev/peps/pep-0345/#requires-python), the value of this field must be a valid [Version Specifier](https://www.python.org/dev/peps/pep-0345/#version-specifiers). Which in turn expects this to comply with [PEP 440](https://www.python.org/dev/peps/pep-0440/).

While not an issue for those that directly use `pip`, this will cause issues for `poetry` users in the next release (if their current stance is maintained). Discussion of the issue and there stance can be found here: https://github.com/python-poetry/poetry/issues/4095.

resolves #387

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -46,7 +46,7 @@ project_urls =
 [options]
 packages = find:
 
-python_requires = >=3.5.*
+python_requires = >=3.5
 
 include_package_data = True
 install_requires =


adrienverge (approved):
Thanks, it looks good!

Can you please add an explanation *inside the commit message*, e.g. copy-paste the description from https://github.com/adrienverge/yamllint/issues/387? (it wasn't really needed to open an issue if you solve it directly).

ITProKyle:
@adrienverge - sure thing. updated commit message pushed.

adrienverge:
Thank you @ITProKyle.

Issue #399, opened 2021-08-05
Ignoring file that is a broken symbolic link produces "[Errno 2] No such file or directory" 


thomasdstewart:
Hi,

When running yamllint on a directory of files if a file is a broken symbolic link it produces "[Errno 2] No such file or directory" even if the file is set to be ignored. I was able to reproduce using latest yamllint[0] along with a small test environment[1] and by running "yamllint ." to show the error[2].

Kind Regards
Tom

[0]
```
$ python3 -m venv .venv
$ source .venv/bin/activate
$ pip install yamllint
$ yamllint --version
yamllint 1.26.2
$
```

[1]
```
$ mkdir test; cd test
$ ln -s not_there ignore_file.yaml
$ ls -l ignore_file.yaml 
lrwxrwxrwx 1 thomas thomas 9 Aug  5 10:45 ignore_file.yaml -> not_there
$ cat .yamllint 
---
extends: default

ignore: |
  ignore_file.yaml
$
```

[2]
```
$ yamllint .
[Errno 2] No such file or directory: './ignore_file.yaml'
$
```

adrienverge:
Hello @thomasdstewart, thanks for the report.

I can confirm the problem. Any clean contribution to fix this (and include non-regression tests) would be welcome!

Out of curiosity: do you have a use-case where you have broken `.yaml` symlinks, and want to keep them?

thomasdstewart:
Hi @adrienverge,

Indeed filing an issue is a bit strong wording for what is more like a feature request from some random on the Internet and by far not a demand for you to fix. :)

Apologies, yes some context would have been useful. I've got a public git repo with a set of Ansible playbooks, roles, templates, vars, vaults, etc. The vault files are in the standard locations but are symlinks to the real files in a private repo, eg separating the public and private data.  eg:

$ file group_vars/all/vault.yml
group_vars/all/vault.yml: symbolic link to ../../safe/group_vars/all/vault.yml
$ file safe/group_vars/all/vault.yml
safe/group_vars/all/vault.yml: ASCII text
$

To side step this, in my CI I just swapped out "yamllint ." with: find . -type f \( -name "*.yml" -o -name "*.yaml" \) ! -name 'vault.yml' | xargs yamllint

Kind Regards
Tom

thomasdstewart:
[...]

PR #400, opened 2021-08-05
Initial fix for issue 399

thomasdstewart:
First look at #399, sorry I've not written any test's, but it still seems to pass al existing ones.

### DIFF
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -201,6 +201,8 @@ def run(argv=None):
 
     for file in find_files_recursively(args.files, conf):
         filepath = file[2:] if file.startswith('./') else file
+        if(conf.is_file_ignored(filepath) or None):
+            continue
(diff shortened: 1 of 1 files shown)


yamllint/cli.py
```diff
@@ -201,6 +201,8 @@ def run(argv=None):
 
     for file in find_files_recursively(args.files, conf):
         filepath = file[2:] if file.startswith('./') else file
+        if(conf.is_file_ignored(filepath) or None):
```
DimitriPapadopoulos, line 204:
```suggestion
        if conf.is_file_ignored(filepath):
```
I don't understand how `or None` would help.

DimitriPapadopoulos: Perhaps change the commit text to:
```
cli: match symlinks to ignore early on, before following them
```

PR #412, opened 2021-10-09
Add support for Python 3.10, drop Python 3.5

bonddim:
- Add support for Python 3.10 released on 2021-10-04
- Drop support for Python 3.5 since it has reached end-of-life
- Fix github actions workflow:
  - install correct Python version
  - run tests after linting
  - add ~/.local/bin to system path for both jobs

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,8 +17,6 @@ jobs:
         uses: actions/checkout@v2
       - name: Set up Python
         uses: actions/setup-python@v2
-        with:
-          python-version: 3.9
       - run: python -m pip install flake8 flake8-import-order doc8 sphinx
       - run: python -m pip install .
       - run: flake8 .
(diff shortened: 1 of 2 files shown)


bonddim:
Hello @adrienverge ,
Sure, I can split this PR, if needed.
The main change here matches PR title - drop support of python 3.5 and add 3.10. But I realized, that current workflow doesn't run tests with required python versions.
About workflow:

- `fail-fast: false`  - to not stop all jobs if one fails  - will provide possibility to see the test status on other versions
- `needs: lint` - tests will not run if linting fails
- `echo "$HOME/.local/bin" >> $GITHUB_PATH` - avoid hard-coded values where possible
- about `nightly` version you may see in my previous comment

I may remove step to append system path on **Lint** job, but there is step to install current project version also.
Also `with: {python-version: ...}` could be completely removed from **Lint** to use latest from *actions/setup-python@v2*

adrienverge:
Hey,

> * `fail-fast: false`  - to not stop all jobs if one fails  - will provide possibility to see the test status on other versions
> * `needs: lint` - tests will not run if linting fails
> * `echo "$HOME/.local/bin" >> $GITHUB_PATH` - avoid hard-coded values where possible
> * about `nightly` version you may see in my previous comment

Can you include these in the 2nd commit message?

- About `fail-fast`: OK.
- About `needs: lint`: I would prefer running them anyway.
- I understand using `$HOME` instead of `/home/runner`, but there doesn't seem to be a need to add it on lines 22-23 (in the *lint* job), as you point out. I prefer removing it. The step to install current project version is needed to be able to run `yamllint` on line 26.
- About `nightly`, yes got it, OK :+1: 

> Also `with: {python-version: ...}` could be completely removed from **Lint** to use latest from _actions/setup-python@v2_

Sure, why not.

adrienverge:
Thanks a lot @bonddim! (and sorry for the delay)

Issue #361, opened 2021-02-23
Rule 'comments' should ignore Cloud-Init "#cloud-config" similar to shebangs


lwbt:
I noticed that shebangs can be configured to be ignored in recent versions. Cloud-Init has a similar string at the start of a document which should be also ignored: `#cloud-config`

https://cloudinit.readthedocs.io/en/latest/topics/examples.html?highlight=%23cloud-config

adrienverge:
It makes sense. However there will probably be other magic values `#foo` in the future. Maybe the best would be to specify the list of regex to ignore?
```yaml
rules:
  comments:
    require-starting-space: true
    ignore-regex:
      - ^#!
      - ^#cloud-config
    min-spaces-from-content: 2
```

lwbt:
That would be even better, yes.

james-callahan:
Another example of this would be [sops](https://github.com/mozilla/sops) encrypted comments which look like:
```
    #ENC[AES256_GCM,data:XXXXXXXXXXXXXXX,iv:YYYYYYYYYYYYYYYY,tag:ZZZZZZZZZZZZZZ,type:comment]
```

adrienverge:
Closing in favor of https://github.com/adrienverge/yamllint/issues/545.

PR #417, opened 2021-10-26
Support ignoring cloud-init #cloud-config directives

### LINKED ISSUE #416: Handle the special #cloud-config to support cloud-init
[Cloud-init requires some configuration files to start with #cloud-config](https://cloudinit.readthedocs.io/en/latest/topics/format.html#cloud-config-data), violating the `rules.comments.require-starting-space: true`.

This is very similar to the #116 issue, fixed in #151.

Would implementing another exception be feasible in this case?

The workaround would be to disable that rule for the first line (which I'm still tuning to make it work), but it's suboptimal because you have to disable that check on each file.

Example:

```yaml
#cloud-config # yamllint disable-line rule:comments
```

Thanks!

ferrarimarco:
[Cloud-init requires some configuration files to start with #cloud-config](https://cloudinit.readthedocs.io/en/latest/topics/format.html#cloud-config-data), violating the `rules.comments.require-starting-space: true`.

This PR adds support for ignoring those directives.

Fixes #416

### DIFF
--- a/tests/rules/test_comments.py
+++ b/tests/rules/test_comments.py
@@ -80,6 +80,23 @@ class CommentsTestCase(RuleTestCase):
                    problem3=(9, 2), problem4=(10, 4),
                    problem5=(15, 3))
 
+    def test_cloud_init_cloud_config(self):
+        conf = ('comments:\n'
+                '  require-starting-space: true\n'
+                '  ignore-cloud-init-cloud-config: false\n'
+                'document-start: disable\n')
+        self.check('#cloud-config\n',
+                   conf, problem1=(1, 2))
+        self.check('# cloud-config\n', conf)
+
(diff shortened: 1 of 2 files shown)


doyou89:
@adrienverge

adrienverge:
Hello @doyou89, no, as explained in https://github.com/adrienverge/yamllint/issues/416#issuecomment-951981783, specific options for specific software won't likely be embedded in yamllint.

However, an `ignore-regex` config option would be great because it solves the problem for `#cloud-config`, as well as for other requests from users ([latest example here](https://github.com/adrienverge/yamllint/issues/361#issuecomment-1216250893) for `#ENC...`).

So contributions are welcome! Let's follow this in https://github.com/adrienverge/yamllint/issues/361.

PR #419, opened 2021-11-02
output-format: Add support for Checkstyle

hunts:
[Checkstyle](https://github.com/linkedin/pygradle/blob/master/pygradle-plugin/src/test/resources/checkstyle/checkstyle.xsd) format is useful to CI toots that support it to display the linting results in a nice way. E.g. TeamCIty support it: https://confluence.jetbrains.com/display/TCD10/XML+Report+Processing

Feels like we might need to refactor the code of how we display problems in a better way, but in this pull request I'll make it straight forward without adding too much changes.

### DIFF
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -22,6 +22,7 @@ import locale
 import os
 import platform
 import sys
+from xml.sax.saxutils import escape
 
 from yamllint import APP_DESCRIPTION, APP_NAME, APP_VERSION
 from yamllint import linter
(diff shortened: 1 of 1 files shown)


adrienverge:
Hello, thanks for contributing to yamllint.

There are plenty of output formats, each with advantages and drawbacks. This repo cannot maintain them all, so new ones are not likely to be accepted (except for JSON or YAML format).

What about writing a tiny script to convert yamllint output to Checkstyle format?

hunts:
Agreed. Have an external script to parse the **parsable** output is more maintainable. But build-in support is nicer because it avoid users to pull another dependencies.

The **parsable** output does not include the **rule** that detects a problem, do you think we should add the field, or may be better to introduce the JSON output format that include all fields?

adrienverge:
> The **parsable** output does not include the **rule** that detects a problem

Are you sure? I just tried and the rule is correctly printed.

hunts:
Oops, sorry my bad. Didn't realize the problem.message contains both desc and rule. 

Thank you for taking time to review, I'll close this pr.

hunts:
For anyone looking for checkstyle output format, I've created a converter https://github.com/hunts/yamllint-to-checkstyle

Issue #395, opened 2021-07-20
yamllint allows referencing a non-existant anchor 


mckaymatt:
yamllint 1.26.1 (latest) does not flag references to anchors that don't exist.

```yaml
---
foo: &foo
    keyA: value1

bar:
    <<: *no_anchor_with_this_name
    keyB: value2
```

If you try to parse this using a tool like PyYAML it errors.

```
  File ".local/lib/python3.8/site-packages/ruamel/yaml/composer.py", line 115, in compose_node
    raise ComposerError(
ruamel.yaml.composer.ComposerError: found undefined alias 'no_anchor_with_this_name'
  in "test.yaml", line 6, column 9

```

adrienverge:
Hello,

Indeed yamllint does not currently detect inexistent anchors, or unmatching aliases.

It could be implemented in a new yamllint rule. In theory, it should be possible if aliases aren't used before anchors are declared (and this seems to be the case according to the [YAML specification 6.9.2.](https://yaml.org/spec/1.2/spec.html#id2785586): "*An anchor marks a node for *future* reference*").

amimas:
I'd also like to check if there are unused anchors in the yaml files. Is this rule being considered for yamllint?

adrienverge:
Yes, I think this is within the scope of a YAML linter. An option like `forbid-unused-anchors` could do the job.

In the end, a new rule like this would be great:

```yaml
rules:
  anchors:
    forbid-unknown-aliases: true      # true by default, because YAML spec fordids it
    forbid-duplicated-anchors: false  # false by default, because YAML spec allows it
    forbid-unused-anchors: false      # false by default, because YAML spec allows it
```

amimas:
That's good to hear @adrienverge  . I like the example you gave. It gives more flexibility depending on the use case.

Not sure when we can expect those rules. I'm not familiar with the code base in this project and haven't implemented any custom rule. I can try open a PR if you can give some hints or guides on how to build those rules.

adrienverge:
Answering https://github.com/adrienverge/yamllint/pull/420#discussion_r1096388404 from @amimas, about the `forbid-unused-anchors` option.

> Maybe we should discuss it in the feature request issue.

Sure :+1:

> I understand what you're saying and it makes sense. I was just asking because I wanted to look at the `forbid-unused-anchors` option. Without checking the entire file, not sure how to verify that an anchor isn't being used. Let me know if you have any suggestions.

[...]

PR #420, opened 2021-11-08
Check for missing or duplicated anchors

newrushbolt:
Basic anchor\aliases checks, also fix for https://github.com/adrienverge/yamllint/issues/395.


tests/rules/test_anchors.py
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
```
adrienverge, line 1: This line shouldn't be needed now that yamllint only supports Python 3. Can you remove it (along with the one from `yamllint/rules/anchors.py`)?

tests/rules/test_anchors.py
```diff
@@ -0,0 +1,117 @@
...
+        self.check(HIT_ANCHOR_POINTER, conf)
+        self.check(HIT_ANCHOR_MERGE, conf)
+        self.check(MULTI_HIT_ANCHOR_POINTER, conf)
+        self.check(MULTI_DOC_HIT_ANCHOR_POINTER, conf)
+
+        self.check(MISS_ANCHOR_POINTER, conf,
+                   problem=(3, 11))
+        self.check(MISS_ANCHOR_MERGE, conf,
+                   problem_first=(5, 7),
+                   problem_second=(5, 7))
```
adrienverge, line 110: To keep consistency with all other tests, can you use the following?
```suggestion
                   problem1=(5, 7), problem2=(5, 7))
```

yamllint/rules/anchors.py
```diff
@@ -0,0 +1,117 @@
...
+    target_address: *address
+
+   the following code snippet would **FAIL**:
+   ::
+
+    address: &address |
+      Williams St. 13
+    target_address: *wrong_address
+
+#. This appends to merges as well:
```
adrienverge, line 38: The convention is to use `#.` for configuration examples, not for plain English explanations.

I suggest to remove this line (same for next occurence), and after keep only the `FAIL` examples.

yamllint/rules/anchors.py
```diff
@@ -0,0 +1,117 @@
...
+from yaml import DocumentStartToken, AnchorToken, AliasToken, BlockEndToken
+
+from yamllint.linter import LintProblem
+
+
+ID = 'anchors'
+TYPE = 'token'
+
+
+def check(__conf, token, __prev, __next, __nextnext, context):
```
adrienverge, line 92: I guess these `__` are for unused variables. Other rules don't use that, so I prefer keeping the same code for all.

yamllint/rules/anchors.py
```diff
@@ -0,0 +1,117 @@
...
+    if isinstance(token, (DocumentStartToken)):
+        context['anchors'] = set()
+        context['aliases'] = {}
+    elif ('anchors' in context and 'aliases' in context):
```
adrienverge, line 96: 1. Other rules do it more simply: drop the `if/else`, save some indentation, and just:
   ```python
       if 'anchors' not in context:
           context['anchors'] = {}
           context['aliases'] = {}
   ```
2. Is `set()` different from `{}`?

yamllint/rules/anchors.py
```diff
@@ -0,0 +1,117 @@
...
+        if isinstance(token, (AnchorToken)):
+            if token.value in context['anchors']:
+                yield LintProblem(
+                    token.start_mark.line + 1, token.start_mark.column + 1,
+                    f"anchor '{token.value}' duplicates in document")
```
adrienverge, line 101: I propose:
```suggestion
                    f'duplicated anchor "{token.value}"')
```

(notice the usage of `"` instead of `'` inside the printed string: that's what other rules already do)

yamllint/rules/anchors.py
```diff
@@ -0,0 +1,117 @@
...
+            context['anchors'].add(token.value)
+        elif isinstance(token, (AliasToken)):
+            if token.value not in context['aliases']:
+                context['aliases'][token.value] = []
+            alias_obj = {"line": token.start_mark.line,
+                         "column": token.start_mark.column}
+            context['aliases'][token.value].append(alias_obj)
+        elif isinstance(token, (BlockEndToken)):
```
adrienverge, line 109: I'm not sure I understand this part.
1. Why `BlockEndToken`? Did you mean `DocumentEndToken`? (There can be many `BlockEndToken` inside a document.)
2. You don't need to wait for the end of the document to report missing anchors. You can print it as soon as the unknown alias is found, because the YAML specification states: *It is an error for an alias node to use an anchor that does not previously occur in the document*.

yamllint/rules/anchors.py
```diff
@@ -0,0 +1,117 @@
...
+            missing_anchors = [alias for alias in list(context['aliases'])
+                               if alias not in context['anchors']]
+            if len(missing_anchors) > 0:
+                for miss_anchor in missing_anchors:
+                    for location in context['aliases'][miss_anchor]:
+                        yield LintProblem(
+                            location['line'] + 1, location['column'] + 1,
+                            f"anchor '{miss_anchor}' is not found in document")
```
adrienverge, line 117: I propose:
```suggestion
                            f'anchor "{miss_anchor}" was not declared')
```

adrienverge (commented):
Hello @newrushbolt, thanks for contributing!

The change is great and I'm favorable to merging it, but before I have a few remarks and concerns.

1. According to the YAML specification:
   - *It is an error for an alias node to use an anchor that does not previously occur in the document* -- https://yaml.org/spec/1.2/spec.html#id2785586

     Your commit implements that :+1:
     But it waits for the end of the document to report it, whereas it's not needed, and can output errors in a wrong order. Since anchors *must* be declared before aliases, you should report the problem as soon as the unknown alias is found.

   - *The alias refers to the most recent preceding node having the same anchor* → it's allowed to override anchor -- https://yaml.org/spec/1.2-old/spec.html#id2786448

     Your commit implements and documents "*Duplicate anchors are not allowed*", which disagrees with the spec.

2. To solve the previous point, and also to prepare the future (if we want to allow the rule to be extended later), I propose to make the rule configurable:
   ```yaml
   rules:
     anchors:
       forbid-unknown-aliases: true  # true by default, because YAML spec fordids it
       forbid-duplicated-anchors: false  # false by default, because YAML spec allows it
   ```

   The Python file would look like:
    ```python
    CONF = {'forbid-unknown-aliases': bool,
            'forbid-duplicated-anchors': bool}
    DEFAULT = {'forbid-unknown-aliases': True,
               'forbid-duplicated-anchors': False}
    
    if conf['forbid-unknown-aliases']:
    ```

3. I'm not sure whether this new rule (at least `forbid-unknown-aliases`) should be enabled in the `default` set of rules: it might break things for many users.  On the other hand, broken anchors would mean broken YAML and shouldn't exist...

   An option could be to enable it, release a new yamllint version, and if this causes problems for users, you would quickly create a pull request to disable it by default.

   What do you think?

newrushbolt:
Hello @adrienverge, thanks for the review. Sorry for the delay.
Read all the remarks, and realized that this PR had an infinite number of ways to be improved 🙈.

First of all, I decided to split this into two separate rules.
I did some rework, feel free to continue review.

1. Absolutely
2. Sure
3.  As well as `unknown-aliases` is correlates with YAML spec I think it should be enabled by default, and we should add a couple of words in the release note about it.
And I don't think we should revert this behavior even if the sh#t hits the fan after release. 
Talking about `anchor-duplicates`, it should probably stay disabled.

_Of course l'll squash all the commits after final approve._

yamllint/rules/anchors.py
newrushbolt: I'd suggest to keep this syntax to highlight that variables is unused, and "declared" only in order to comply with the function args order.
It seems to be pretty intuitive, and also keeps linters\IDE from warning about unused variables.

yamllint/rules/anchors.py
newrushbolt: 2. Replaced `set()` with `[]`.

yamllint/rules/anchors.py
newrushbolt: Yes, indeed, I should use `DocumentEndToken`.

yamllint/rules/anchors.py
newrushbolt: Replaced it with the following message which seems to be more helpful
`f'anchor "{token.value}" is used before assignment'`

yamllint/rules/unknown_aliases.py
```diff
@@ -0,0 +1,82 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Use this rule to prevent aliases pointing to pointers that don't exist.
```
adrienverge, line 17: I propose:
```suggestion
Use this rule to prevent aliases referencing pointers that don't exist.
```
to avoid repetition?

yamllint/rules/anchors.py
adrienverge: I don't know, why not but in this case it should be done for all yamllint, ideally enforced by `flake8`, and most of all done in a separated *refactor* PR. In the meantime, can you please remove them to make the code consistent? Maybe it's not obvious, but consistent code makes the PR / review process easier for everyone.

yamllint/rules/anchors.py
adrienverge: What about point 1.? It still applies to current PR code.

adrienverge (commented):
Hello Sergei,

Thanks for the good work!

It looks better but I see a problem now: you split the feature into 2 rules, whereas:
- they're both related to anchors,
- it feels strange to have a rule `anchor-duplicates` containing an option
  with the same name `anchor-duplicates`. Same for `unknown-aliases:
  {unknown-aliases: true}`.

I insist on the following conf, inside a united `anchors` rule:
```py
CONF = {'forbid-unknown-aliases': bool,
        'forbid-duplicated-anchors': bool}
DEFAULT = {'forbid-unknown-aliases': True,
           'forbid-duplicated-anchors': False}

if conf['forbid-unknown-aliases'] and AliasToken and token.value not in context['anchors']:
    # anchor "{token.value}" is used before assignment
if conf['forbid-duplicated-anchors'] and AnchorToken and token.value in context['anchors']:
    # duplicated anchor "{token.value}"
```
which is more in the spirit of yamllint. Moreover it would allow extending the rule with more options later.

You can see rules [`empty-values`](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.empty_values) and [`octal-values`](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.octal_values) for comparison.

Other than that, please address all my other remarks above and below, so we can merge this :+1:

newrushbolt:
@adrienverge is everything ok now?
If so can you merge this request?

yamllint/rules/anchors.py
```diff
@@ -0,0 +1,124 @@
...
+        if isinstance(token, DocumentStartToken):
+            context['anchors'] = []
+        elif isinstance(token, AnchorToken):
+            context['anchors'].append(token.value)
+
+    if conf['forbid-unknown-aliases'] and isinstance(token, AliasToken):
+        if token.value not in context['anchors']:
+            yield LintProblem(
+                token.start_mark.line + 1, token.start_mark.column + 1,
+                f'anchor "{token.value}" is used before assignment')
```
amimas, line 116: Should this message also indicate a potentially reference to a non-existent anchor?

amimas (commented):
Thanks so much for this PR. Looking forward seeing this merged and released.

In #395, a third option for this rule was also discussed (`forbid-unused-anchors`). It's similar to the duplicated anchors option. Can that option be added into this PR. I feel that will make this whole.

Thanks again.

yamllint/rules/anchors.py
```diff
@@ -0,0 +1,124 @@
...
+
+   the following code snippet would **FAIL**:
+   ::
+
+    clients:
+        jack: &jack_client
+            billing_id: 1234
+        bill: &jack_client
+            billing_id: 5678
+    target_client: *jack_client
```
adrienverge, line 47: All other examples use 2-space indentation: let's be consistent!

yamllint/rules/anchors.py
adrienverge: Yamllint parses files sequencially, so it does not know whether the anchor *is defined later in the file* **or** *is totally missing from the document*.

(Maybe I didn't get your point, in this case please give example of what you mean :))

tests/rules/test_anchors.py
```diff
@@ -0,0 +1,165 @@
...
+NORMAL_ANCHOR_NO_DOC_START = '''---
+key: &keyanchor a
+otherkey: *keyanchor
+'''
+
+DUPLICATED_ANCHOR = '''---
+key1: &keyanchor a
+key2: &keyanchor b
+otherkey: *keyanchor
+'''
```
adrienverge, line 33: Please add a test for anchors duplicated 3 times. Latest version of the PR (commit 542e05f) seems to have this bug!

adrienverge (commented):
@newrushbolt not everything seems to be OK, it looks like the Python linter reports a dozen errors. Can you check it? You can run it locally on your machine with `python -m flake8 .`.
Apart from that, I haven't re-reviewed the PR thoroughly yet since last year, but if all my remarks were handled it should be good to go.


> In #395, a third option for this rule was also discussed (forbid-unused-anchors). It's similar to the duplicated anchors option. Can that option be added into this PR. I feel that will make this whole.

@amimas this development has been open for more than a year, I feel it would be better to just finish and merde it. Then, you can implement this new `forbid-unused-anchors` option if you'd like too.

yamllint/rules/anchors.py
amimas: The error message as written right now specifically says something like this:

> anchor "foo-bar" is used before assignment

But, as you said already

>  it does not know whether the anchor is defined later in the file or is totally missing from the document.

That's why I was suggesting that the error message probably could be less specific because it's possible that the anchor hasn't been defined in the entire file. It could be something like:

> anchor "foo-bar" is used before assignment/defined or does not exist

Hope that makes sense.

amimas: @adrienverge  - Trying to learn the code, as I'm new to this. Would like to contribute if I can.

> Yamllint parses files sequencially, so it does not know whether the anchor is defined later in the file or is totally missing from the document.

Right now problems are reported from each line as tokens are parsed. Is it possible to save them in a global variable instead of reporting right away? Then  if the "end of document token" is reached, report the problems? I didn't see any of the existing rule doing it. Also, can errors from multiple locations be reported? In this example: if an anchor is duplicated 3 times, report the 2nd and 3rd occurrences as problem. Not sure if this is valid:

```python
    # Skipping codes here to keep this snippet short
    yield LintProblem(....)
    yield LintProblem(....)

```

adrienverge: > Hope that makes sense.

Yes it does, thanks, I see what you mean now :pray: 
What I like in the sentence `anchor "foo-bar" is used before assignment` is that it's simpler (it counts) and isn't incompatible with "this anchor will actually never be assigned later in the file".
I see this sentence as a good trade-off between simplicity and understandability (compared to `anchor "foo-bar" is used before assignment/defined or does not exist`)


> Right now problems are reported from each line as tokens are parsed. Is it possible to save them in a global variable instead of reporting right away? Then if the "end of document token" is reached, report the problems? I didn't see any of the existing rule doing it.

That looks like a solution, but yamllint is designed to output errors right away, which allows:
\- having them sorted naturally,
\- being performant and avoid consuming too much RAM,
\- not losing past errors if the script crashes at some point.
Changing this would need a major refactor, and I'm not sure this potentially adaptative error message (`... used before assignment` / `... never assigned`) is worth it.

amimas: I understand what you're saying and it makes sense. I was just asking because I wanted to look at the `forbid-unused-anchors` option. Without checking the entire file, not sure how to verify that an anchor isn't being used. Let me know if you have any suggestions.

Maybe we should discuss it in the feature request issue.

adrienverge: See https://github.com/adrienverge/yamllint/issues/395#issuecomment-1423044402

amimas:
@newrushbolt - are you able to look into the remaining feedbacks?

adrienverge:
@newrushbolt we need feedback from you, because @amimas is implementing a new option for your new rule in https://github.com/adrienverge/yamllint/pull/537.

I know that yamllint requires high-quality contributions, so it's OK if you don't want to spend time on it. Someone else (maybe me) can implement this new rule in a new pull request. Please tell us what you prefer.

adrienverge:
Due to lack of activity here, and the need of @amimas in #537, I've implemented this new rule in #550.

The code is simpler, should be more performant (use of `set()` and save anchors only when needed), the implementation more robust (it handles anchors duplicated 2 or 3+ times), and the doc now includes the default values. I've also renamed `forbid-unknown-aliases` to `forbid-undeclared-aliases`.

@newrushbolt @amimas could you test it and report whether it works for you?

Issue #364, opened 2021-03-04
Missing filenames in Summary->Annotations in Github Actions


dsvua:
I looked at other tickets and all of them mentions that filenames and line numbers suppose to be in Annotation. However, I do not see them there:
![yamllinter_no_filenames_no_lines](https://user-images.githubusercontent.com/9003215/110023260-f6d8be80-7cfa-11eb-956e-d669aca8bf71.png)

dsvua:
Closing as comments are added in `Files Changed` tab:

![github](https://user-images.githubusercontent.com/9003215/110035129-43c39180-7d09-11eb-9577-9471d08461b6.png)

AngrySoundTech:
I think this is still worth looking at, as warnings an errors from rule changes in PR's don't have an easy way to locate because those files were not changed.

Issue #357, opened 2021-01-26
Missing filenames and line numbers in github action output


metbril:
Sorry in advance if this isn't the right place to ask. I have tried both the default yamllint and a custom action with my github workflow. In basic it's working just fine. But what i'm missing in both bases is the filename and line number for the error or warning. When run from the command line (both with/without `--format github`) I can see both, but the github action logs don't seem to include those. 

How can I configure yamllint in the github action to show filenames and line numbers?

<img width="586" alt="Schermafbeelding 2021-01-26 om 22 05 38" src="https://user-images.githubusercontent.com/559346/105905472-c4181800-6022-11eb-8848-0f2d0de5857b.png">

adrienverge:
Hello @metbril, sorry I cannot help here. Maybe @andrewimeson will have a clue? Otherwise, have you tried asking on Stack Overflow or a specialized forum on the internet? In the meantime I'll close this, since this is not a bug or issue.

metbril:
Thanks for the quick reply. I've search with Google and Stackoverflow and this repo specifically, but no one seems to have asked this before.

It looks like when yamllint is run from within a github action, filenames are suppressed OR stripped from the output. My guess is the former? In that case it would be a bug or feature request?

Perhaps this commit caused this:

https://github.com/adrienverge/yamllint/commit/67cb4eb24db7998e67c7ec478c2ac166c35e4a24

adrienverge:
Let's reopen since this looks like an issue. @andrewimeson implemented the feature, he will probably have an answer!

metbril:
I've tried with a forced `format --standard` now and this does show filename and line numbers.

<img width="656" alt="Schermafbeelding 2021-01-26 om 22 14 16" src="https://user-images.githubusercontent.com/559346/105906460-05f58e00-6024-11eb-900e-3cbc5c97541e.png">

metbril:
When run from the command line this is the output:

```shell
$ yamllint --version
yamllint 1.25.0
$ yamllint automations.yaml --format github
::warning file=automations.yaml,line=1,col=1::[document-start] missing document start "---"
::error file=automations.yaml,line=11,col=3::[indentation] wrong indentation: expected 4 but found 2
::error file=automations.yaml,line=12,col=81::[line-length] line too long (82 > 80 characters)
::error file=automations.yaml,line=13,col=81::[line-length] line too long (118 > 80 characters)
[...]

PR #422, opened 2021-11-14
Update github formatting output to utilize groups

### LINKED ISSUE #421: Github format does not display file/line with errors
When utilizing the github format inside of a GitHub Action, the filename, and line number is not displayed in the GitHub Action UI.  This makes it challenging to resolve issues before you have opened an PR to view the annotations on the PR.  It would be great to add this information the the portion that is output in the action logs.

Currently the github format looks like this in raw text:

```
::error file=./myfile.yaml,line=14,col=22::[trailing-spaces] trailing spaces
```

The UI outputs:

```
Error: [trailing-spaces] trailing spaces
```

It would be great to update the output format to something like:

```
::error file=./myfile.yaml,line=14,col=22::./myfile.yaml:14:22: [trailing-spaces] trailing spaces
```

Or to utilize something like [log grouping](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#grouping-log-lines) to provide a better structure similar to the default output format.

strangiato:
Resolves #421

Update the github formatting to utilize groups in the output and provide the line/column number for the error in the output log.

### DIFF
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -93,6 +93,10 @@ class Format(object):
         line += 'line=' + format(problem.line) + ','
         line += 'col=' + format(problem.column)
         line += '::'
+        line += format(problem.line)
+        line += ':'
+        line += format(problem.column)
+        line += ' '
         if problem.rule:
             line += '[' + problem.rule + '] '
         line += problem.desc
@@ -117,6 +121,9 @@ def show_problems(problems, file, args_format, no_warn):
         if args_format == 'parsable':
             print(Format.parsable(problem, file))
         elif args_format == 'github':
+            if first:
+                print('::group::%s' % file)
+                first = False
             print(Format.github(problem, file))
         elif args_format == 'colored':
             if first:
@@ -128,6 +135,9 @@ def show_problems(problems, file, args_format, no_warn):
                 print(file)
                 first = False
             print(Format.standard(problem, file))
+
+    if not first and args_format == 'github':
+        print('::endgroup::')
 
     if not first and args_format != 'parsable':
         print('')
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -553,11 +553,13 @@ class CommandLineTestCase(unittest.TestCase):
         with RunContext(self) as ctx:
             cli.run((path, '--format', 'github'))
         expected_out = (
-            '::error file=%s,line=2,col=4::[trailing-spaces] trailing'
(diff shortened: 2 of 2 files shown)


adrienverge:
Thanks for your contribution @strangiato!

Could you provide screenshots of how it looked before, how it would look after?

@andrewimeson you contributed to the `github` format, do you approve this change?

strangiato:
Let me mock up a repo to grab some screenshots.  Looks like some tests will need to be updated as well.

strangiato:
Example run with Github actions:

https://github.com/strangiato/yamllint-demo/runs/4362632239?check_suite_focus=true

Screen Shot 2021-11-29 at 9 59 26 PM

strangiato:
Tests have been updated and looks like everything is passing.

Let me know if you have any additional requests or feedback!

yamllint/cli.py
```diff
@@ -129,6 +138,9 @@ def show_problems(problems, file, args_format, no_warn):
                 first = False
             print(Format.standard(problem, file))
 
+    if github_group:
```
adrienverge, line 141: Can we get rid of this variable `github_group` and simply use:
```suggestion
    if not first and args_format == 'github':
```
?

adrienverge (commented):
Thanks @strangiato, the demo repo is helpful and the code looks good! (see my comment)

In [the original PR](https://github.com/adrienverge/yamllint/pull/322) that added support for `--format github`, annotations were added like this. How does your change integrates with this?
image

@andrewimeson last chance to say something? The implementation you wrote is about to change.

andrewimeson:
I haven't tested it, but based on the documentation for the groups feature, I think this is a good enhancement and won't be a disruptive change. This will hopefully mitigate the UX issue that has caused multiple issue reports (#377, #364, and #357).

andrewimeson:
As a nitpick, I would probably squash the three commits into one

adrienverge (approved):
> I haven't tested it, but based on the documentation for the groups feature, I think this is a good enhancement and won't be a disruptive change. This will hopefully mitigate the UX issue that has caused multiple issue reports (#377, #364, and #357).

@andrewimeson thanks for your constructive feedback!

> I updated the demo repo and setup a PR to show how it looks here:

It looks perfect.
For future reference, here is a screenshot of how it looks:
Capture d’écran de 2021-12-01 08-49-48

Thanks for your contribution!

PR #429, opened 2021-12-11
fix: allow whitespace after the shebang marker. Closes #428.

### LINKED ISSUE #428: yamllint doesn't ignore shebang lines with whitespace after the shebang marker
Whitespace after the _#!_ marker on shebang lines is authorized and optional, as explained on [Wikipedia's entry for shebang line](https://en.wikipedia.org/wiki/Shebang_(Unix)) as can be seen from the extracts below : 

**_White space after #! is optional_**

and 

**_It has been claimed[20] that some old versions of Unix expect the normal shebang to be followed by a space and a slash (#! /), but this appears to be untrue;[21] rather, blanks after the shebang have traditionally been allowed, and sometimes documented with a space_**

However yamllint fails to ignore such shebang lines as can be seen with the following YAML file

```
#! /usr/bin/env blah
---
name: blah
```

which gives this result with _yamllint_'s default configuration 

```
stdin:1:2: [warning] missing starting space in comment (comments)
```

This issue seems to be related to the presence of '\S' in the regex used to match the shebang lines in _comments.py_

A  simpler condition only veryfying the presence of the _!_ character at the start of the buffer would probably fix this.

I'll try to send a PR.

tamere-allo-peter:
Basically, any character is now allowed after the shebang marker.

### DIFF
--- a/yamllint/rules/comments.py
+++ b/yamllint/rules/comments.py
@@ -74,8 +74,6 @@ Use this rule to control the position and formatting of comments.
 """
 
 
-import re
-
 from yamllint.linter import LintProblem
 
 
--- a/yamllint/rules/comments.py
+++ b/yamllint/rules/comments.py
@@ -105,7 +105,7 @@ def check(conf, comment):
             if (conf['ignore-shebangs'] and
                     comment.line_no == 1 and
                     comment.column_no == 1 and
-                    re.match(r'^!\S', comment.buffer[text_start:])):
+                    comment.buffer[text_start:].startswith('!')):
                 return
             # We can test for both \r and \r\n just by checking first char
             # \r itself is a valid newline on some older OS.
--- a/yamllint/rules/comments.py
+++ b/yamllint/rules/comments.py
@@ -103,7 +103,7 @@ def check(conf, comment):
             if (conf['ignore-shebangs'] and
                     comment.line_no == 1 and
                     comment.column_no == 1 and
-                    comment.buffer[text_start:].startswith('!')):
+                    comment.buffer[text_start] == '!'):
                 return
             # We can test for both \r and \r\n just by checking first char
             # \r itself is a valid newline on some older OS.
--- a/tests/rules/test_comments.py
+++ b/tests/rules/test_comments.py
@@ -97,7 +97,7 @@ class CommentsTestCase(RuleTestCase):
                    '#!/bin/env my-interpreter\n'
                    '', conf,
                    problem1=(1, 2), problem2=(3, 2), problem3=(4, 2))
-        self.check('#! not a shebang\n',
+        self.check('#! is a valid shebang too\n',
                    conf, problem1=(1, 2))
         self.check('key:  #!/not/a/shebang\n',
                    conf, problem1=(1, 8))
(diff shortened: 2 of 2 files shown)


adrienverge:
According to the Wikipedia examples (none show a space) and my experience, spaces after `#!` are rare. Changing current behavior is a breaking change that could unsatisfy many users, for example those who like to enforce the "standard" way of writing shebangs (without an extra space), even though this space is theoretically allowed.

Instead of taking this risk, I wonder if implementing an option `ignore-regex` (proposed at https://github.com/adrienverge/yamllint/issues/361, https://github.com/adrienverge/yamllint/issues/416 and https://github.com/adrienverge/yamllint/issues/428#issuecomment-991512486) wouldn't be more suited for everyone: no breaking change and ability to enforce exactly what each user wants?

I'm curious about the reason for this PR. May I ask what is your use-case?

tamere-allo-peter:
Hello. Still according to Wikipedia, in the _History_ paragraph, there's the complete email from Dennis Ritchie (inventor or co-inventor of the C langage, the UNIX operating system, and the shebang line amongst other things) announcing the shebang functionnality, and the only exemple given in his email contains a space after _#!_,  It even states _Blanks after ! are OK._.

Over my 30 years working with UNIX I've encountered such shebang lines quite often, although they are of course less present in YAML file than in shell scripts.

My use case consists in fixing YAML linting problems reported by yamllint in around 250 Ansible roles and 200 Ansible playbooks, some of the latter containing such shebang lines. That's why I encountered the problem in the first place.

The solution you are proposing is more general, but as I understand it allowing such perfectly fine and valid shebang lines would necessitate changing yamllint's default configuration. IMHO it would be better to allow such shebang lines in the default configuration, because they ARE valid after all, while still alllowing people to enforce the stricter behavior if needed. Not sure how to do this though.

I'll study yamllint's code as time permits and see if I can contribute something better.

Thanks for your time

tamere-allo-peter:
Maybe a good solution would be for the regex used to match shebang lines to come from yamllint's configuration file, the default configuration woud be _^#!_, and people wanting stricter enforcement could set it to _^#!\S_ (the actual one) and finally people wanting special shebang lines could have them be supported as well the same way. So instead of _ignore-regex_ you would instead have _shebang-regex_ with a default value of _^#!_

yamllint/rules/comments.py
```diff
@@ -105,7 +103,7 @@ def check(conf, comment):
             if (conf['ignore-shebangs'] and
                     comment.line_no == 1 and
                     comment.column_no == 1 and
-                    re.match(r'^!\S', comment.buffer[text_start:])):
+                    comment.buffer[text_start:].startswith('!')):
```
adrienverge, line 106: `comment.buffer[text_start:]` creates a new buffer, so this line is probably not very efficient (especially if the YAML file is large).

Moreover we just need the first char so `startswith()` is not that useful. I propose the following:
```suggestion
                    comment.buffer[text_start] == '!'):
```

adrienverge (commented):
Hello Jérôme, sorry about the delay.

Thanks for the pointers and historical context. I have great respect for Dennis
Ritchie (rest in peace), so if he says so, let's do like this ;) More seriously
and more importantly, you are right that spaces *are* valid after all, and
unlike I thought they aren't so rare. I just checked my `/usr/bin` and I got:
\- 492 files starting with `#!/`
\- 55 files starting with `#! /`

Can you just fix a small thing (see my comment)?

yamllint/rules/comments.py
tamere-allo-peter: Done. Thanks for your time.

adrienverge (approved):
Thanks for your contribution to yamllint @tamere-allo-peter!

PR #440, opened 2022-02-03
Add rule to forbid empty values

mpicque:
Today we had an big error due to yamllint **not** considering the following yaml as invalid:
```yaml
key:value
```

That yaml can't be parsed by some yaml parsers because of the white-space missing just after the colon.

As yaml spec states:

> Normally, YAML insists the “:” mapping value indicator be [separated](https://yaml.org/spec/1.2.2/#separation-spaces) from the [value](https://yaml.org/spec/1.2.2/#nodes) by [white space](https://yaml.org/spec/1.2.2/#white-space-characters). A benefit of this restriction is that the “:” character can be used inside [plain scalars](https://yaml.org/spec/1.2.2/#plain-style), as long as it is not followed by [white space](https://yaml.org/spec/1.2.2/#white-space-characters). This allows for unquoted URLs and timestamps. It is also a potential source for confusion as “a:1” is a [plain scalar](https://yaml.org/spec/1.2.2/#plain-style) and not a [key/value pair](https://yaml.org/spec/1.2.2/#mapping).

See here for details: https://yaml.org/spec/1.2.2/#example-flow-mapping-entries

So we introduced a new option `forbid-key-without-values-eof` in the empty-values rule to forbid keys without values at the end of a file.
The default value is set to `false` to avoid to have new lint error magically appearing.
With the example yaml, we obtain the following error:
```
../test.yml
  1:10      error    key with no value found  (empty-values)
```
that option also handle

Other cases are correctly handled by the PyYaml parser.
Example with the following yaml:
```yaml
key1: value1
key2:value2
key3: value3
```
We obtain the following error (output from yamllint, even before adding the new option):
```
../test.yml
  3:1       error    syntax error: could not find expected ':' (syntax)
```

### DIFF
--- a/tests/rules/test_empty_values.py
+++ b/tests/rules/test_empty_values.py
@@ -62,6 +62,30 @@ class EmptyValuesTestCase(RuleTestCase):
                    'implicitly-null:with-colons:in-key2:\n', conf,
                    problem1=(2, 37))
 
+    def test_forbid_key_without_values_eof(self):
+        conf = ('empty-values: {forbid-in-block-mappings: false,\n'
+                '               forbid-in-flow-mappings: false,\n'
+                '               forbid-key-without-values-eof: true}\n')
+        self.check('---\n'
+                   'implicitly-null\n', conf, problem1=(2, 16))
+        self.check('---\n'
+                   'implicitly-null:\n', conf)
+        self.check('---\n'
+                   'explicitly-null: null\n', conf)
+        self.check('---\n'
+                   'implicitly-null:with-colons:in-key\n', conf,
+                   problem1=(2, 35))
+        self.check('---\n'
+                   'implicitly-null:with-colons:in-key:\n', conf)
+        self.check('---\n'
+                   'explicitly-null:with-colons:in-key: null\n', conf)
+        self.check('---\n'
+                   'implicitly-null:with-colons:in-key\n'
+                   'key: value\n', conf, problem1=(3, 4, 'syntax'))
+        self.check('---\n'
+                   'implicitly-null:with-colons:in-key:\n'
+                   'key: value\n', conf)
+
     def test_in_block_mappings_all_lines(self):
         conf = ('empty-values: {forbid-in-block-mappings: true,\n'
                 '               forbid-in-flow-mappings: false}\n')
--- a/yamllint/rules/empty_values.py
+++ b/yamllint/rules/empty_values.py
@@ -22,6 +22,7 @@ Use this rule to prevent nodes with empty content, that implicitly result in
 
 * Use ``forbid-in-block-mappings`` to prevent empty values in block mappings.
 * Use ``forbid-in-flow-mappings`` to prevent empty values in flow mappings.
+* Use ``forbid-key-without-values-eof`` to prevent empty values at the end of a file.
 
 .. rubric:: Default values (when enabled)
 
@@ -31,6 +32,7 @@ Use this rule to prevent nodes with empty content, that implicitly result in
    empty-values:
      forbid-in-block-mappings: true
      forbid-in-flow-mappings: true
+     forbid-key-without-values-eof: false
 
 .. rubric:: Examples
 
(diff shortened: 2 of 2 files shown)


yamllint/rules/empty_values.py
```diff
@@ -73,23 +75,35 @@
...
+
+    key:value
+
 """
 
 import yaml
 
 from yamllint.linter import LintProblem
 
-
```
adrienverge, line 82: Please don't change that

tests/rules/test_empty_values.py
```diff
@@ -62,6 +62,30 @@ def test_in_block_mappings_single_line(self):
...
+        self.check('---\n'
+                   'implicitly-null:with-colons:in-key:\n', conf)
+        self.check('---\n'
+                   'explicitly-null:with-colons:in-key: null\n', conf)
+        self.check('---\n'
+                   'implicitly-null:with-colons:in-key\n'
+                   'key: value\n', conf, problem1=(3, 4, 'syntax'))
+        self.check('---\n'
+                   'implicitly-null:with-colons:in-key:\n'
+                   'key: value\n', conf)
```
adrienverge, line 87: 1. If *true positive* checks are added, you must also add the same *true negative* ones (i.e. the same YAML, but with the new option disabled). Please see how other options do, to see the best place to add them.
2. It would be a good idea to add at least an example of a nested dict:
   ```yaml
   # inside a list, itself inside a surrounding dict:
   a:
     - key1:value1
   ```

adrienverge (commented):
Hello Maxime,

Thanks for contributing, moreover the PR is nicely prepared with docs, tests and all.

I read the commit message with attention, however I'm not sure about the use-case. If I understand correctly, this feature would be a circumvention to report problems only *at the last dict item at the end of the YAML file*?

To see things differently, can you confirm that the problem is that yamllint reports a problem here:
1. ```yaml
   a: 1
   b:2
   c: 3
   ```

but not here:

2. ```yaml
   a: 1
   b:2
   ```

... whereas both examples are invalid YAML? (`yaml.load(file)` fails on both)

If this is the case, then I think the right thing to do is to make yamllint report a problem `syntax` in the 2nd case (like in the 1st one, which is a `syntax` error).

What do you think?

mpicque:
> Hello Maxime,
> 
> Thanks for contributing, moreover the PR is nicely prepared with docs, tests and all.
> 
> I read the commit message with attention, however I'm not sure about the use-case. If I understand correctly, this feature would be a circumvention to report problems only _at the last dict item at the end of the YAML file_?
> 
> To see things differently, can you confirm that the problem is that yamllint reports a problem here:
> 
>     1. ```yaml
>        a: 1
>        b:2
>        c: 3
>        ```
> 
> 
> but not here:
> 
>     2. ```yaml
>        a: 1
>        b:2
>        ```
> 
> 
> ... whereas both examples are invalid YAML? (`yaml.load(file)` fails on both)
> 
> If this is the case, then I think the right thing to do is to make yamllint report a problem `syntax` in the 2nd case (like in the 1st one, which is a `syntax` error).
> 
> What do you think?

It make sense to report the error as a 'syntax' problem, i will look how to change that.

tests/rules/test_empty_values.py
mpicque: ok, i will add thoses checks

yamllint/rules/empty_values.py
mpicque: sorry, maybe caused by autoformat in my IDE...

mpicque:
In fact, after testing further, i have found some cases that can't really be detected, like the example with the dictionary:
```yaml
a:
  - b:1
  - c: 1
```

In that case, a linter can't detect which line is an error between the line 2 and 3.
So i will focus my fix with obvious cases.

adrienverge:
Yes because your example is perfectly valid YAML, it's the simplest way to represent this:
```yaml
a:
  ["b:1", {"c": 1}]
```
which shouldn't report any error because `"b:1"` is a valid string.

What I described in https://github.com/adrienverge/yamllint/pull/440#discussion_r801338513 is "inside a list, itself inside a surrounding dict" (not inside *multiple* lists, like your last example). In other words, you should also add tests for problematic dicts that are *nested*. An example:
```yaml
a:
  - b: 1
    c:2
```

mpicque:
in your last example, the yaml parser already return an error:
`  3:8       error    syntax error: could not find expected ':' (syntax)`

So the only case i need to detect is if the file contain only one scalar without colon and value, all other cases seems to be handled correctly by the yaml parser.

adrienverge:
> So the only case i need to detect is if the file contain only one scalar without colon and value,

... which is valid YAML, right?

Before implementing anything, I suggest making clear in a message what cases you expect to be reported (even if they are already reported by current yamllint) and what cases are valid. Reporting an error for documents that contain a sole scalar string are not likely to be merged.

mpicque:
so i close that pull request as it is most likely to be a bug in the yaml parser we use.
sorry for the inconvenience.

perlpunk:
I've been following this issue, and I think it's not clear yet what you consider invalid.

> Today we had an big error due to yamllint **not** considering the following yaml as invalid:
> 
> ```yaml
> key:value
> ```
> 
> That yaml can't be parsed by some yaml parsers because of the white-space missing just after the colon.

...

> so i close that pull request as it is most likely to be a bug in the yaml parser we use.

The example you gave is valid YAML. If some parser has a problem with it, then this parser has a bug.
It's perfectly fine to use colons in plain scalars if there is no space following.
So the parser *you* use is correct.

PR #446, opened 2022-03-19
Man page lint fix, docs typos

### LINKED ISSUE #76: minor issues in manpage
Hi,
lintian complains about some aspects of the manpage, see here:
https://lintian.debian.org/sources/yamllint/1.24.2-1.html

andrewimeson:
This _maybe_ should be two PRs, but I think the changes are straightforward and small enough to go into one.

Closes #76

The man page looked like:

```man
YAMLLINT(1)                     yamllint                    YAMLLINT(1)




A linter for YAML files.

yamllint does not only check for syntax validity, but for weirdnesses
like key repetition and cosmetic problems such as lines length,
trailing spaces, indentation, etc.
```

but now will have a `NAME` section and look like

```man
YAMLLINT(1)                     yamllint                    YAMLLINT(1)



NAME
       yamllint - Linter for YAML files

       A linter for YAML files.

       yamllint does not only check for syntax validity, but for
       weirdnesses like key repetition and cosmetic problems such as
       lines length, trailing spaces, indentation, etc.
```

I don't like that it repeats `[A] [l]inter for YAML files`, but I couldn't figure out a clean way to exclude the second occurrence, and we need the first one for `apropos` to work. Sphinx is getting the second one from `yamllint/__init__.py`. I tried to use `.. only:: not man`, but it doesn't seem to work in docstrings.


adrienverge:
Very nice, thanks Andrew!

PR #456, opened 2022-04-29
make rule new_lines default to dos on windows systems

Cube707:
As the tile sates, this PR intreduces a check for the operating system and chooses the default value for the new_line rule accordingly.

This is very usefull since this project integrates so nicely with pre-commit, but on windows git will checkout dos newlines, so the precommit can't be used since it isn't protable. This should be fixed now.

I only inteduced the actuall `win32` platform and did not use `cygwin` or `msys`, which are used on windows but intreduce a unix like enviroment with unix like tool that expect unix newlines. But this could be disscuesd

### DIFF
--- a/yamllint/rules/new_lines.py
+++ b/yamllint/rules/new_lines.py
@@ -29,8 +29,17 @@ Use this rule to force the type of new line characters.
  rules:
    new-lines:
      type: unix
+
+**only** on windows (if `sys.platform == "win32"`):
+
+.. code-block:: yaml
+
+ rules:
+   new-lines:
+     type: dos
 """
 
+from sys import platform
 
 from yamllint.linter import LintProblem
 
@@ -38,7 +47,10 @@ from yamllint.linter import LintProblem
 ID = 'new-lines'
 TYPE = 'line'
 CONF = {'type': ('unix', 'dos')}
-DEFAULT = {'type': 'unix'}
+if platform == "win32":
+  DEFAULT = {'type': 'dos'}
+else:
+  DEFAULT = {'type': 'unix'}
 
 
 def check(conf, line):


adrienverge:
Hello, thanks for contributing to yamllint.

I'm not favorable to changing the rules defaults, as this would break linting for many users.

I suggest using a configuration file, either in your project directory, or in `~/.config/yamllint/config`.

Cube707:
I agree that changing default will be challenging, but at the same time the current default makes it quite unintuitive and annoying.

For example, my current use case: I check out a project using yamllint on my Windows machine. Git automagically converts the line endings to dos-crlf for me. I write some code, make a commit, but of course I want to check for any lining errors before pushing. And poof I get a bounch of errors for having the wrong line endings. Now I have to

- a) edit the `.yamllint` file manually. Every time. While being super careful not to commit that change
- b) change all file endings manually. Every time.
- c) reconfigure git to check out unix-style file endings
- d) ignore the errors and just push, hoping I didn't overlook a real error in between all the false-positives.

I can **NOT**:

- use a global `~/.config/yamllint/config`, as it gets overwritten by the project, specific `.yamllint`
- commit a `.yamllint` file that enforces type: dos for all project users
- put the `.yamllint` on `.gitignore` to have custom configuration for every developer, as this defeats the point of using a unified linter

For this purpose c) seems like the most reasonable. But I find that to be a quite big step just to get linting to work.

Projects should be as portable as possible out of the box and especially since this tool is written in Python and it makes portability easy, I feel this should be addressed.

---

May this could be put back untill the next mayor release? That might be a better time for a somewhat breaking change. 

(I still think this isn't a breaking change as everyone affected by this  will allready be jumping through hoops right now..)

vkottler:
> I'm not favorable to changing the rules defaults, as this would break linting for many users.

@adrienverge wat? This tool doesn't work on Windows at all. [The above comment](https://github.com/adrienverge/yamllint/pull/456#issuecomment-1119438897) precisely describes the problem.

Can we merge this? If you're worried about "breaking users" (???) I suggest improving your test suite. Or maybe we should re-upload a fork of this that someone's willing to maintain.

adrienverge:
@vkottler thanks for your very valuable input.

@Cube707 I really believe that making the default conf platform-dependent would be a breaking change. Here are 2 examples:
1. All repos that use yamllint using the default conf (type UNIX) would become configured to "use either UNIX or Windows, depending on where they are run from". If they are run from Windows, new errors won't be reported anymore (without the users knowing why).
2. If 2 persons (one Linux or Mac OS, one Windows) share a Git repo which contains a YAML file with line endings, this file will have yamllint errors in either one of the 2 computers.

Cube707:
## to your first example:

I agree that there will be a few windows side users who reconfigured their workflow to create Unix style line endings in their repos, but I don't think this would lead to errors being not detected. Instead, they will get a lot off errors on there first run telling them about their now wrong line endings. This may bare the risk of them dropping yamllint because of "new random errors", but most likely these are profound users that know to check the repo for news. (They went through the trouble of configuring a non platform standard, they now what they are doing...). They can than either change the `.yamllint` to enforce unix line endings on all platforms, which is more explicit about what they have been doing already, or they can switch to a simpler workflow, just relying on the platform default.

Most windows side user will however have been scared off in the past by yamllint showing theme a bunch of line ending errors seemingly out of the blue. Some will have dropped yamllint and others will have overlooked many errors in between the false positives. This is especially problematic with first time contributors that open faulty PR's, because they thought all yamllint errors on their system where merely windows specific false positives.

## to your second example:

I don't really get what your were pointing to here. 

When using unix and a windows client and gits default settings, with the current version of yamllint, only the windows client would through errors about line endings. After this PR, none of the clients would through errors (which is exactly the point, as these are false positives...)

An if you are talking about the very rare case that different lines of a yaml file are getting checked in with different line endings, so the Unix developer checks in the first line with LF at the end and then the windows client checks in the second line with CRLF at the end. This will be caught by both clients running yamllint, as either one will through an error about a line not matching the default. (even if for this to happen would require someone to hugely screw up there git config)

Cube707:
But I agree, merging this will lead to confusion on quite a few existing projects and need to be documented well.

I would recommend adding a line to the README and maybe pinning an issue about this.

andrewimeson:
What about adding a new option for `new-lines` called `auto` or `platform` that chooses based on the OS running yamllint? This would at least allow projects to start setting that in the repo-level config, but we wouldn't be changing the default.

Cube707:
that is a very good idea actually!

adrienverge:
@Cube707 I'm sorry we didn't understand each other. Let me rephrase my second example.

Let's say you have a repo with only one file `file.yaml`, which contains a certain type of line endings (let's say, DOS line endings `\r\n`). This repo is shared and used by multiple users: some are on Windows, others on Linux.
If `.yamllint` is configured with `new-lines` enabled (which is the default) AND *if the type of required endings is infered from the platform*, then for all users on Linux, the conf will be `new-lines: {type: unix}`, and **yamllint will report errors whereas the repo is pristine and its files are unmodified**. Which is incoherent.

The same conclusion happens for the same example but with `file.yaml` containing UNIX line endings in the commited repo: then all users on Windows will have errors reported on the just-cloned repo, which doesn't make sense because they didn't even touch a file.

Cube707:
Ok, I understand what you mean.

But this is the exact behaviour we have right now for Windows users. A Windows user checking out a file created on linux will get errors about every lineending. Because git changed the lineending on checkout for him and the default config doesn't matche the platform default. 

Also I belive this problem will not come up. Like I said there are very few Windows users that have reconfigurd their workflow to use unix lineending and should their be any projects using dos style lineending on unix mashines, they must already have that in the `.yamlconfig` and will not be affected by this change.

And secondly if a repo wants to enforce a certain type of lineending for all of its useres, regardless of operating system, they can still and should explicitly put that into the `.yamlconfig`

adrienverge:
> But this is the exact problem we have right now for Windows users. A Windows user checking out a file created on linux will get errors about every lineending. Because git changed the lineending on checkout for him and the default config doesn't matche the platform default.

Either I don't understand or I disagree. Git doesn't change line endings on checkout, does it?

> Also I belive this problem will not come up. Like I said there are very few Windows users that have reconfigurd their workflow to use unix lineending and should their be any projects using dos style lineending on unix mashines, they must already have that in the `.yamlconfig` and will not be affected by this change.

I'm not sure about the statement "there are very few Windows users that [...] use unix lineending". Do you have reliable metrics about that?

Also, and again (that's why I tried to explain [in my first example](https://github.com/adrienverge/yamllint/pull/456#issuecomment-1148298472)), there would definitely be problems for projects that want to enforce UNIX line endings, and use the yamllint default conf for that. With your proposed change, UNIX line endings won't be checked anymore on all machines, leading Windows users to push erroneous files without being warned beforehand.

> And secondly if a repo wants to enforce a certain type of lineending for all of its useres, regardless of operating system, they can still and should explicitly put that into the `.yamlconfig`

For sure anyone could rewrite their conf to keep the same behavior. But this is what we call a breaking change.

Cube707:
> By default, Git doesn't change line endings on checkout, does it?

Ah, there is the missunderstanding. Yes it does. This is the default for git on windows!! [here are the docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) and this is what the default installation looks like:

image

Cube707:
> I'm not sure about the statement "there are very few Windows users that [...] use unix lineending". Do you have reliable metrics about that?

I looked around a bit and couldn't find any statistics on that, so this is only from personal experience. I myself have committed incorrect yaml files
to my gitrepos becaus I didn't want to bother to run yamllint because of all the false positives and I've had two discussions on two different repos
where the PR author didn't run the yamllint precommit hook because he was on windows and it spits out a bunch of errors on unchanged lines. I also
have a repo where it took me two days to figure out why pre-commit wouldn't run on my machine until I figured out that it would stop any commit
because of yamllint errors about line endings. I then didn't want to bother setting it up properly as my work on that repo was done. But this is what
let me open this PR..

Maybe @vkottler can also tell what let him here, as it seems to be a similar problem?

I believe that this change would make the project much more beginner-friendly and easier to use. And as this change is only breaking for local development machines while automated checks and pipelines will be fine (they run on unix..) and this local setups are made by experienced developers (as only they would know where the problem comes from and bother to make a workaround). They will quickly be able to pick up on this and adjust their workflow.

I also believe that explicitness should be encouraged, so requiring current projects that rely on this default behaviour to enforce a non platform standard to change to a more explicit approach is actually a good thing in my opinion. Even if it's a little breaking.

But that's only my personal opinion. So maybe the Idea of adding a third option `auto` to the config is the better way to go.

vkottler:
I also make the assumption that the "Checkout Windows-style, commit Unix-style line endings" is canonical for a Windows user, but I can't say if other VCS or cloud storage clients all do the same.

In CI I get around this "problem" by only running `yamllint` on `ubuntu-latest` rather than `windows-latest` (or `macos-latest`), but it does mean that small parts of the local development workflow on Windows have some inconveniences.

_We could consider a configuration option that intends to "infer" the line endings for the platform, and if we want this behavior, we can opt-in to that?_ (+ @Cube707, curious what you think)

@adrienverge I think it makes the most sense for a tool like this to use the current platform's line endings and otherwise require an opt-out config mechanism. Can't back it up without anything more than anecdotal personal experience though.

vkottler:
A [concrete example](https://github.com/vkottler/python-package-template/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/tasks/conf.py#L26) where I have code that _has_ to be platform aware because the code will run on Windows/Unix.

You could JIT your `yamllint` configuration, but the juice isn't really worth the squeeze.

adrienverge:
Thanks for your feedback @Cube707 and @vkottler.

> This is the default for git on windows!!

Beware that your screenshot shows *Git for Windows*, which is not the same thing as the command-line tool *Git*.

> I also make the assumption that the "Checkout Windows-style, commit Unix-style line endings" is canonical for a Windows user

This is the crucial point. I don't know about this, and would definitely appreciate feedback from other users using regular Git (not *Git for Windows* graphical interface).

In any case, changing the default behavior of yamllint when using the `default` conf is not acceptable, for reasons stated above. However, @andrewimeson's proposal to create `new-lines: {type: platform}` seems reasonable.

Cube707:
> Beware that your screenshot shows Git for Windows, which is not the same thing as the command-line tool Git.

Yes it is! This is the normal (and recomendet) way to aquire the the comand-line-tool for windows. It is a bundle that containes the comand-line-tools and other utilityes that are missing on windows and are required to make git work. As windows has no package managment it must be deliverd this way. See:  [Git website Installation instructions](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git#_installing_on_windows),  [Github Setup Tutorial](https://github.com/git-guides/install-git#install-git-on-windows) or [Lenode Turorial for Git](https://www.linode.com/docs/guides/how-to-install-git-on-linux-mac-and-windows/#how-to-install-git-on-windows)

It _additionaly_ containes a GUI, a Bash version for windows and some more things, but that'T not relevant for us.

As for other Windows devopment enviroments that intreduce proper packegemanagment and other unix features (like msys, cygwin or others), like I mentiond above, they are not effected.

Cube707:
> However, @andrewimeson's proposal to create new-lines: {type: platform} seems reasonable.

Than we will settle on this. Should I change the PR or would you prefere to to it yourself? I don't know how complicated this change would be and if I would finde the code segments.

Cube707:
Ok, I will start work on the new PR in the next couple of days 👍


But I would once again like to point out that git for windows is not only the _recomented_ way to get git (the comand-line-tool) onto windows, but the _only_ way other than building it from source (which requires a working compiler toolchain as well as numerous prerequisits I have no Idea about). Or do you now about another way, becaus I would love to know about that?

PR #465, opened 2022-06-06
Float Value Linting

DerekTBrown:
This diff introduces four new lint rules for linting float values:
* Use `require-numeral-before-decimal` to require floats to start with a numeral (ex 0.0 instead of .0)
* Use `forbid-scientific-notation` to forbid scientific notation
* Use `forbid-nan` to forbid NaN values
* Use `forbid-inf` to forbid infinite values


adrienverge:
Hello @DerekTBrown, thanks for contributing!

I think this is in the scope of yamllint, as this lints the form, not the content. And we already have a rule `octal-values`. @perlpunk @andrewimeson what do you think?

I we agree on this, here are some thoughts on my side:
- use full, non-ambiguous words:
  \- `infinity` instead of `inf`, `not a number` instead of `nan`...
  \- `floating-point-values`, `integer-values`, `octal-values`... instead of `float-values` and `int-values`
- check that the new option names (`scientific-notation`...) are consistent with ones used in [the YAML spec](https://yaml.org/spec/1.2.2/#10214-floating-point)
- check that the regexes are consistent with ones from the YAML spec
- check that the reported messages are in the same style as others

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,155 @@
...
+TYPE = 'token'
+CONF = {
+    'require-numeral-before-decimal': bool,
+    'forbid-scientific-notation': bool,
+    'forbid-nan': bool,
+    'forbid-inf': bool,
+}
+DEFAULT = {
+    'require-numeral-before-decimal': False,
+    'forbid-scientific-notation': bool,
```
andrewimeson, line 107: This should be `False` by default too?

andrewimeson (commented):
This seems like a suitable feature to add to the project. There are some flake8 complaints you'll have to clean up, and I agree with the rest of what @adrienverge said.

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 Adrien Vergé
```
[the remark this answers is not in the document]
DerekTBrown: Indeed I do! We should probably use the project name:

https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/

DerekTBrown:
I believe I have made the requisite changes! Thanks so much for the quick review!

adrienverge (commented):
This looks good! Please see my bunch of remarks.

Also, you'll have to add this new rule inside `yamllint/conf/default.yaml` (but disabled of course).

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,162 @@
...
+IS_SCIENTIFIC_NOTATION_PATTERN = re.compile(
+    "[-+]?(\\.[0-9]+|[0-9]+(\\.[0-9]*)?)([eE][-+]?[0-9]+)"
+)
+IS_INF_PATTERN = re.compile("[-+]?(\\.inf|\\.Inf|\\.INF)")
+IS_NAN_PATTERN = re.compile("\\.nan|\\.NaN|\\.NAN")
+
+
+def check(conf, token, prev, next, nextnext, context):
+    if prev and isinstance(prev, yaml.tokens.TagToken):
+        return
```
mattclay, line 125: There's no test for this branch. It should be easy to cover, similarly to what I did in https://github.com/adrienverge/yamllint/pull/461 to cover: https://github.com/adrienverge/yamllint/blob/e49a1011600d8e9831f205e659c389490c2cf7f5/yamllint/rules/octal_values.py#L94-L95

https://github.com/adrienverge/yamllint/blob/82cf1ab222e599e316acb24fefc0ca84ae69229a/tests/rules/test_octal_values.py#L36

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,162 @@
...
+    if not isinstance(token, yaml.tokens.ScalarToken):
+        return
+    if token.style:
+        return
```
mattclay, line 129: There's no test for this branch. A test for a quoted float should suffice.

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,86 @@
...
+            "float-values:\n"
+            "  require-numeral-before-decimal: false\n"
+            "  forbid-scientific-notation: false\n"
+            "  forbid-not-a-number: false\n"
+            "  forbid-infinity: true\n"
+            "new-line-at-end-of-file: disable\n"
+            "document-start: disable\n"
+        )
+        self.check("angle: -.inf", conf, problem=(1, 13))
+        self.check("angle: -.INF", conf, problem=(1, 13))
```
mattclay, line 86: The following tests will fill in the gaps in the test coverage I mentioned:

```suggestion
        self.check("angle: -.INF", conf, problem=(1, 13))
        
    def test_skipped(self):
        conf = (
            "float-values:\n"
            "  require-numeral-before-decimal: true\n"
            "  forbid-scientific-notation: true\n"
            "  forbid-not-a-number: true\n"
            "  forbid-infinity: true\n"
            "new-line-at-end-of-file: disable\n"
            "document-start: disable\n"
        )
        self.check('after-tag: !custom_tag -.inf', conf)
        self.check('has-style: "-.inf"', conf)        
```

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,162 @@
...
+.. rubric:: Default values (when enabled)
+
+.. code-block:: yaml
+
+ rules:
+   float-values:
+        require-numeral-before-decimal: false
+        forbid-scientific-notation: false
+        forbid-not-a-number: false
+        forbid-infinity: false
```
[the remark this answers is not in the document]
DerekTBrown: I am not getting an error from `flake8`.  Is there maybe a configuration I am missing?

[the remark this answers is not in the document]
adrienverge: Your example (and the others following below) use a 2-spaces indentation, but the third sublevel options (starting at `require-numeral-before-decimal`) have an indentation of 7 spaces (instead of 2 + 2 = 4):
```yaml
rules:
  float-values:
       require-numeral-before-decimal: false
       forbid-scientific-notation: false
       forbid-nan: false
       forbid-inf: false
```
Hence, this YAML content would report errors, if linted by yamllint.

DerekTBrown:
I think this should be set for another review!

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,158 @@
...
+Use this rule to limit the permitted values for floating-point numbers.
+YAML permits three classes of float expressions: approximation to real numbers,
+positive and negative infinity and "not a number".
+
+.. rubric:: Options
+
+* Use ``require-numeral-before-decimal'`` to require floats to start
+    with a numeral (ex ``0.0`` instead of ``.0``).
+* Use ``forbid-scientific-notation`` to forbid scientific notation.
+* Use ``forbid-nan`` to forbid NaN values.
```
adrienverge, line 26: Not everybody knows what NaN means, and it's easy to be explicit in this line, so I propose:

```suggestion
* Use ``forbid-nan`` to forbid NaN (not a number) values.
```

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,158 @@
...
+        return
+    if token.style:
+        return
+    val = token.value
+
+    if conf["forbid-nan"] and IS_NAN_PATTERN.match(val):
+        yield LintProblem(
+            token.start_mark.line + 1,
+            token.end_mark.column + 1,
+            'forbidden non a number value "%s"' % token.value,
```
adrienverge, line 132: `non` → `not`:
```suggestion
            'forbidden not a number value "%s"' % token.value,
```

yamllint/conf/default.yaml
```diff
@@ -19,6 +19,11 @@ rules:
     level: warning
   empty-lines: enable
   empty-values: disable
+  float-values:
+    require-numeral-before-decimal: false
+    forbid-scientific-notation: false
+    forbid-nan: false
+    forbid-inf: false
```
adrienverge, line 26: Since it's a new rule I prefer it to be completely disabled by default. It enhances performance and reduces risks of bugs.
```suggestion
  float-values: disable
```

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,158 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Use this rule to limit the permitted values for floating-point numbers.
+YAML permits three classes of float expressions: approximation to real numbers,
+positive and negative infinity and "not a number".
+
+.. rubric:: Options
+
+* Use ``require-numeral-before-decimal'`` to require floats to start
```
adrienverge, line 23: Extra `'` here, to be removed.

adrienverge (commented):
I think you missed some of my comments, please check everything before reposting:
\- It seems like the incorrect indentation in docs is not fixed: https://github.com/adrienverge/yamllint/pull/465#discussion_r894199039
\- There is still an indentation problem on line 24 (previously line 28 in my initial comment).

Otherwise this looks good.

Good idea to include tests with custom tags, like @mattclay proposed.
Can you also add tests for anchors and aliases, e.g.:

```python
self.check('- &anchor .1\n'
           '- *anchor\n',
           conf, problem=(1, 11))
```

By the way I suggest using only one `self.check()` per test function and multiple `problem1=(), problem2=()...`, it could be more readable.

Thank you!

DerekTBrown:
Hey @adrienverge! I think I have fixed everything (sorry, there were a lot of details).

DerekTBrown:
> By the way I suggest using only one self.check() per test function and multiple problem1=(), problem2=()..., it could be more readable.

Personally, I find multiple `self.check()` calls more clear:
- If a test fails, you get an error pointing to a specific line number, indicating a single test condition. I find it much more difficult to map the line/character ranges back to a specific test case.
- This avoids having to deal with multi-line input strings.
- This is consistent with the [existing test style](https://github.com/adrienverge/yamllint/blob/master/tests/rules/test_octal_values.py#L35-L54)

adrienverge:
Hello,

Doc examples still have 4 indentations instead of 2, some lines aren't correctly indented for RST to be valid (e.g. line 80), the first `.. code-block:: yaml` is not indented at all so it won't render as code.

Please take 5 minutes to really compare your new doc to existing docs, so I don't have to do it myself one more time :pray: :pray:

> This is consistent with the existing test style

It depends on which file, most of them use the other way. But OK with your code!

andrewimeson:
`.DS_Store` files made it in. Don't forget to delete those. I recommend setting up a global gitignore file to ignore `.DS_Store` and other OS-specific or editor specific junk files that get left behind that you don't ever want tracked, no matter the repo.

You can see an example of this setup in [my `~/.gitconfig`](https://github.com/andrewimeson/dotfiles/blob/master/.gitconfig#L1-L3), and [my `~/.gitignore_global`](https://github.com/andrewimeson/dotfiles/blob/master/.gitignore_global).

.gitignore
```diff
@@ -5,3 +5,4 @@ __pycache__
 /yamllint.egg-info
 /build
 /.eggs
+.DS_Store
```
andrewimeson, line 8: I don't think this belongs in the repo-specific gitignore (because it shouldn't have to care about OS and editor specific junk files), but as a matter of practicality, I see no harm.

As always, whatever Adrien says.

adrienverge: As you say, this `.DS_Store` has nothing to do with yamllint nor Python, so it should be excluded in a global system-specific conf.

Also, this is completely unrelated from the "float value linting" change.

DerekTBrown: I see no harm in adding it, this is a pretty common annoyance; it is something I see in a lot of `.gitignore`s.

Reverted for now.

adrienverge:
@DerekTBrown before requesting my review, please take the time to review your PR by yourself, to check everything is OK. A first glance on GitHub shows a problem on the very first file:
Capture d’écran du 2022-06-20 18-39-38

Once you've done that, I'll happily re-review your PR.

Thank you!

adrienverge:
It seems you still haven't pre-reviewed the PR yourself:
Capture d’écran du 2022-06-21 17-54-53

> This is not a "problem" but rather a visual indication 

Please don't include changes that don't relate to the feature you propose.

DerekTBrown:
@adrienverge you might need to hit refresh:
Screen Shot 2022-06-22 at 9 02 26 AM

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,128 @@
...
+            "float-values:\n"
+            "  require-numeral-before-decimal: true\n"
+            "  forbid-scientific-notation: true\n"
+            "  forbid-nan: true\n"
+            "  forbid-inf: true\n"
+            "new-line-at-end-of-file: disable\n"
+            "document-start: disable\n"
+        )
+
+        def make_alias_example(value: str) -> str:
```
adrienverge, line 108: We don't use typing yet, which allows the project to work on old Python versions even if they are not officially supported. Can you remove this?

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,128 @@
...
+            return (
+                f"- &angle {value}\n"
+                + "- *angle\n"
```
adrienverge, line 111: I'm not sure the `+` is needed here

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+        # Require leading numeral
+        self.check(make_alias_example("0.0"), conf)
```
adrienverge, line 115: Could you use explicit YAML code here? I understand you want to save some visual space and bytes, but the code will be easier to read and understand, both for newcomers and maintainers, if the full code is visible:
```suggestion
        self.check("- &angle 0.0\n"
                   "- *angle\n",
                   conf)
```

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,128 @@
...
+            "float-values:\n"
+            "  require-numeral-before-decimal: false\n"
+            "  forbid-scientific-notation: true\n"
+            "  forbid-nan: false\n"
+            "  forbid-inf: false\n"
+            "new-line-at-end-of-file: disable\n"
+            "document-start: disable\n"
+        )
+        self.check("angle: 10e6", conf, problem=(1, 12))
+        self.check("angle: 10e-6", conf, problem=(1, 13))
```
adrienverge, line 60: Why is the error at position 12, then at position 13?

These positions don't exist on these lines: their lengths are 11 and 12. So a visual tool that reports problems in an editor wouldn't be able to show the problems.

It would make sense to have it at position 8 (or 10, why not) in both cases. What do you think? Is it possible to fix this?

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,128 @@
...
+            "  forbid-inf: true\n"
+            "new-line-at-end-of-file: disable\n"
+            "document-start: disable\n"
+        )
+        self.check("angle: .inf", conf, problem=(1, 12))
+        self.check("angle: .INF", conf, problem=(1, 12))
+        self.check("angle: -.inf", conf, problem=(1, 13))
+        self.check("angle: -.INF", conf, problem=(1, 13))
+        self.check("angle: \".inf\"", conf)
+        self.check('angle: !custom_tag .inf', conf)
```
adrienverge, line 95: This code is mixing single quotes and double quotes. I propose to use single quotes everywhere (with exceptions if they make sense), for consistency with other files.

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def test_with_aliases(self):
```
adrienverge, line 97: Instead of adding a new function, I would propose to simply append the anchors/aliases after custom tags:
```diff
         self.check("angle: .inf", conf, problem=(1, 12))
         self.check("angle: .INF", conf, problem=(1, 12))
         self.check("angle: -.inf", conf, problem=(1, 13))
         self.check("angle: -.INF", conf, problem=(1, 13))
         self.check("angle: \".inf\"", conf)
         self.check('angle: !custom_tag .inf', conf)
+        self.check('- &angle: .inf\n'
+                   '- *angle\n',
+		   conf, problem=...)
```
What do you think?

adrienverge (commented):
Thanks for your work again. I tried to see everything I could, and hope this is the last batch!

tests/rules/test_float_values.py
DerekTBrown: If this is a stylistic expectation, it might make sense to install a linter to enforce this:

https://github.com/zheller/flake8-quotes/

tests/rules/test_float_values.py
DerekTBrown: > Why is the error at position 12, then at position 13?

I followed the implementation of `octal_values` when implementing `float_values`. The current behavior is to set `LintProblem.column` equal to the last column where the problem occurs (exclusive):

https://github.com/adrienverge/yamllint/blob/ac19d1e427747f91f287d5b6cca4eeac3a232493/yamllint/rules/octal_values.py#L115

`test_octal_values` likewise has the same behavior (column is always 1  + string length):

https://github.com/adrienverge/yamllint/blob/ac19d1e427747f91f287d5b6cca4eeac3a232493/tests/rules/test_octal_values.py#L36

adrienverge: This reveals a bug in `octal-value`, thanks for noticing it!

In any case, this new rule should report the problem where it really is. Said differently, the error should be located here:
```yaml
angle: 10e-6
       ↑
     error
```

not here:
```yaml
angle: 10e-6
            ↑
          error
```

What do you think?

DerekTBrown: Done

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,141 @@
...
+        conf = (
+            'float-values: disable\n'
+            'new-line-at-end-of-file: disable\n'
+            'document-start: disable\n'
+        )
+        self.check('angle: 0.0', conf)
+        self.check('angle: .NaN', conf)
+        self.check('angle: .INF', conf)
+        self.check('angle: .1', conf)
+        self.check('angle: 10e-6', conf)
```
adrienverge, line 32: Custom tags and anchors/aliases are missing here.

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,141 @@
...
+            '  require-numeral-before-decimal: true\n'
+            '  forbid-scientific-notation: false\n'
+            '  forbid-nan: false\n'
+            '  forbid-inf: false\n'
+            'new-line-at-end-of-file: disable\n'
+            'document-start: disable\n'
+        )
+        self.check('angle: .1', conf, problem=(1, 8))
+        self.check('angle: 0.0', conf)
+        self.check('angle: \'0.0\'', conf)
```
adrienverge, line 46: Do you mean 
```suggestion
        self.check('angle: \'.1\'', conf)
```
?
To make sure `.1` is not reported if it's a string?

Otherwise, I don't understand why this line is for, can you explain?

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,141 @@
...
+            '  require-numeral-before-decimal: false\n'
+            '  forbid-scientific-notation: false\n'
+            '  forbid-nan: true\n'
+            '  forbid-inf: false\n'
+            'new-line-at-end-of-file: disable\n'
+            'document-start: disable\n'
+        )
+        self.check('angle: .NaN', conf, problem=(1, 8))
+        self.check('angle: .NAN', conf, problem=(1, 8))
+        self.check('angle: \'NaN\'', conf)
```
adrienverge, line 105: Do you mean 
```suggestion
        self.check('angle: \'.NaN\'', conf)
```
?
To make sure `.NaN` is not reported if it's a string?

Otherwise, I don't understand why this line is for, can you explain?

adrienverge (commented):
I just had a glance on the PR and I still see minor mistakes, see my (new) inline comments.

And I still see many double-quotes in `float_values.py`... Please either change them or answer my comments about them...

@DerekTBrown last time I took the time to re-re-re-review your pull request (again) and ask questions (again), could you please answer those comments, so we don't lose more time? I have a hard time understanding what's on your mind. For example why you changed to backslashes: `'angle: \'10e-6\''` (although I'm OK with it).

tests/rules/test_float_values.py
DerekTBrown: Added

tests/rules/test_float_values.py
DerekTBrown: I am guessing that was my intent... but it was so long ago I have no idea.

tests/rules/test_float_values.py
DerekTBrown: Same with this one, I am guessing this was my intent.

DerekTBrown:
> Could you please answer those comments, so we don't lose more time?

I believe I have addressed all of the comments you have left. If there are any I am missing, please let me know.

tests/rules/test_float_values.py
```diff
@@ -0,0 +1,161 @@
...
+            '  forbid-scientific-notation: false\n'
+            '  forbid-nan: false\n'
+            '  forbid-inf: false\n'
+            'new-line-at-end-of-file: disable\n'
+            'document-start: disable\n'
+        )
+        self.check('angle: .1', conf, problem=(1, 8))
+        self.check('angle: 0.0', conf)
+        self.check('angle: \'.1\'', conf)
+        self.check('angle: !custom_tag 0.0', conf)
```
adrienverge: Here again (like in previous review), the purpose is to test a potential false positive, so it should test `!custom_tag .1` instead.

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,158 @@
...
+        return
+    if token.style:
+        return
+    val = token.value
+
+    if conf['forbid-nan'] and IS_NAN_PATTERN.match(val):
+        yield LintProblem(
+            token.start_mark.line + 1,
+            token.start_mark.column + 1,
+            'forbidden not a number value "%s"' % token.value,
```
adrienverge: This could use the same notation `f'{token.value}'`, like for next lines.

yamllint/rules/float_values.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+
+    if conf['forbid-inf'] and IS_INF_PATTERN.match(val):
+        yield LintProblem(
+            token.start_mark.line + 1,
+            token.start_mark.column + 1,
+            f"forbidden infinite value {token.value}",
```
adrienverge: This line (and the 2 next messages) should have quotes around the value, like in the first message above.

adrienverge (commented):
Thanks for posting again!

I really like your new rule proposal, but I lack time to review and rereview this pull request. So I've taken some time to fix the errors myself, and will push them as separated commits to preserve your authorship. I hope you're fine with this. (You can then ignore the comments I started writing, below this review.)

Thank you again for contributing.

PR #467, opened 2022-06-09
linter: Remove unreachable exception handler

mattclay:
Remove two `try/except UnicodeError` exception handlers which were added in commit c8ba8f7e99ad00ba84c65cd49e2cd288388cc232 for Python 2.x compatibility. Now that Python 2.x is no longer supported, the `except` is unreachable and is no longer needed.


adrienverge (approved):
Thanks!

PR #468, opened 2022-06-09
octal values: pre-compile octal regex

DerekTBrown:
This change introduces a fairly simple performance improvement; it statically compiles the `_is_octal_number_regex` rather than compiling it each time the checker is called.

### DIFF
--- a/yamllint/rules/octal_values.py
+++ b/yamllint/rules/octal_values.py
@@ -84,9 +84,7 @@ CONF = {'forbid-implicit-octal': bool,
 DEFAULT = {'forbid-implicit-octal': True,
            'forbid-explicit-octal': True}
 
-
-def _is_octal_number(string):
-    return re.match(r'^[0-7]+$', string) is not None
+IS_OCTAL_NUMBER_PATTERN = re.compile(r'^[0-7]+$')
 
 
 def check(conf, token, prev, next, nextnext, context):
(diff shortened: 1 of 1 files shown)


adrienverge (commented):
OK!

In this case I propose to simply drop `_is_octal_number()`, because `if IS_OCTAL_NUMBER_PATTERN.match(val)` is self-explicit. What do you think?

Also, have you done tests on how this improves performance? E.g. run `time yamllint dir/` on a large directory with many octal values?
I suppose it's negligeable, since most of yamlllint's execution time comes from IO (reading files), not from using the CPU.

DerekTBrown:
> In this case I propose to simply drop _is_octal_number(), because if IS_OCTAL_NUMBER_PATTERN.match(val) is self-explicit. What do you think?

Sounds good to me!

> Also, have you done tests on how this improves performance? E.g. run time yamllint dir/ on a large directory with many octal values?

In practice, I think the benefit is going to be small: the behavior change only occurs after checking for the first digits being `0` or `0o`. In the abstract, however, this is always going to be faster:

https://replit.com/@DerekTBrown/Regex-Compilation-Perf?v=1

> PS: The commit would better fit our convention if named "octal-values: Pre-compile regex for performance".

Done. People I work with are very into [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) at the moment, hence the `perf:`.

adrienverge (approved):
> https://replit.com/@DerekTBrown/Regex-Compilation-Perf?v=1

Interesting!

PR #469, opened 2022-06-10
linter: Remove redundant conditional

mattclay:
Remove the redundant conditional used when reporting a syntax error at the same location as a cosmetic problem. Also reword the comment explaining the logic to more accurately describe the situation.

This eliminates an unreachable `syntax_error = None` assignment.


adrienverge (approved):
Makes sense.

This repetition is a left-over of commit 222f7a2 in 2016.

PR #474, opened 2022-06-18
rule new-lines: add `type: platform` configuration option

Cube707:
This PR follows up on #456 in which it was concluded that the default for the new-lines rule should not be changed but rather a new option is to be introduced.

## Problem

As explained in [this comment on the parent PR](https://github.com/adrienverge/yamllint/pull/456#issuecomment-1119438897) yamllint currently has portability issues that make it hard to use on Windows clients. These stem from git converting line-endings to the platform default on checkout and yamllint assuming that the line-ending type should be enforced for all development platforms. This leads to windows-clients needing to be heavily reconfigured to make yamllint work there properly.

## PR's content

This PR introduces a new option for the `new-lines` rule: `type: auto`.

The new option infers the correct newline character(s) from the operating system that is currently running yamllint. This allows developers to create yaml-files with their systems default settings and relying on git to convert them.

Projects that wish to enforce a certain type of line-ending are still able to do so by keeping one of the original two options.

### Linting / Testing

`flake8`, `doc8` and `unittest` completet without error on my local machine.

### DIFF
--- a/tests/rules/test_new_lines.py
+++ b/tests/rules/test_new_lines.py
@@ -13,6 +13,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from unittest import mock
+
 from tests.common import RuleTestCase
 
 
@@ -58,3 +60,37 @@ class NewLinesTestCase(RuleTestCase):
         self.check('\r\n', conf)
         self.check('---\ntext\n', conf, problem=(1, 4))
         self.check('---\r\ntext\r\n', conf)
+
+    def test_platform_type(self):
+        conf = ('new-line-at-end-of-file: disable\n'
+                'new-lines: {type: platform}\n')
+
+        self.check('', conf)
+
+        # mock the Linux new-line-character
+        with mock.patch('yamllint.rules.new_lines.linesep', '\n'):
+            self.check('\n', conf)
+            self.check('\r\n', conf, problem=(1, 1))
+            self.check('---\ntext\n', conf)
+            self.check('---\r\ntext\r\n', conf, problem=(1, 4))
+            self.check('---\r\ntext\n', conf, problem=(1, 4))
+            # FIXME: the following tests currently don't work
+            # because only the first line is checked for line-endings
+            # see: issue #475
+            # ---
+            # self.check('---\ntext\r\nfoo\n', conf, problem=(2, 4))
+            # self.check('---\ntext\r\n', conf, problem=(2, 4))
+
+        # mock the Windows new-line-character
+        with mock.patch('yamllint.rules.new_lines.linesep', '\r\n'):
+            self.check('\r\n', conf)
+            self.check('\n', conf, problem=(1, 1))
+            self.check('---\r\ntext\r\n', conf)
+            self.check('---\ntext\n', conf, problem=(1, 4))
+            self.check('---\ntext\r\n', conf, problem=(1, 4))
+            # FIXME: the following tests currently don't work
+            # because only the first line is checked for line-endings
+            # see: issue #475
+            # ---
+            # self.check('---\r\ntext\nfoo\r\n', conf, problem=(2, 4))
+            # self.check('---\r\ntext\n', conf, problem=(2, 4))
--- a/yamllint/rules/new_lines.py
+++ b/yamllint/rules/new_lines.py
@@ -41,13 +41,13 @@ DEFAULT = {'type': 'unix'}
 
 
 def check(conf, line):
+    if conf['type'] == 'unix':
+        newline_char = '\n'
+    elif conf['type'] == 'dos':
+        newline_char = '\r\n'
+
     if line.start == 0 and len(line.buffer) > line.end:
-        if conf['type'] == 'dos':
-            if (line.end + 2 > len(line.buffer) or
-                    line.buffer[line.end:line.end + 2] != '\r\n'):
-                yield LintProblem(1, line.end - line.start + 1,
-                                  'wrong new line character: expected \\r\\n')
-        else:
-            if line.buffer[line.end] != '\n':
-                yield LintProblem(1, line.end - line.start + 1,
-                                  'wrong new line character: expected \\n')
+        if line.buffer[line.end:line.end + len(newline_char)] != newline_char:
+            yield LintProblem(1, line.end - line.start + 1,
+                              'wrong new line character: expected {}'
+                              .format(repr(newline_char).strip('\'')))
--- a/yamllint/rules/new_lines.py
+++ b/yamllint/rules/new_lines.py
@@ -48,6 +48,6 @@ def check(conf, line):
                 yield LintProblem(1, line.end - line.start + 1,
                                   'wrong new line character: expected \\r\\n')
         else:
-            if line.buffer[line.end] == '\r':
+            if line.buffer[line.end] != '\n':
                 yield LintProblem(1, line.end - line.start + 1,
                                   'wrong new line character: expected \\n')
--- a/yamllint/rules/new_lines.py
+++ b/yamllint/rules/new_lines.py
@@ -18,8 +18,11 @@ Use this rule to force the type of new line characters.
 
 .. rubric:: Options
 
-* Set ``type`` to ``unix`` to use UNIX-typed new line characters (``\\n``), or
-  ``dos`` to use DOS-typed new line characters (``\\r\\n``).
+* Set ``type`` to ``unix`` to enforce UNIX-typed new line characters (``\\n``),
+  set ``type`` to ``dos`` to enforce DOS-typed new line characters
+  (``\\r\\n``), or set ``type`` to ``platform`` to infer the type from the
+  system running yamllint (``\\n`` on POSIX / UNIX / Linux / Mac OS systems or
+  ``\\r\\n`` on DOS / Windows systems).
 
 .. rubric:: Default values (when enabled)
 
@@ -30,19 +33,22 @@ Use this rule to force the type of new line characters.
      type: unix
 """
 
+from os import linesep
 
 from yamllint.linter import LintProblem
 
 
 ID = 'new-lines'
 TYPE = 'line'
-CONF = {'type': ('unix', 'dos')}
+CONF = {'type': ('unix', 'dos', 'platform')}
 DEFAULT = {'type': 'unix'}
 
 
 def check(conf, line):
     if conf['type'] == 'unix':
         newline_char = '\n'
+    elif conf['type'] == 'platform':
+        newline_char = linesep
     elif conf['type'] == 'dos':
         newline_char = '\r\n'
 


yamllint/rules/new_lines.py
```diff
@@ -30,13 +34,14 @@
      type: unix
 """
 
+from os import linesep as os_linesep
```
adrienverge, line 37: There is no need to rename the import here, please use:
```suggestion
from os import linesep
```

yamllint/rules/new_lines.py
```diff
@@ -47,7 +52,12 @@ def check(conf, line):
...
-        else:
+        elif conf['type'] == 'unix':
             if line.buffer[line.end] != '\n':
                 yield LintProblem(1, line.end - line.start + 1,
                                   'wrong new line character: expected \\n')
+        elif conf['type'] == 'auto':
+            if line.buffer[line.end:line.end + len(os_linesep)] != os_linesep:
+                yield LintProblem(1, line.end - line.start + 1,
+                                  'wrong new line character: expected {}'
+                                  .format(repr(os_linesep)))
```
adrienverge, line 63: I propose something more performant, that doesn't induce code repetition: don't touch this part, but just add right after `def check(conf, line):`:
```python
if conf['type'] == 'platform':
    conf['type'] = 'dos' if linesep == '\r\n' else 'unix'
```
What do you think?

tests/rules/test_new_lines.py
```diff
@@ -58,3 +60,28 @@ def test_dos_type(self):
...
+            self.check('\n', conf, problem=(1, 1))
+            self.check('---\ntext\n', conf, problem=(1, 4))
+
+        # Windows:
+        if (platform == 'win32'):
+            self.check('\r\n', conf)
+            self.check('---\r\ntext\r\n', conf)
+        else:
+            self.check('\r\n', conf, problem=(1, 1))
+            self.check('---\r\ntext\r\n', conf, problem=(1, 4))
```
adrienverge, line 87: Adding checks that rely on the platform is not reliable nor reproduction-safe, I disagree with adding those.

A much better way would be to mock `os.linesep` to 

```python
with mock.patch('os.linesep', return_value='\n'):
    # pretend you're on Linux
    self.check('\n', conf)
    self.check('---\ntext\n', conf)
    self.check('\r\n', conf, problem=(1, 1))
    self.check('---\r\ntext\r\n', conf, problem=(1, 4))

with mock.patch('os.linesep', return_value='\r\n'):
    # pretend you're on Windows
    self.check('\n', conf, problem=(1, 1))
    self.check('---\ntext\n', conf, problem=(1, 4))
    self.check('\r\n', conf)
    self.check('---\r\ntext\r\n', conf)
```

You may need to create two distinct test functions, for the mock to reset the `os.linesep` value in-between.

tests/rules/test_new_lines.py
```diff
@@ -58,3 +60,28 @@ def test_dos_type(self):
         self.check('\r\n', conf)
         self.check('---\ntext\n', conf, problem=(1, 4))
         self.check('---\r\ntext\r\n', conf)
+
+    def test_auto_type(self):
+        conf = ('new-line-at-end-of-file: disable\n'
+                'new-lines: {type: auto}\n')
+
+        self.check('', conf)
```
adrienverge, line 68: Why this?

tests/rules/test_new_lines.py
```diff
@@ -58,3 +60,28 @@ def test_dos_type(self):
...
+            platform == 'darwin'
+        ):
+            self.check('\n', conf)
+            self.check('---\ntext\n', conf)
+        else:
+            self.check('\n', conf, problem=(1, 1))
+            self.check('---\ntext\n', conf, problem=(1, 4))
+
+        # Windows:
+        if (platform == 'win32'):
```
adrienverge, line 82: No useless parentheses in Python, otherwise linters may complain.

adrienverge (commented):
@Cube707 thanks for this PR!

I still think `type: platform` is better than `type: auto`, because `auto` relates to auto-detection from *the file* in other parts of yamllint. On the other hand, `platform` causes no ambiguity.

Please see the lint results, there are some errors.

You can remove `rule` from the commit title: `rule new-lines: ...`, for consistency with other commits.

About the first commit *explicitly check \n for type: unix*, I don't understand it, and I'm not sure what kind of side-effects it could have. Please check out [bullet 6 of the contributing guidelines](https://github.com/adrienverge/yamllint/blob/master/CONTRIBUTING.rst): anyone should be able to quickly read and understand the reason for a change (in case it induces problems on some platforms, for example...)

yamllint/rules/new_lines.py
Cube707: I wanted to make it clear what this line-seperator is, namely that it's OS dependent. But if your finde with `linespe` I will use that

yamllint/rules/new_lines.py
Cube707: I also thought about this option, but the current implementations would allow for any type of `linesep` (even if there are probably no others...). Also should windows ever decide to switch to `\n` this part would still work. 

Maybe changing the order of the `if`, `elif`, blocks would be the better way to improve performance:

```python
if conf['type'] == 'unix':  # most performant, only 1 if check, probably most used
    ...
elif conf['type'] == 'platform':  # will probably become second most used, 2 if check
    ...
elif conf['type'] == 'dos':  # will probably be used the least, 2 if checks
    ...
```

Checking for the platform first would intreduce two if statements in any case, but I don't know if branch prediction helps here enough to make this the better option so.

I also find it a little more readable to have if, elif blocks, but thats personal preference. Also the code is already duplicated, so one more duplication doesn't hurt to badly in my opinion.

Your call, whatever you prefere.

Cube707: I would recomend this how ever:

```python
if conf['type'] == 'platform':
    conf['type'] = 'dos' if sys.platform == 'win32' else 'unix'
```

tests/rules/test_new_lines.py
adrienverge: It's true that this check *may* have sense in the case of the `new-lines` rule, because `''` is a kind of line. OK to keep it!

tests/rules/test_new_lines.py
Cube707: Took me way too long to fugre this out: 

we can not pass the new-line charater to `return_value`, as this makes `os.linesep` a callable and we need it to be a varaibel. This works however and I'm implementing it:

```python
with mock.patch('os.linesep', '\n'):
```

Cube707: as `new_lines.py` uses `from os import linesep` we acttually need to override that copy of `linesep` by using:

```python
with mock.patch("yamllint.rules.new_lines.linesep", '\n'):
```

alternativly we could use `import os` in `new_lines.py` (and òs.linesep` everywhere in the file)

What would you prefere? I would personally make the test more complex and keep the import small for the package code.

yamllint/rules/new_lines.py
```diff
@@ -47,7 +52,12 @@ def check(conf, line):
...
-            if line.buffer[line.end] == '\r':
+        elif conf['type'] == 'unix':
+            if line.buffer[line.end] != '\n':
                 yield LintProblem(1, line.end - line.start + 1,
                                   'wrong new line character: expected \\n')
+        elif conf['type'] == 'auto':
+            if line.buffer[line.end:line.end + len(os_linesep)] != os_linesep:
+                yield LintProblem(1, line.end - line.start + 1,
+                                  'wrong new line character: expected {}'
+                                  .format(repr(os_linesep)))
```
Cube707, line 63: I just realised that this line result in `wrong new line character: expected '\r\n'` where as the other option don't have the `'`, should I crop it, eithr with `.format(repr(os.linesep)[1:-1]))` or `.format(repr(os.linesep).strip('\'')))` or just keep it?

tests/rules/test_new_lines.py
adrienverge:
```python
with mock.patch("yamllint.rules.new_lines.linesep", '\n'):
```
sounds good :+1:

yamllint/rules/new_lines.py
adrienverge: I would prefer implementing my proposition at https://github.com/adrienverge/yamllint/pull/474#discussion_r901639536, which doesn't involve modifying user messages.

yamllint/rules/new_lines.py
Cube707: I just thought of another alternative that reduces code duplication by a lot:

 ```python
def check(conf, line):
    if conf['type'] == 'unix':
        newline_char = '\n'
    elif conf['type'] == 'platform':
        newline_char = linesep
    elif conf['type'] == 'dos':
        newline_char = '\r\n'

    print(len(line.buffer))
    if line.start == 0 and len(line.buffer) > line.end:
        if line.buffer[line.end:line.end + len(newline_char)] != newline_char:
            yield LintProblem(1, line.end - line.start + 1,
                                'wrong new line character: expected {}'
                                .format(repr(newline_char).strip('\'')))
```

what do you think about that?

---

potential Problem: this looses the line-lenth check on dos systems:

https://github.com/adrienverge/yamllint/blob/632665c3e68b0743a98035e0f173a7ad85e37c4f/yamllint/rules/new_lines.py#L46

I am not sure what exacly this is guarding aganinst? I assume files that end in `\n` might result in an IndexError? or am I overlooking something?

Comment (Cube707) [reply yamllint/rules/new_lines.py]: > I propose:
> ```
> -if line.buffer[line.end:line.end + len(newline_char)] != newline_char:
> +if line.buffer[line.end:] != newline_char:
> ```

this was also my fist idea, but doesn't work because `line.buffer` containes all lines...

File: tests/rules/test_new_lines.py [hunk 7]
```diff
@@ -58,3 +61,27 @@ def test_dos_type(self):
...
+        self.check('', conf)
+
+        # mock the Linux new-line-character
+        with mock.patch("yamllint.rules.new_lines.linesep", '\n'):
+            self.check('\n', conf)
+            self.check('\r\n', conf, problem=(1, 1))
+            self.check('---\ntext\n', conf)
+            self.check('---\r\ntext\r\n', conf, problem=(1, 4))
+            self.check('---\r\ntext\n', conf, problem=(1, 4))
+            self.check('---\ntext\r\n', conf, problem=(2, 4))
```
Cube707, line 78: see below

tests/rules/test_new_lines.py
```diff
@@ -58,3 +61,27 @@ def test_dos_type(self):
...
+
+        # mock the Windows new-line-character
+        with mock.patch("yamllint.rules.new_lines.linesep", '\r\n'):
+            self.check('\r\n', conf)
+            self.check('\n', conf, problem=(1, 1))
+            self.check('---\r\ntext\r\n', conf)
+            self.check('---\ntext\n', conf, problem=(1, 4))
+            self.check('---\ntext\r\n', conf, problem=(1, 4))
+            self.check('---\r\ntext\n', conf, problem=(2, 4))
```
Cube707, line 87: both test return no error when only the end of file doesn't match.

Is that by design? (presumably because the end-of-file rule is handled sepertely?)

yamllint/rules/new_lines.py
adrienverge: Argh...

yamllint/rules/new_lines.py
```diff
@@ -30,24 +33,27 @@
...
-                yield LintProblem(1, line.end - line.start + 1,
-                                  'wrong new line character: expected \\r\\n')
-        else:
-            if line.buffer[line.end] == '\r':
-                yield LintProblem(1, line.end - line.start + 1,
-                                  'wrong new line character: expected \\n')
+        if line.buffer[line.end:line.end + len(newline_char)] != newline_char:
+            yield LintProblem(1, line.end - line.start + 1,
+                                'wrong new line character: expected {}'
+                                .format(repr(newline_char).strip('\'')))
```
Cube707, line 59: The `.strip()` removes the apostrophes `'` from the beginning and end of the `newline_char`. This is consistent with the old messages.

We could lose the `.strip()` (making the code a little less complex) and I personally prefer the message with the apostrophe. This breaks `tests/test_cli.py, line 678` and would require the apostrophes to be added there to, so it's probably not worth it.

Cube707 (commented):
Is the commit history so acceptable? Or should I still remove the `explicitly check \n for type: unix` commit? (or explain it better?)

Are you OK with having the refactoring as a separate commit? I expect this would make it easier to find the source of problems between the new option
and the refactoring. But I can also make it a single commit.

Thanks for spending so much time on this!

tests/rules/test_new_lines.py
```diff
@@ -13,6 +13,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from unittest import mock
+
+import yamllint
```
adrienverge, line 18: Is this import useful? I guess you wrote it for *mock*, but I don't think it's required.

tests/rules/test_new_lines.py
```diff
@@ -58,3 +61,27 @@ def test_dos_type(self):
...
         self.check('---\r\ntext\r\n', conf)
+
+    def test_platform_type(self):
+        conf = ('new-line-at-end-of-file: disable\n'
+                'new-lines: {type: platform}\n')
+
+        self.check('', conf)
+
+        # mock the Linux new-line-character
+        with mock.patch("yamllint.rules.new_lines.linesep", '\n'):
```
adrienverge, line 72: Let's use single quotes for consistency :+1: 
(same below)

yamllint/rules/new_lines.py
adrienverge: I agree with your point of view. If the rule was to be created now, maybe I would use the apostrophes too. But let's avoid breaking changes, in case an automated tool relies on an exact string to be output by yamllint.

tests/rules/test_new_lines.py
adrienverge: This is a good question, and I'm not sure I remember perfectly. But I believe *by design*, the `new-lines` rule was made to check only the first line of a file.

In my opinion, this example `---\r\ntext\n` *has* a bug, and either `new-line-at-end-of-file` or `new-lines` should report a problem. But maybe this isn't in the scope of this PR, since this "bug" already exists? What do you think?

adrienverge (commented):
Thanks @Cube707!

The history looks good but I would prefer:
\- either merging the 2 last commits (refactor + add 'platform'),
\- or applying the refactor before the 'platform' addition,
in order to reduce the diff and the hassle of `git blame`.

tests/rules/test_new_lines.py
Cube707: yes, also saw that while testing the import XD

fixed

tests/rules/test_new_lines.py
Cube707: I also just tested `---\ntext\r\nfoo\n` which also raises no error. And I agree that this should defnetly be caught. So the problem really seems to ba that only the first line is checked.

But as the problem does not come from the actuall rule file (`yamllint/rules/new_lines.py`) but from the calling system, I would consider it out of scope of this PR too.

I will open an issue about this and coment out the tests (with a TODO note), if thats OK with you?

Cube707: I handeld it like this now:

```python
            # FIXME: the following tests currently don't work
            # because only the first line is checked for line-endings
            # see: #
            # ---
            # self.check('---\r\ntext\nfoo\r\n', conf, problem=(2, 4))
            # self.check('---\r\ntext\n', conf, problem=(2, 4))
```

I will add the link to the issue once its opend.

That OK with you?

adrienverge: Good idea :+1:

Cube707:
should be good now.

`unittest` and `flake8` pass on my machine.

Cube707:
found the problem. There was a `\n` in the docstring that was not properly escaped.

`python setup.py build_sphinx` complets now on my machine.

Cube707:
rebased

adrienverge:
> found the problem. There was a \n in the docstring that was not properly escaped. (See compare)

Well seen.

adrienverge (approved):
Thanks for this work Jan! :+1:

Issue #476, opened 2022-06-22
bug: duplicate copyright string


chrillep:
https://github.com/adrienverge/yamllint/blob/89b75b7c05cebadb867b621af3ae7cf394abb884/yamllint/__init__.py#L28
https://github.com/adrienverge/yamllint/blob/632665c3e68b0743a98035e0f173a7ad85e37c4f/docs/conf.py#L23

if you care :) .  the footer in the docs has duplicate copyright string.

`© Copyright Copyright 2016, Adrien Vergé Revision 058fef75.`

<img width="529" alt="Skärmavbild 2022-06-22 kl  11 26 59" src="https://user-images.githubusercontent.com/1267931/174994707-b2e93dbb-3ff6-455e-a089-9b3a6169ae12.png">

https://github.com/sphinx-doc/sphinx/blob/1786e989ca16cac8f9d8a966a481d983d3b23f9c/doc/conf.py#L18
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-copyright

chrillep:
> I have a branch up for this, but I don't want to steal the commit credit from you two, so I'll hold off 😄

you go ahead :)

PR #477, opened 2022-06-24
Remove word 'copyright', bump year

andrewimeson:
Brought up in #476. I ran `make html man` and then inspected the copyright text and it looks good to me. Bumped the year so it shows that stuff still happens here 😁


adrienverge (approved):
Haha thanks @andrewimeson!

PR #480, opened 2022-07-10
docs: more badges from shield.io

Cube707:
I personally find badges to be a very helpfull quick and visual way to get the most importand informations about a project. So I added some more to the project. This is the current result:

image

I chose a layout that shows the generall package informations first and the existing "development" informatins on the second line.

### Discussusion:

additional Badges that I didn't find super relevant but could also be included:

PyPI - Status

PyPI - Downloads

PyPI - Format PyPI - Wheel

I wasn't sure if the LICENCE badge woulde be relevant to you or if you prefere the section at the bottom. Also should the badge fully replace the existing section or should both stay? (thats why its currently a seperate commit so I can easyer revert it)

Should the supported Python-versions bade fully replace the "_Written in Python (compatible with Python 3 only)_" line. It feels redundant to me but I didn't want to remove it. Alternativly there could also be a custom badge somewhat like this:

custom python badge

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,7 @@
+| |GitHub Badge| |PyPi version Badge| |Python version Badge|
+| |CI status Badge| |Coverage Badge| |Docs Badge|
+
+
 yamllint
 ========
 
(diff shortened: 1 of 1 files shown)


adrienverge:
Hello, thanks for this proposal!

I also find these badges pretty, but I don't think any of them brings a valuable information to the README file. I would rather let the reader focus on the main content below in the page.

Cube707:
I personaly find it very helpful to quickly see the current released version number and suppored python versions in a quick and visual way instead of having to read through the text or skip over to the Tags list or to PyPi.

And the other way around if I find a project linked on another side (like PyPi for example) having a visual at the top of the readme that show where its sourcecode is hosted is helpful.

But yes they are mostly eye-candy, so if you don'T want them I will close the PR

Cube707:
I only now realised that pypi doesn't show the full readme as the project description but only a short version read from the packages docstring. So this would make the GitHub badge defnetly useless

Issue #424, opened 2021-11-15
Rule quoted-strings enforce quote-type unless the string contains quotes or escape sequences


bendem:
Hello, `quoted-strings` with `quote-type` is a great rule for uniformity, but there are times when it just doesn't make sense. Once of those times is when we use quotes inside the string. In that case, it doesn't make sense to enforce the quote-type.
```yaml
something: 'use the "force" luke'  # yamllint tells me I should write "use the \"force\" luke"
```

Another case is when `quote-type` is set to `double`, but the string contains escape sequences. In those cases, single quote is better, since the content is not interpreted (so we don't need to double backslashes):
```yaml
regexp: '^(.*\scifs\s.*)vers=[12].0(.*)$'  # yamllint tells me I should write regexp: "^(.*\\scifs\\s.*)vers=[12].0(.*)$"
```

adrienverge:
Hello, interesting issue! I guess you've already looked [`required: only-when-needed`](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.quoted_strings) and it doesn't help here.

Do you have an idea for a new option (or extending an option) so that yamllint can handle that?

mwgamble:
I've just run into this while linting some Cloudformation YAML that contains DNS record configuration. How about `prefer-double` and `prefer-single`, with the caveat that the style that results in the least amount of escaping is best?

mwgamble:
I see there's a partial solution implemented in #487 and merged. Is there any timeline for releasing this change?

bendem:
> I see there's a partial solution implemented in #487 and merged. Is there any timeline for releasing this change?

I totally missed the PR, this is great.

As expected, escape exception is harder, but hey, it's a first step 🎉

adrienverge:
> Is there any timeline for releasing this change?

I just released version 1.28.0 for this :+1: 

I'll close the issue, because https://github.com/adrienverge/yamllint/pull/487 aims to solve this problem, and adding new options like `prefer-single` or `prefer-double` would be redundant and conflict with `allow-quoted-quotes`.

TL;DR: users facing [the problem in the example above ↑](https://github.com/adrienverge/yamllint/issues/424#issue-1053447948) can use:
```yaml
rules:
  quoted-strings:
    quote-type: single
    allow-quoted-quotes: true
```
More details: https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.quoted_strings

mwgamble:
[...]

PR #487, opened 2022-08-05
quoted-strings: add allow-quoted-quotes option

rgeraskin:
This PR implements ~~`skip-quoted-quote`~~ `allow-quoted-quotes` option that allows strings like:
- `'foo"bar'` on `quote-type: double`
- `"foo'bar"` on `quote-type: single`

Closes use-case with quotes inside the string from #424. But still no solution for escape sequences.

### DIFF
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -149,7 +149,7 @@ CONF = {'quote-type': ('any', 'single', 'double'),
         'required': (True, False, 'only-when-needed'),
         'extra-required': [str],
         'extra-allowed': [str],
-        'allow-quoted-quotes': (True, False)}
+        'allow-quoted-quotes': bool}
 DEFAULT = {'quote-type': 'any',
            'required': True,
            'extra-required': [],
@@ -225,7 +225,6 @@ def check(conf, token, prev, next, nextnext, context):
     if (not token.plain) and (token.style == "|" or token.style == ">"):
         return
 
-    # Check value is quoted qoute
     if (conf['allow-quoted-quotes'] is True and (not token.plain)
             and ((token.style == "'" and '"' in token.value) or
                  (token.style == '"' and "'" in token.value))):
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -30,6 +30,8 @@ used.
   ``required: false`` and  ``required: only-when-needed``.
 * ``extra-allowed`` is a list of PCRE regexes to allow quoted string values,
   even if ``required: only-when-needed`` is set.
+* ``skip-quoted-quote`` allows (``true``) using disallowed quotes for strings
+  with allowed quotes inside. Default ``false``.
 
 **Note**: Multi-line strings (with ``|`` or ``>``) will not be checked.
 
@@ -43,6 +45,7 @@ used.
      required: true
      extra-required: []
      extra-allowed: []
+     skip-quoted-quote: false
 
 .. rubric:: Examples
 
@@ -112,6 +115,28 @@ used.
 
     - "localhost"
     - this is a string that needs to be QUOTED
+
+#. With ``quoted-strings: {quote-type: double, required: true,
+   skip-quoted-quote: false}``
+
+   the following code snippet would **PASS**:
+   ::
+
+    foo: "bar\"baz"
+
+   the following code snippet would **FAIL**:
+   ::
+
+    foo: 'bar"baz'
+
+#. With ``quoted-strings: {quote-type: double, required: true,
+   skip-quoted-quote: true}``
+
+   the following code snippet would **PASS**:
+   ::
+
+    foo: 'bar"baz'
+
 """
 
 import re
@@ -125,11 +150,13 @@ TYPE = 'token'
 CONF = {'quote-type': ('any', 'single', 'double'),
         'required': (True, False, 'only-when-needed'),
         'extra-required': [str],
-        'extra-allowed': [str]}
+        'extra-allowed': [str],
+        'skip-quoted-quote': (True, False)}
 DEFAULT = {'quote-type': 'any',
            'required': True,
            'extra-required': [],
-           'extra-allowed': []}
+           'extra-allowed': [],
+           'skip-quoted-quote': False}
 
 
 def VALIDATE(conf):
@@ -139,6 +166,8 @@ def VALIDATE(conf):
         return 'cannot use both "required: true" and "extra-required"'
     if conf['required'] is False and len(conf['extra-allowed']) > 0:
         return 'cannot use both "required: false" and "extra-allowed"'
+    if conf['skip-quoted-quote'] is True and conf['quote-type'] == 'any':
+        return '"skip-quoted-quote" has no effect with "quote-type: any"'
 
 
 DEFAULT_SCALAR_TAG = u'tag:yaml.org,2002:str'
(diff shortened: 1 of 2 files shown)


yamllint/rules/quoted_strings.py
```diff
@@ -139,6 +166,8 @@ def VALIDATE(conf):
         return 'cannot use both "required: true" and "extra-required"'
     if conf['required'] is False and len(conf['extra-allowed']) > 0:
         return 'cannot use both "required: false" and "extra-allowed"'
+    if conf['skip-quoted-quote'] is True and conf['quote-type'] == 'any':
+        return '"skip-quoted-quote" has no effect with "quote-type: any"'
```
adrienverge, line 170: I fear it can be unfriendly to the user to fail on that. Maybe we can just silently ignore `skip-quoted-quote` in that case, and run yamllint normally?

yamllint/rules/quoted_strings.py
```diff
@@ -112,6 +115,28 @@
 
     - "localhost"
     - this is a string that needs to be QUOTED
+
+#. With ``quoted-strings: {quote-type: double, required: true,
+   skip-quoted-quote: false}``
```
adrienverge, line 120: What do you think of this:
```suggestion
#. With ``quoted-strings: {skip-quoted-quote: false}``
```

to make the docs more easily understandable?

yamllint/rules/quoted_strings.py
```diff
@@ -112,6 +115,28 @@
...
+
+    foo: "bar\"baz"
+
+   the following code snippet would **FAIL**:
+   ::
+
+    foo: 'bar"baz'
+
+#. With ``quoted-strings: {quote-type: double, required: true,
+   skip-quoted-quote: true}``
```
adrienverge, line 133:
```suggestion
#. With ``quoted-strings: {skip-quoted-quote: true}``
```

yamllint/rules/quoted_strings.py
```diff
@@ -125,11 +150,13 @@
 CONF = {'quote-type': ('any', 'single', 'double'),
         'required': (True, False, 'only-when-needed'),
         'extra-required': [str],
-        'extra-allowed': [str]}
+        'extra-allowed': [str],
+        'skip-quoted-quote': (True, False)}
```
adrienverge, line 154: Better:
```suggestion
        'skip-quoted-quote': bool}
```

yamllint/rules/quoted_strings.py
```diff
@@ -200,19 +229,29 @@ def check(conf, token, prev, next, nextnext, context):
     if (not token.plain) and (token.style == "|" or token.style == ">"):
         return
 
+    # Check value is quoted qoute
```
adrienverge, line 232: Can you remove this comment? It's not very useful since the following code creates a var named `is_quoted_quote`.

yamllint/rules/quoted_strings.py
```diff
@@ -235,7 +274,8 @@ def check(conf, token, prev, next, nextnext, context):
                     quote_type)
 
         # But when used need to match config
-        elif token.style and not _quote_match(quote_type, token.style):
+        elif token.style and not is_quoted_quote and not _quote_match(
+                quote_type, token.style):
```
adrienverge, line 278: For better performance (because yamllint can take some time to run on large repositories...), I propose to compute "*does this string contain quoted quote?*" only if needed. This means computing this in last resort (after `_quote_match()`).

Something like:
```suggestion
        elif (token.style and not _quote_match(quote_type, token.style) and
                not (conf['skip-quoted-quote'] and _has_quoted_quotes(token))):
```

adrienverge (commented):
Hello @rgeraskin, thanks for contributing!

Can you add tests? Make sure the new tests take example from existing ones and insert nicely, for a full coverage and easier code maintenance.

About the new option name: I propose `allow-quoted-quotes` instead of `skip-quoted-quote`, to use conventional "allow"/"forbid" of yamllint options, and the plural form.

> Closes use-case with quotes inside the string from #424. But still no solution for escape sequences.

What do you mean by "no solution for escape sequences"?

yamllint/rules/quoted_strings.py
rgeraskin: I agree that now it looks a bit overloaded with pre-settings. We could easily drop `required: true` because it wouldn't affect test case logic. But without `quote-type: double` it won't complain about quotes at all.

yamllint/rules/quoted_strings.py
rgeraskin: I extracted `is_quoted_quote` variable logic to `_has_quoted_quotes` function and added `conf['allow-quoted-quotes']` check in `if` clauses. Also, I rearranged `if` statements and placed quoted quotes things to the end.

yamllint/rules/quoted_strings.py
rgeraskin: done

yamllint/rules/quoted_strings.py
rgeraskin: You are right, thanks! Done

yamllint/rules/quoted_strings.py
rgeraskin: Sounds reasonable. I've removed this check.

rgeraskin:
Hello @adrienverge, thanks for the detailed and fast feedback!

I've added tests to cover all cases for this feature. Also, I agree with you about the feature name, I've renamed it to `allow-quoted-quotes`.

About "no solution for escape sequences". As stated in #424:
> Another case is when quote-type is set to double, but the string contains escape sequences. In those cases, single quote is better, since the content is not interpreted (so we don't need to double backslashes):

So to achieve this functionality yamllint should be able to determine escape sequences in a string and not to complain about wrong quotes even through `quote-type` setting. It's not so easy as the feature from this PR.

yamllint/rules/quoted_strings.py
rgeraskin: I've dropped `required: true` for now.

adrienverge: Right!

yamllint/rules/quoted_strings.py
adrienverge: Perfect.

adrienverge:
Thanks for these modifications. The code in `yamllint/` looks perfect.

However for tests I would really prefer explicit YAML blocks, even if it causes a lot of repetition. It allows easy and reliable visual check of what we expect for correct/wrong YAML code.

Example:
```python
    def test_allow_quoted_quotes(self):
        conf = 'quoted-strings: {required: true, quote-type: '"', allow-quoted-quotes: true}\n'
        self.check('---\n'
                   'foo1: "[barbaz]"\n'
                   'foo2: "[bar'baz]"\n'
                   conf,
                   problem=...)

        conf = 'quoted-strings: {required: false, quote-type: '"', allow-quoted-quotes: true}\n'
        self.check('---\n'
                   'foo1: "[barbaz]"\n'
                   'foo2: "[bar'baz]"\n'
                   conf,
                   problem=...)

        # quote-type: any should accept both:
                   'foo3: '[bar"baz]'\n'
                   'foo3: "[bar'baz]"\n'
	# etc.
```

yamllint/rules/quoted_strings.py
rgeraskin: I've dropped `required: true` and kept `quote-type: double` because without `quote-type: double` yamllint won't complain about quotes at all.

rgeraskin:
@adrienverge I've updated tests for `allow-quoted-quotes` following your advice. Now they looks similar to other tests.

adrienverge (approved):
Hey Roman, thanks for this cool addition to yamllint, and for your reactivity.

Let's merge this!

adrienverge:
I see that [the generated documentation](https://yamllint.readthedocs.io/en/latest/rules.html#module-yamllint.rules.quoted_strings) is wrong, because the first backslash is removed:
Capture d’écran du 2022-08-07 13-51-52

Do you think you could fix this in a new pull request? (Docs can be generated locally using `python setup.py build_sphinx`.)

rgeraskin:
@adrienverge my bad, I've fixed it in #490 

thank you for review!

bendem:
Thank you for this, this is great!

Issue #485, opened 2022-07-18
With complex nesting, wrong indentation isn't caught for all items


bcbrookman:
In the file below, I'd expect the indentation for all three "foo" maps nested in the sequences under the "items" to be caught by the default "wrong indentation" rules.

```
---
- item1:
  - foo:
      bar: baz
  item2:
  - foo:
      bar: baz
  item3:
  - foo:
      bar: baz

```
However, the first one doesn't seem to be caught; only the subsequent ones are.

```shell
$ yamllint -d default test.yaml 
test.yaml
  6:3       error    wrong indentation: expected 4 but found 2  (indentation)
  9:3       error    wrong indentation: expected 4 but found 2  (indentation)

```

If I properly indent only the last two as shown below, I get no errors.

```
---
- item1:
  - foo:
      bar: baz
  item2:
    - foo:
        bar: baz
  item3:
    - foo:
        bar: baz

```
The provided examples, of course, look unnecessarily complicated, but I uncovered this issue while linting an [Ansible playbook with handlers](https://docs.ansible.com/ansible/latest/user_guide/playbooks_handlers.html#handler-example). I'm unable to change the layout of the file. Regardless, I feel that yamllint should detect this wrong indentation, in general.

I was using yamllint 1.26.3 installed via pip, and also upgraded to 1.27.1 with the same behavior.

I also didn't find any other issues reported that seem to match what I've encountered. Thanks!

adrienverge:
Hello @bcbrookman, first of all thanks for this comprehensive and precise report!

That's a very interesting case, I wasn't aware of this bug. I noticed that it's not about "the first one", it occurs for *all* keys, until a sub-item (`bar: baz`) is inserted:
```yaml
- item1:
  - foo:
  item2:
  - foo:
  item3:
  - foo:
      bar: baz
  item4:
  - foo:   ← the problem is reported here only
```

I guess something needs to be fixed in `yamllint/rules/indentation.py`, and a new test case added. Contributions are welcome!

bcbrookman:
I dug into this and found that it's related to the arbitrary expected indent value of 2 is set when a block sequence should be indented, but the expected indent value is unknown/hasn't been computed yet. I've proposed PR #491 as a potential fix.

erlangparasu:
any updates?

bcbrookman:
Hi @erlangparasu, this was fixed in #491 but the issue was left open. Unless there are objections, I think we can close it at this point (nearly two years later 😆)

[...]

PR #491, opened 2022-08-22
fix: ensure indent-seq works in nested collections

bcbrookman:
Currently, when spaces: consistent, and indent-sequences: true (the defaults), the expected indent for an unindented block sequence is arbitrarily set to 2 expecting it to be at the start of the document. Unfortunately, this causes a bug when the expected indent should be more than 2. For example, in nested collections like the one below, linting would pass, even though the expected indent is _at least_ 3  (since spaces: consistent doesn't enforce how many spaces should be used for indentation). I first reported this behavior in issue #485, which this PR resolves. 

```yaml
---
- item:
  - elem 1
  - elem 2 
```

Instead of setting the expected indent value to 2 arbitrarily, this change sets the value to -1, to explicitly indicate that indentation is required, but the expected indent value is unknown. The problem description that is returned when this wrong indentation is caught then displays the minimum number of spaces expected, since that is known (indent found + 1)

```
3:3       error    wrong indentation: expected at least 3  (indentation)
```

As a side effect, this also fixes a minor related bug I uncovered where the expected indentation printed when linting a nested collection like the one below, is always 2, even though _1 or more_ would also be acceptable with spaces: consistent. 

```yaml
---
item:
- elem 1
- elem 2 
```

### DIFF
--- a/tests/rules/test_indentation.py
+++ b/tests/rules/test_indentation.py
@@ -1370,6 +1370,45 @@ class IndentationTestCase(RuleTestCase):
                    '               key: value\n'
                    '...\n', conf, problem=(2, 2))
 
+    def test_nested_collections_with_spaces_consistent(self):
+        """Tests behavior of {spaces: consistent} in nested collections to
+        ensure wrong-indentation is properly caught--especially when the
+        expected indent value is initially unkown. For details, see
+        https://github.com/adrienverge/yamllint/issues/485.
+        """
+        conf = ('indentation: {spaces: consistent,\n'
+                '              indent-sequences: true}')
+        self.check('---\n'
+                   '- item:\n'
+                   '  - elem\n'
+                   '- item:\n'
+                   '    - elem\n'
+                   '...\n', conf, problem=(3, 3))
+        conf = ('indentation: {spaces: consistent,\n'
+                '              indent-sequences: false}')
+        self.check('---\n'
+                   '- item:\n'
+                   '  - elem\n'
+                   '- item:\n'
+                   '    - elem\n'
+                   '...\n', conf, problem=(5,5))
+        conf = ('indentation: {spaces: consistent,\n'
+                '              indent-sequences: consistent}')
+        self.check('---\n'
+                   '- item:\n'
+                   '  - elem\n'
+                   '- item:\n'
+                   '    - elem\n'
+                   '...\n', conf, problem=(5,5))
+        conf = ('indentation: {spaces: consistent,\n'
+                '              indent-sequences: whatever}')
+        self.check('---\n'
+                   '- item:\n'
+                   '  - elem\n'
+                   '- item:\n'
+                   '    - elem\n'
+                   '...\n', conf)
+
     def test_return(self):
         conf = 'indentation: {spaces: consistent}'
         self.check('---\n'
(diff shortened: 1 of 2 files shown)


tests/rules/test_indentation.py
```diff
@@ -1369,6 +1369,12 @@ def test_nested_collections(self):
                    '     - - - - - very nested: a\n'
                    '               key: value\n'
                    '...\n', conf, problem=(2, 2))
+        conf = 'indentation: {indent-sequences: true}'
+        self.check('---\n'
+                   '- item:\n'
+                   '  - elem 1\n'
+                   '  - elem 2\n'
+                   '...\n', conf, problem=(3, 3))
```
adrienverge, line 1377: Could you move this new test case in a dedicated function, e.g. `test_nested_collections_with_spaces_consistent()`, and add a small link to `https://github.com/adrienverge/yamllint/issues/485` so we easily understand what it's about and avoid touching it too much?

adrienverge (commented):
Hello Brian, thanks for working on this!

The change seems to make sense, although I'm not sure of the implications / side effects it could have. Apart from the 2 things it fixes (#485 + the minor related bug you noticed), have you thought about corner cases that could be impacted by this? Could `-1` be misinterpreted somewhere else in the code? Potential changes of behavior for other rare YAML code? (not every possing YAML is covered by our tests unfortunately)

tests/rules/test_indentation.py
bcbrookman: Sure, I just pushed an amended commit with the test moved to its own function with a reference to the issue along with additional tests to include the other possible indent-sequences values.

bcbrookman:
Hi @adrienverge, thank you for reviewing and so sorry it's taken me this long to respond. 

The only corner case I can think of that this might impact is for potential downstream projects that might always expect the specific `wrong indentation: expected %d but found %d` text. This change would add a new possible output `wrong indentation: expected at least %d'`. I expect parsing the output in this way would be very odd and rare, but I could be mistaken. I also didn't find any other places  in the code where `-1` could be misinterpreted because it's only used to check whether the indent found matches the expected indent. 

Thanks again, and I welcome any additional input you or others might have.

tamere-allo-peter:
As far as parsing yamllint's output is concerned, I really think that yamllint should prefix all different warning and error messages with a unique 4 (for example) digits hex number, e.g. Exxxx for errors and Wxxxx for warnings (like what several other linters do already). This way there would be no difficulty, and also yamllint's output could be translated to other languages without breaking any downstream project.

adrienverge (approved):
@bcbrookman thank you once again! The new commit looks perfect.

About the implication it could have, I agree it would be very odd and rare, so let's take the risk. It should be fairly simple to adapt to the new string, if needed.

@tamere-allo-peter for this purpose, one should use `yamllint -f parsable` which produce an unambiguous output where the line, column, rule name, error level, etc. can be parsed with confidence by any program. In general, such programs are interested by the rule name.
But I believe Brian was mentioning the "human-readable" part of the string (e.g. `wrong indentation: expected at least 2`), which no one should rely on.

tamere-allo-peter:
@adrienverge Yes I know about the parsable format, however this will become ambiguous as soon as i18n comes to yamllint, this is why, IMHO, prefixing each message with a unique ID would be helpful.

PR #493, opened 2022-09-12
ci: Update GitHub Actions

DimitriPapadopoulos:
https://github.com/actions/checkout
https://github.com/actions/setup-python

PR #497, opened 2022-10-07
ci: Automatically update GitHub Actions

DimitriPapadopoulos:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file


DimitriPapadopoulos:
The linter issue seems unrelated. Perhaps an internal CI bug?

DimitriPapadopoulos:
It just updates GitHub actions, not Python dependencies.
Like https://github.com/adrienverge/yamllint/pull/493, but automates the create of the merge request.

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
andrewimeson, line 8: None of the strings in this YAML document need to be quoted 🙂

adrienverge:
I don't see this PR-opening automation as something very useful nor as a time-saver. So unless I miss a big benefit (in which case it should go in the commit message), I'd prefer staying without it and keep the codebase minimal. What do you think?

(By the way I try to stay alert about security updates for PyYAML and Pathspec!)

DimitriPapadopoulos:
Automation in this case should not be seen as a timer-saver, but as a reminder. You will be notified (by an automated pull request) when new versions of GitHub Actions are available.

DimitriPapadopoulos:
[Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)

adrienverge:
I took time to read the documentation (especially the one you just posted https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot), and frankly I don't see the need for such a change (for the above reasons). If you're OK I'd rather keep the codebase minimal and straightforward to dive into.

DimitriPapadopoulos:
I see most other projects in GitHub use Dependabot to automatically notify of updates of dependencies. I guess they are happy with the functionality. The downside might be how complex the CI environment itself is getting – sometimes more intricate than the code itself. However, Dependabot itself is pretty simple. I would recommend adding Dependabot, but certainly can understand that you would like to avoid it in a project with few dependencies.

DimitriPapadopoulos:
For what it's worth, this PR would tick a box in the issues reported by the [Repo-Review](https://learn.scientific-python.org/development/guides/repo-review/?repo=adrienverge%2Fyamllint&branch=master) of [Scientific Python](https://scientific-python.org/):
* [GH200](https://learn.scientific-python.org/development/guides/gha-basic#GH200): Maintained by Dependabot
   All projects should have a .github/dependabot.yml file to support at least GitHub Actions regular updates. Something like this:
    ```yaml
  version: 2
  updates:
  # Maintain dependencies for GitHub Actions
    - package-ecosystem: "github-actions"
      directory: "/"
      schedule:
        interval: "weekly"
  ```

adrienverge:
OK, I note that Scientific Python encourages this.

In my opinion there is no urge to keep GitHub Action files always up-to-date, and I prefer keeping this repo as simple as possible.

PR #499, opened 2022-10-07
docs: Remove Python 2 reference from code comment

DimitriPapadopoulos:


yamllint/linter.py
```diff
@@ -227,7 +227,7 @@ def run(input, conf, filepath=None):
 
     if isinstance(input, (bytes, str)):
         return _run(input, conf, filepath)
-    elif hasattr(input, 'read'):  # Python 2's file or Python 3's io.IOBase
+    elif hasattr(input, 'read'):  # isinstance(input, io.IOBase)
```
adrienverge, line 230: What do you think about:
```suggestion
    elif isinstance(input, io.IOBase)
```
now that yamllint only supports Python 3?

DimitriPapadopoulos: Sure. I just thought that you might want to support optional types with a `read` method, but no `io.IOBase` inheritance. That could make sense in a general purpose library, but probably not here.

adrienverge: `io.IOBase` is supposed to be [the abstract base class for all I/O classes](https://docs.python.org/3/library/io.html#io.IOBase), so these hypothetical *optional types* should theoretically derive from it (I hope they do!)

adrienverge:
One last thing: the commit has changed is no longer about docs. Can you update the title to reflect it? E.g. "linter: Use proper Python 3 I/O type for reading"?

DimitriPapadopoulos:
Should be OK to merge now.

adrienverge (approved):
Perfect, thanks Dimitri!

PR #500, opened 2022-10-07
Security hardening for GitHub Actions

DimitriPapadopoulos:
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions


adrienverge:
Hello!

I've looked at the URL, but it's a long documentation and is not self-explanatory (nor is the code change). I don't get *what* this code change does, nor *why* it's needed?

Ideally commit messages should contain some explanation (especially the *why*), so everyone can understand the context and reasons for each commit (even without access to GitHub, e.g. with `git log`).

DimitriPapadopoulos:
Right, I will explain in the commit. The idea is that your supply chain depends on third party code (GitHub Actions) that can be compromised. Giving minimal permissions (read-only access to the repository) mitigates this risk.

DimitriPapadopoulos:
I think comment is OK now, can you have a look?

adrienverge:
Thanks! Alhough I got the general idea on first review, I have to admit I didn't really understand what was going on in this commit. The commit message is perfectly clear now!

The second link (https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) is quite useful too (it could have gone in the commit message -- another time).

Thank you again for this improvement :+1:

PR #512, opened 2022-10-26
CI: Fix Sphinx-related errors on 'linter' step

adrienverge:
### docs: Fix Sphinx error on non-YAML code snippet

This problem was just introduced by commit cec4f33 "Clarify disable-line
and parser errors, workaround" and produced this error when building
documentation:

    docs/disable_with_comments.rst:120:Could not lex literal_block as
    "yaml". Highlighting skipped.

---

### CI: Fix pip install problem with Pygments version

Recently `python setup.py build_sphinx` started failing with:

    pkg_resources.VersionConflict: (Pygments 2.3.1
    (/usr/lib/python3/dist-packages), Requirement.parse('Pygments>=2.12'))

The reason is that `doc8` 1.0.0 installs `Pygments` 2.3.1, then `Sphinx`
5.3.0 needs `Pygments` ≥ 2.12.

The easiest fix is to change the install order.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,8 +21,8 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
       - run:
-          python -m pip install flake8 flake8-import-order doc8 sphinx
-            rstcheck[sphinx]
+          python -m pip install flake8 flake8-import-order sphinx
+            rstcheck[sphinx] doc8
       - run: python -m pip install .
       - run: flake8 .
       - run: doc8 $(git ls-files '*.rst')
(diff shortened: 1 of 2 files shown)


michael-k:
> The reason is that doc8 1.0.0 installs Pygments 2.3.1, then Sphinx
> 5.3.0 needs Pygments ≥ 2.12.
>
> The easiest fix is to change the install order.

No wonder I wasn't able to easily reproduce that locally.  pip 20.3+ has a dependency resolver that can properly handle those cases without relying on a magic install order: https://pip.pypa.io/en/stable/topics/dependency-resolution/

A `python -m pip install "pip>=20.3"` or `python -m pip install --upgrade pip` before the actual `python -m pip install` should prevent issues like that in the future.

PR #514, opened 2022-10-27
All strings are Unicode in Python 3

DimitriPapadopoulos:
No need for `u'€'`, just use `'€'`.


adrienverge (approved):
Thank you!

I just edited the commit for consistency with the commit title convention and for easier maintenance (especially writing the changelog, but also `git bisect`, etc.)

    Remove Unicode marker before strings

    All strings are Unicode in Python 3. No need for u'€', just use '€'.

PR #515, opened 2022-10-27
Apply some pyupgrade suggestions

DimitriPapadopoulos:
* `io.open()` is an alias for for the builtin `open()` function:
  	https://docs.python.org/3/library/io.html#io.open
* If `open()` cannot open a file, an `OSError` is raised:
  	https://docs.python.org/3/library/functions.html#open
* `EnvironmentError` is kept for compatibility with previous versions; starting from Python 3.3, it is an alias of `OSError`:
  https://docs.python.org/3/library/exceptions.html#EnvironmentError
* Directly yield from an iterable instead of iterating to yield items.

### DIFF
--- a/tests/test_spec_examples.py
+++ b/tests/test_spec_examples.py
@@ -13,7 +13,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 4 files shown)


adrienverge (approved):
Great :+1:

PR #534, opened 2023-01-29
Fix misleading Python API example

okue:
`yamllint.linter.run("example.yaml", yaml_config)` example seems `yamllint.linter.run` opens a given file.
It's misleading.


adrienverge (approved):
Thanks!

PR #544, opened 2023-03-01
docs: Explicitly specify language even when it's plain text

andrewimeson:
rstcheck succeeds with a failure (heh) when there's a code block without a language specified. This can lead to false negatives as the file is no longer being checked by rstcheck.

Error:

    An `AttributeError` error occured. This is most propably due to a
    code block directive (code/code-block/sourcecode) without a
    specified language. This may result in a false negative for source:
    'docs/disable_with_comments.rst'. See
    https://rstcheck-core.readthedocs.io/en/latest/faq/#code-blocks-without-language-sphinx
    for more information.  Success! No issues detected.


adrienverge (approved):
Interesting! Thanks Andrew :+1:

PR #548, opened 2023-03-16
Avoid runtime exception caused by newer pathspec

ssbarnea:
Some changes made in the minor pathspec update recently released break yamllint. This makes the code work with newer version too.

- pathspec==0.11.0 worked
- pathspec==0.11.1 no longer works, fixed by this change

See: https://github.com/ansible/ansible-lint/actions/runs/4439000759/jobs/7790864220?pr=3194

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -44,6 +44,11 @@ class SimpleConfigTestCase(unittest.TestCase):
         with self.assertRaises(config.YamlLintConfigError):
             config.YamlLintConfig('not: valid: yaml')
 
+    def test_is_file_ignored_allows_none(self):
+        c = config.YamlLintConfig('ignore: foobar\n')
+        self.assertTrue(c.is_file_ignored('foobar'))
+        self.assertFalse(c.is_file_ignored(None))
+
     def test_unknown_rule(self):
         with self.assertRaisesRegex(
                 config.YamlLintConfigError,
--- a/yamllint/config.py
+++ b/yamllint/config.py
@@ -45,7 +45,7 @@ class YamlLintConfig:
         self.validate()
 
     def is_file_ignored(self, filepath):
-        return self.ignore and self.ignore.match_file(filepath)
+        return self.ignore and filepath and self.ignore.match_file(filepath)
 
     def is_yaml_file(self, filepath):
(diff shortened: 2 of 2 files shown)


adrienverge:
Hello @ssbarnea, thanks for the fix! Could you give a small example of a directory of YAML files that triggers the bug with pathspec 0.11.1?

ssbarnea:
I will try to add a test for it. I was hoping you will forget to ask me about a test ;)

ssbarnea:
@adrienverge I added the test, it is needed, if you remove the code change the test will fail with exception. It was a good call to ask me to add a test.

adrienverge:
I'm sorry to ask again, but please, could you provide a real-life example of a directory structure + yamllint conf that causes a failure? The added test does't explain it, not why `YamlLintConfig.is_file_ignored()` is called with argument `None` (which could be the real problem -- maybe #549 is a better fix).

ssbarnea:
That happens for anyone trying to call `run_yamllint` without passing the last optional argument, which defaults to None.

```
from yamllint.linter import run as run_yamllint
from yamllint.config import YamlLintConfig

run_yamllint("", YamlLintConfig("ignore: foo.txt"))
```

Sorry for having to mention it but, lack of typing support in yamllint combined with lack of testing for invalid arguments, can cause bugs like this.

The linter is expected to allow linting content that does not come from a specific file (stdin), but once you have a config file that has an `ignore:` section, it chokes when it ends having to check if the file without a name is excluded.

That is the kind of bug that can be prevented when using typing/mypy.

I hope this helps explaining it.

adrienverge:
OK, so from my understanding this is not a bug with command-line tool `yamllint`. It's a bug that only happens when using the Python lib `yamllint`. Right?

The current version of this pull request still doesn't seem appropriate to me, checking that `is_file_ignored()` doesn't ignore a file `None` doesn't make sense. The right thing to do is make sure `is_file_ignored()` is only called with strings (file paths). I'll merge #549 for this purpose.

alanbchristie:
If yesterday's linter fix https://github.com/adrienverge/yamllint/commit/15f820442788d6015793eddb8baced7f061fd19b is a solution (it looks like it is) then can we expect a new PyPI release soon?

PR #549, opened 2023-03-16
linter: Prevent testing is_file_ignored() with filepath == None

adrienverge:
As reported in https://github.com/adrienverge/yamllint/pull/548, there might be a problem with pathspec 0.11.1 which does't allow calling `match_file()` with argument `None` anymore.

The `linter.run()` function shouldn't call `YamlLintConfig.is_file_ignored(None)` anyway.


ssbarnea:
@adrienverge And this bugfix does not have a test attached to it ;)

PR #550, opened 2023-03-19
anchors: Add new rule to detect undeclared or duplicated anchors

### LINKED ISSUE #395: yamllint allows referencing a non-existant anchor

adrienverge:
According to the YAML specification [^1]:

- > It is an error for an alias node to use an anchor that does not
  > previously occur in the document.

  The `forbid-undeclared-aliases` option checks that aliases do have a matching anchor declared previously in the document. Since this is required by the YAML spec, this option is enabled by default.

- > The alias refers to the most recent preceding node having the same
  > anchor.

  This means that having a same anchor repeated in a document is allowed. However users could want to avoid this, so the new option `forbid-duplicated-anchors` allows that. It's disabled by default.

- > It is not an error to specify an anchor that is not used by any
  > alias node.

  This means that it's OK to declare anchors but don't have any alias referencing them. However users could want to avoid this, so a new option (e.g. `forbid-unused-anchors`) could be implemented in the future. See https://github.com/adrienverge/yamllint/pull/537.

Fixes #395
Closes #420

[^1]: https://yaml.org/spec/1.2.2/#71-alias-nodes

### DIFF
--- a/tests/rules/test_anchors.py
+++ b/tests/rules/test_anchors.py
@@ -1,0 +1,209 @@
+# Copyright (C) 2023 Adrien Vergé
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from tests.common import RuleTestCase
+
+
+class AnchorsTestCase(RuleTestCase):
+    rule_id = 'anchors'
+
+    def test_disabled(self):
+        conf = 'anchors: disable'
+        self.check('---\n'
+                   '- &b true\n'
+                   '- &i 42\n'
+                   '- &s hello\n'
+                   '- &f_m {k: v}\n'
+                   '- &f_s [1, 2]\n'
+                   '- *b\n'
+                   '- *i\n'
+                   '- *s\n'
+                   '- *f_m\n'
+                   '- *f_s\n'
+                   '---\n'  # redeclare anchors in a new document
+                   '- &b true\n'
+                   '- &i 42\n'
+                   '- &s hello\n'
+                   '- *b\n'
+                   '- *i\n'
+                   '- *s\n'
+                   '---\n'
+                   'block mapping: &b_m\n'
+                   '  key: value\n'
+                   'extended:\n'
+                   '  <<: *b_m\n'
+                   '  foo: bar\n'
+                   '---\n'
+                   '{a: 1, &x b: 2, c: &y 3, *x: 4, e: *y}\n'
+                   '...\n', conf)
+        self.check('---\n'
+                   '- &i 42\n'
+                   '---\n'
+                   '- &b true\n'
+                   '- &b true\n'
+                   '- &b true\n'
+                   '- &s hello\n'
+                   '- *b\n'
+                   '- *i\n'    # declared in a previous document
+                   '- *f_m\n'  # never declared
+                   '- *f_m\n'
+                   '- *f_m\n'
+                   '- *f_s\n'  # declared after
+                   '- &f_s [1, 2]\n'
+                   '---\n'
+                   'block mapping: &b_m\n'
+                   '  key: value\n'
+                   '---\n'
+                   'block mapping 1: &b_m_bis\n'
+                   '  key: value\n'
+                   'block mapping 2: &b_m_bis\n'
+                   '  key: value\n'
+                   'extended:\n'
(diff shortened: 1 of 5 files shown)


tests/rules/test_anchors.py
```diff
@@ -0,0 +1,209 @@
...
+                   '- *s\n'
+                   '---\n'
+                   'block mapping: &b_m\n'
+                   '  key: value\n'
+                   'extended:\n'
+                   '  <<: *b_m\n'
+                   '  foo: bar\n'
+                   '---\n'
+                   '{a: 1, &x b: 2, c: &y 3, *x: 4, e: *y}\n'
+                   '...\n', conf)
```
amimas, line 110: I must be missing something. Is it a valid yaml? I think `*x: 4` is invalid, is it? I tried the [yaml-to-json converting tool](https://onlineyamltools.com/convert-yaml-to-json) and also [this online yamllint](https://www.yamllint.com/). Both tools are unable to process it because of `*x: 4`.

tests/rules/test_anchors.py
```diff
@@ -0,0 +1,209 @@
...
+                   '---\n'
+                   '{a: 1, &x b: 2, c: &x 3, *x: 4, e: *y}\n'
+                   '...\n', conf,
+                   problem1=(9, 3),
+                   problem2=(10, 3),
+                   problem3=(11, 3),
+                   problem4=(12, 3),
+                   problem5=(13, 3),
+                   problem6=(24, 7),
+                   problem7=(27, 36))
```
amimas, line 145: Should the referenced problem lines be commented on for consistency like the other problems?

amimas (commented):
I haven't had a chance to try it out with actual yaml files. I think it looks good so far though. The unit test currently tests following scenarios:

- rule disabled
- rule enabled with option forbid-undeclared-aliases but disabled forbid-duplicated-anchors
- rule enabled with option forbid-duplicated-anchors but disabled forbid-undeclared-aliases

Should there be another test with both options enabled? Just to cover all the scenarios.

Also, like the yaml samples; they cover all the different types of contents for each tests I think.

tests/rules/test_anchors.py
```diff
@@ -0,0 +1,209 @@
...
+                   '- *s\n'
+                   '---\n'
+                   'block mapping: &b_m\n'
+                   '  key: value\n'
+                   'extended:\n'
+                   '  <<: *b_m\n'
+                   '  foo: bar\n'
+                   '---\n'
+                   '{a: 1, &x b: 2, c: &y 3, *x: 4, e: *y}\n'
+                   '...\n', conf)
```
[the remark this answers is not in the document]
adrienverge: I understand. This style is common to all other rules tests, so I guess it's more a general question. In my case, I got used to this notation and can read it easily.

tests/rules/test_anchors.py
adrienverge: That's a good point, but I don't feel the need for that: I only put comments on problem lines that are tricky to understand/maintain. If you look at other tests files, comments are rare, and placed only if needed.

adrienverge (commented):
Thanks for your feedback @amimas :+1:
I would be best if you could try it, because once merged and released, I won't be able spend much more time on it.

> Should there be another test with both options enabled? Just to cover all the scenarios.

I asked myself the same question while writing them, but concluded no, otherwise it would create a lot of repetition especially when you add your third option `forbid-unused-anchors`. And each option should work individually so I think we're OK. What do you think?

adrienverge:
I'll merge this now so it can go in the same release as https://github.com/adrienverge/yamllint/pull/548#issuecomment-1476606732.

tests/rules/test_anchors.py
amimas: Ahh... Got it. Thanks for explaining. I have not used anchors that way yet. Learned something new today.

tests/rules/test_anchors.py
amimas: Sounds good.

amimas:
Thanks for merging it. Apologies for the delay here. Have been busy lately and didn't get a chance to look into it further.

tests/rules/test_anchors.py
perlpunk: `*x: something`
is not valid YAML.
It should be
`*x : something`.
It is required to put a space after the alias.
It's just that libyaml and PyYAML didn't implement this correctly.
The reason is that a colon can be part of an anchor name, but libyaml and PyYAML don't allow this, so they also don't require a space between the alias and the colon.

To be compatible with all YAML processors, add a space.
A linter where the underlying parser does not error on this should probably error or warn on this itself.

adrienverge: @perlpunk thanks for clarifying this :+1:
I just added the missing spaces in https://github.com/adrienverge/yamllint/pull/558.

PR #554, opened 2023-03-22
Updated the version specifier in the document to latest

takanakahiko:
Usage examples in the current documentation do not refer to the latest version, so users may mistakenly refer to older versions.
To prevent this, the usage examples in the documentation now refer to the latest version.


adrienverge:
Hello,

OK why not, but the yamllint version in this example was already updated yesterday. I don't think it needs to be updated for each yamllint release, especially if someone needs to ensure that it keeps working every time.

Please reword your commit message to follow the contributing guidelines (have a look at past commits touching this file).

takanakahiko:
I think the opinion of repository owners should come first.
If you don't need to update the documentation every time, you don't need to update it this time either.
Closing this PR. :pray:

PR #572, opened 2023-05-11
Look for per-project configuration file in parent directories

### LINKED ISSUE #571: Look for per-project configuration file in parent directories
Can we have yamllint look for a configuration file in parent directories if it cannot find one in the the current

It would be great if we could have an ESLint-inspired configuration file location, similar to the ESLint-inspired output :). Just finding the file would be enough, even without the cascading configuration.

The use 

Yamllint well integrated with Vim through ALE shows linting errors when I edit the YAML files in a project (a puppet control repository that heavily uses YAML files).

The situation is that we have a `.yamllint` at the top level of the repository and it works great if I use vim to edit the files while I am at the top level. If I change into a subdirectory and edit a YAML file from there, Vim certainly shows linting errors as yamllint no longer find the `.yamllint` for the project.

chutzimir:
Inspired be ESLint's search, it looks for configuration files in all parent directories up until it reaches the user's home or root.

closes #571

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -15,7 +15,8 @@ If ``-c`` is not provided, yamllint will look for a configuration file in the
 following locations (by order of preference):
 
 - a file named ``.yamllint``, ``.yamllint.yaml``, or ``.yamllint.yml`` in the
-  current working directory
+  current working directory, or a parent directory (the search for this file is
+  terminated at the user's home or filesystem root)
 - a filename referenced by ``$YAMLLINT_CONFIG_FILE``, if set
 - a file named ``$XDG_CONFIG_HOME/yamllint/config`` or
   ``~/.config/yamllint/config``, if present
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -734,3 +734,64 @@ class CommandLineConfigTestCase(unittest.TestCase):
 
                 self.assertEqual((ctx.returncode, ctx.stdout, ctx.stderr),
                                  (0, '', ''))
+
+    def test_parent_config_file(self):
+        workspace = {'a/b/c/d/e/f/g/a.yml': 'hello: world\n'}
+        conf = ('---\n'
+                'extends: relaxed\n')
+
+        for conf_file in ('.yamllint', '.yamllint.yml', '.yamllint.yaml'):
+            with self.subTest(conf_file):
+                with temp_workspace(workspace):
+                    with RunContext(self) as ctx:
+                        os.chdir('a/b/c/d/e/f')
+                        cli.run(('-f', 'parsable', '.'))
+
+                self.assertEqual((ctx.returncode, ctx.stdout, ctx.stderr),
+                                 (0, './g/a.yml:1:1: [warning] missing '
+                                     'document start "---" (document-start)\n',
+                                     ''))
+
+                with temp_workspace({**workspace, **{conf_file: conf}}):
+                    with RunContext(self) as ctx:
+                        os.chdir('a/b/c/d/e/f')
+                        cli.run(('-f', 'parsable', '.'))
+
+                self.assertEqual((ctx.returncode, ctx.stdout, ctx.stderr),
+                                 (0, '', ''))
+
+    def test_multiple_parent_config_file(self):
+        workspace = {'a/b/c/3spaces.yml': 'array:\n'
+                                          '   - item\n',
+                     'a/b/c/4spaces.yml': 'array:\n'
+                                          '    - item\n',
+                     'a/.yamllint': '---\n'
+                                    'extends: relaxed\n'
+                                    'rules:\n'
+                                    '  indentation:\n'
+                                    '    spaces: 4\n',
+                     }
+
+        conf3 = ('---\n'
+                 'extends: relaxed\n'
+                 'rules:\n'
+                 '  indentation:\n'
+                 '    spaces: 3\n')
+
+        with temp_workspace(workspace):
+            with RunContext(self) as ctx:
+                os.chdir('a/b/c')
+                cli.run(('-f', 'parsable', '.'))
+
+        self.assertEqual((ctx.returncode, ctx.stdout, ctx.stderr),
+                         (0, './3spaces.yml:2:4: [warning] wrong indentation: '
+                         'expected 4 but found 3 (indentation)\n', ''))
+
+        with temp_workspace({**workspace, **{'a/b/.yamllint.yml': conf3}}):
+            with RunContext(self) as ctx:
+                os.chdir('a/b/c')
+                cli.run(('-f', 'parsable', '.'))
+
+        self.assertEqual((ctx.returncode, ctx.stdout, ctx.stderr),
+                         (0, './4spaces.yml:2:5: [warning] wrong indentation: '
+                         'expected 3 but found 4 (indentation)\n', ''))
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -141,6 +141,19 @@ def show_problems(problems, file, args_format, no_warn):
     return max_level
 
 
+def find_project_config_filepath(path='.'):
+    for filename in ('.yamllint', '.yamllint.yaml', '.yamllint.yml'):
+        filepath = os.path.join(path, filename)
+        if os.path.isfile(filepath):
+            return filepath
+
+    if os.path.abspath(path) == os.path.abspath(os.path.expanduser('~')):
+        return None
+    if os.path.abspath(path) == os.path.abspath(os.path.join(path, '..')):
+        return None
+    return find_project_config_filepath(path=os.path.join(path, '..'))
+
+
 def run(argv=None):
     parser = argparse.ArgumentParser(prog=APP_NAME,
                                      description=APP_DESCRIPTION)
@@ -185,6 +198,7 @@ def run(argv=None):
     else:
         user_global_config = os.path.expanduser('~/.config/yamllint/config')
 
+    project_config_filepath = find_project_config_filepath()
     try:
         if args.config_data is not None:
             if args.config_data != '' and ':' not in args.config_data:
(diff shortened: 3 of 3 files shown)


docs/configuration.rst
```diff
@@ -15,7 +15,8 @@ If ``-c`` is not provided, yamllint will look for a configuration file in the
 following locations (by order of preference):
 
 - a file named ``.yamllint``, ``.yamllint.yaml``, or ``.yamllint.yml`` in the
-  current working directory
+  current working directory, or a parent directory. The search for this file is
+  terminated at the user's home, or the root.
```
adrienverge, line 19: To keep the sentence grammatically correct, I propose:
```suggestion
  current working directory, or a parent directory (the search for this file is
  terminated at the user's home or filesystem root)
```

tests/test_cli.py
```diff
@@ -734,3 +734,27 @@ def test_config_file(self):
...
+    def test_parent_config_file(self):
+        workspace = { 'a/b/c/d/e/f/g/a.yml': 'hello: world\n' }
+        conf = ('---\n'
+                'extends: relaxed\n')
+
+        for conf_file in ('.yamllint', '.yamllint.yml', '.yamllint.yaml'):
+            with self.subTest(conf_file):
+                with temp_workspace(workspace):
+                    with RunContext(self) as ctx:
+                        os.chdir('a/b/c/d/e/f')
```
adrienverge, line 747: I was surprised that calling `os.chdir()` here doesn't need to be "undone" after each iteration, but apparently it's OK, it's automatically done by `temp_workspace()`: https://github.com/adrienverge/yamllint/blob/019c87d/tests/common.py#L76-L86


```diff
@@ -741,10 +741,13 @@ class CommandLineConfigTestCase(unittest.TestCase):
                 'extends: relaxed\n')
 
         for conf_file in ('.yamllint', '.yamllint.yml', '.yamllint.yaml'):
+            print('starting test in ' + os.getcwd())
             with self.subTest(conf_file):
                 with temp_workspace(workspace):
+                    print('running subtest in ' + os.getcwd())
                     with RunContext(self) as ctx:
                         os.chdir('a/b/c/d/e/f')
+                        print('running context in ' + os.getcwd())
                         cli.run(('-f', 'parsable', '.'))
 
                 self.assertEqual((ctx.returncode, ctx.stdout, ctx.stderr),
```

yamllint/cli.py
```diff
@@ -140,6 +140,18 @@ def show_problems(problems, file, args_format, no_warn):
 
     return max_level
 
+def find_project_config(path='.'):
+
+    for filename in [ '.yamllint', '.yamllint.yaml', '.yamllint.yml' ]:
```
adrienverge, line 145: You can remove the empty line, and also use parentheses rather than brackets, like in the test file.

yamllint/cli.py
```diff
@@ -140,6 +140,18 @@ def show_problems(problems, file, args_format, no_warn):
...
+        file = os.path.join(path, filename)
+        if os.path.isfile(file):
+            return file
+
+    if os.path.abspath(path) == os.path.abspath(os.path.expanduser('~')):
+        return False
+    if os.path.abspath(path) == os.path.abspath('/'):
+        return False
```
adrienverge, line 153: - I suggest returning `None` rather than `False`, this is more Pythonic.
- Will the comparison to `os.path.abspath('/')` work on Windows-like systems?

yamllint/cli.py
```diff
@@ -192,12 +204,8 @@ def run(argv=None):
...
         elif args.config_file is not None:
             conf = YamlLintConfig(file=args.config_file)
-        elif os.path.isfile('.yamllint'):
-            conf = YamlLintConfig(file='.yamllint')
-        elif os.path.isfile('.yamllint.yaml'):
-            conf = YamlLintConfig(file='.yamllint.yaml')
-        elif os.path.isfile('.yamllint.yml'):
-            conf = YamlLintConfig(file='.yamllint.yml')
+        elif find_project_config():
+            conf = YamlLintConfig(file=find_project_config())
```
adrienverge, line 208: Ideally it would be best to avoid calling `find_project_config()` twice (but also to avoid calling it before the `if/elif` block if not needed). Maybe with a walrus operator `:=`, added in Python 3.8? (but yamllint still supports 3.7...) Or another way?

adrienverge (commented):
Hello, thanks for this idea and implementation!

Although I'm favorable to the idea, this is a **breaking change** (behavior would start changing if top-level `.yamllint` files would be taken into account). **Is anyone following this repo opposed to this?**

Can you also add a `test_multiple_parent_config_file()` that checks that if multiple config files have conflicting rules options, the closest one takes precedence?

tests/test_cli.py
chutzimir: Indeed, I checked that temp_workdir goes back to the previous working directory, so this should be good as is I hope.

yamllint/cli.py
chutzimir: > Will the comparison to `os.path.abspath('/')` work on Windows-like systems?

It seems that it will work...

```
Python 3.10.8 (tags/v3.10.8:aaaf517, Oct 11 2022, 16:50:30) [MSC v.1933 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.getcwd()
'C:\\Program Files\\Python310'
>>> os.path.abspath(os.path.join(os.getcwd(), '..', '..'))
'C:\\'
>>> os.path.abspath('/')
'C:\\'
>>> os.path.abspath(os.path.join(os.getcwd(), '..', '..')) == os.path.abspath('/')
True
```

yamllint/cli.py
chutzimir: Let me call the function before going through all the loops, and only load the file in that spot. I noticed that the CI runs against Python 3.7 as well, so did not go for the `:=` which would have been convenient.

yamllint/cli.py
chutzimir: I addressed all your comments and tried running the tests on Windows, but unfortunately there are too many errors on Windows.

First try:

```
python -m unittest tests/test_cli.py
E
======================================================================
ERROR: test_cli (unittest.loader._FailedTest)
ImportError: Failed to import test module: test_cli
  File "C:\Program Files\Python310\lib\unittest\loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "C:\Users\georgi.georgiev\work\git\yamllint\tests\test_cli.py", line 17, in <module>
    import fcntl
ModuleNotFoundError: No module named 'fcntl'


Ran 1 test in 0.000s

FAILED (errors=1)
```

After deleting the impossible tests, I am getting many errors mostly due to new line format and path separators...

Small sample:
```
First differing element 0:
1
0

- (1,
+ (0,
-  '.\\a.yml:1:1: [warning] missing document start "---" (document-start)\n'
?    ^^

+  './a.yml:1:1: [warning] missing document start "---" (document-start)\n',
?    ^                                                                     +

-  '.\\a.yml:1:13: [error] wrong new line character: expected \\n (new-lines)\n',
   '')
```

```
First differing element 5:
'C:\\[60 chars]3teckb\\s\\s\\s\\s\\s\\s\\s\\s\\s\\s\\s\\s\\s\\s\\s\\file.yaml'
'C:\\[60 chars]3teckb\\s/s/s/s/s/s/s/s/s/s/s/s/s/s/s/file.yaml'
```

But I think the comparison is good, I even ran yamllint on windows and it doesn't trigger a recursion limit.

chutzimir:
I addressed all of the above comments, so let me mark them as resolved. I just now noticed the CONTRIBUTING.md file, so I also fixed the linting problems per flake8 advice. I squashed the commits too, to keep the git history clean.

As to whether this is a breaking change or not... I understand your sentiment. I just want to note that I found it very counterintuitive to get different behavior when I edit a file depending on which directory I start from.

By the way, this makes me realize that what I really intended was to find configuration starting from the lint destination, and not starting from the current directory. Let me think about it a bit...

yamllint/cli.py
```diff
@@ -141,6 +141,19 @@ def show_problems(problems, file, args_format, no_warn):
     return max_level
 
 
+def find_project_config_filename(path='.'):
```
adrienverge, line 144: I see you rename `find_project_config()` to `find_project_config_filename()`, but since it returns a path I would prefer `find_project_config_path()`, what do you think?

yamllint/cli.py
```diff
@@ -141,6 +141,19 @@ def show_problems(problems, file, args_format, no_warn):
...
+    for filename in ('.yamllint', '.yamllint.yaml', '.yamllint.yml'):
+        file = os.path.join(path, filename)
+        if os.path.isfile(file):
+            return file
+
+    if os.path.abspath(path) == os.path.abspath(os.path.expanduser('~')):
+        return None
+    if os.path.abspath(path) == os.path.abspath('/'):
+        return None
```
adrienverge, line 153: > > Will the comparison to `os.path.abspath('/')` work on Windows-like systems?
> 
> It seems that it will work...
> 
> ```
> Python 3.10.8 (tags/v3.10.8:aaaf517, Oct 11 2022, 16:50:30) [MSC v.1933 64 bit (AMD64)] on win32
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import os
> >>> os.getcwd()
> 'C:\\Program Files\\Python310'
> >>> os.path.abspath(os.path.join(os.getcwd(), '..', '..'))
> 'C:\\'
> >>> os.path.abspath('/')
> 'C:\\'
> >>> os.path.abspath(os.path.join(os.getcwd(), '..', '..')) == os.path.abspath('/')
> True
> ```

OK so it works for `C:\\`. But what if the project is under `D:\\`? I fear a possible infinite loop in some situations.

Maybe a safer implementation would be to check whether `os.path.abspath(path) == os.path.abspath(os.path.join(path, '..'))`, and stop if it's the case?

adrienverge (commented):
Thanks for the updated version @chutzimir, it looks good. Please see my 2 comments.

Since no one seems to be opposed to this feature (which is a **breaking change**), I'm +1 to merge it.

yamllint/cli.py
chutzimir: You're spot on. Corrected, squashed, and pushed.

yamllint/cli.py
chutzimir: Valid point. I agree. Corrected, squashed, and pushed.

adrienverge (approved):
Great! Thanks for contributing to yamllint Georgi :+1:

PR #575, opened 2023-06-05
Added Dockerfile

ZuluPro:
Hi,
I added a simple Dockerfile using python:3.
The idea is to ease the usage of yamllint in CI environment.
For instance a Gitlab stage running only yamllint (mounting)

But I cannot setup the Dockerhub/Action part for you

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM python:3
+ADD . /yamllint
+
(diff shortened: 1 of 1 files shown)


adrienverge:
Thanks, but https://github.com/adrienverge/yamllint/pulls?q=dockerfile

Issue #578, opened 2023-06-15
Add SARIF support


mathroule:
It would be nice to be able to format output as [SARIF](https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html).

cheesymoon:
No native SARIF yet, but until (if) it lands: https://github.com/sarif-kit/sarif-kit converts `yamllint -f parsable` output to SARIF 2.1.0. CLI on PyPI, or container action:

```yaml
- run: pipx run yamllint -f parsable . > yamllint.txt || [ $? -eq 1 ]
- uses: sarif-kit/sarif-kit@v0.2.0
  with:
    tool: yamllint
    input: yamllint.txt
    output: yamllint.sarif
- uses: github/codeql-action/upload-sarif@v4
  with:
    sarif_file: yamllint.sarif
    category: yamllint
```

Parsable is the only format it takes, being the one that keeps line, column, level and rule id on every line. Those all carry over. Rule id becomes the SARIF rule and links to its page in the yamllint docs, which is what GitHub builds the alert title from.

yamllint exits 1 on lint errors. Guard swallows that and nothing else, so bad flag or missing binary still fails the job.

Working example: [sarif-kit/demo](https://github.com/sarif-kit/demo), broken on purpose, its [yamllint workflow](https://github.com/sarif-kit/demo/blob/master/.github/workflows/yamllint.yml) runs the snippet above. Alerts look like this: https://sarif-kit.github.io/sarif-kit/yamllint/

Issue #441, opened 2022-02-03
Create `junitxml` output


QuentinN42:
On gitlab, test suite must output in junitxml format to be displayed on the UI.

I think I can add a parser to do this.

QuentinN42:
Soo ...
It's more complicated than I thought.
The xml format must have a head and a tail, need a bit of refactor.
I've forked the repo, I open an MR for that.

QuentinN42:
cf. https://github.com/adrienverge/yamllint/pull/442

adrienverge:
Hello,

Like in other similar issues: new third-party output formats (junitxml, SonarQube, checkstyle, codeclimate, etc.) are not likely to be accepted here.

What about writing a tiny wrapper to convert yamllint's output (YAML format, or JSON once it's implemented) to your favorite XML format?

QuentinN42:
I've already done that but it's flaky as f*

PR #579, opened 2023-06-16
Add SARIF output

mathroule:
Add SARIF output and fixes https://github.com/adrienverge/yamllint/issues/578.

It required a bit of refactoring for being able to output the new format.


mathroule:
@adrienverge do you think SARIF support can be added to YAML lint?

I saw your comment https://github.com/adrienverge/yamllint/issues/441#issuecomment-1030572101, indeed SARIF can be seen as third-party, but it's intended to be a standard compared to JUnit or other formats.

adrienverge:
Hello Mathieu, please excuse my late reply.

You understood it right in #441 (and other discussions): new output formats are not planned for addition. There would be too many of them. I'm sorry.

mathroule:
That's really unfortunate and I can understand your point for not adding a new format. But I would like to insist on the fact that SARIF is standard, from the doc it says:
> The Static Analysis Results Interchange Format (SARIF) has been [approved](https://www.oasis-open.org/news/announcements/static-analysis-results-interchange-format-sarif-v2-1-0-is-approved-as-an-oasis-s) as an [OASIS](https://www.oasis-open.org/) standard. The information and tools on this web site apply to SARIF Version 2.1.0, the version approved by the OASIS.

Since SARIF is a standard, it is supported by [GitHub](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning) and [SonarQube](https://docs.sonarsource.com/sonarqube/9.9/analyzing-source-code/importing-external-issues/importing-issues-from-sarif-reports) out of the box.

adrienverge:
Hello Mathieu, thanks for insisting. About the pull request itself, I have a few comments.

Is the first commit (*Refactor problems output*) really needed? I'm afraid that it postpones the printing of problems to yamllint's output, which is not optimal: it delays results that could be printed on the fly; also, in case of a crash on the last YAML file, problems of previous files won't be shown. Another approach could be to use instances of formatter classes (see my proposal below), and each problem found could be immediately printed with `print(sarifFormatter.format(file, problems))`. For formatters (like SARIF) that need to print `{"$schema": ...` and `...]}` before/after yamllint's results, it could be done with `print(sarifFormatter.header())` and `print(sarifFormatter.footer())`.

About the second commit (*Move formatters to dedicated module*): given the small size of each formatter code, I'd rather avoid creating a new module `yamllint.formatters`. What about just a new file `formatters.py`, possibly with classes `StandardFormatter`, `GitHubFormatter` etc. that implement `format(file, problems)`?
With function names please keep `..._problems()` (not `..._results()`), for consistency. Otherwise there's an unclear mix of `_results` and `_problem`.
There's not much sense that `format_results()` returns `max_level`: `format_results()` is just about printing text. The error level logic would be better kept outside formatters.

In the third commit (*Add support for SARIF format*), please set the copyright of `sarif.py` to Mathieu Rul :wink:
It looks like the output data has many extra informations. Are there needed? Something minimalist would be better in my opinion. Can you keep only what's really required? For example, `help` and `helpUri` if they are not mandatory, because the documentation URI is subject to change; same for `properties.queryUri` which doesn't look intended for this; same for most empty fields like `properties.tags`; same for `ruleIndex`; etc.
Why changing rule names in the output, e.g. `document-start` to `DocumentStart`? These are not the real rule IDs, and will probably create confusion for users.

Commit *Fix null rule for syntax error*: it looks like a breaking change for the `colored` and `github` formatters, and for Python bindings that would use `LintProblem` object. Moreover this change doesn't look related to SARIF. Can you remove it?

In the tests, I suggest removing `test_run_format_sarif_warning()`. It doesn't add much value (the `test_..._warning()` variant is useful for the `colored` format, but not for others).

Finally please review [contribution guidelines](https://github.com/adrienverge/yamllint/blob/master/CONTRIBUTING.rst) to make sure eveything's OK. In particular the SARIF formatter commit should explain what is SARIF and why it's a nice addition. The 2 last commits are just fixups and should be squashed in their relative commit. Tests should go in the same commit as the SARIF feature itself. I don't see any update to the documentation, but it's fine to me (`github` format doesn't have neither).

Bibz87:
Is there any chance that this Pull Request be implemented eventually? I'd really like to use SARIF when linting my YAML files.

mathroule:
> Is there any chance that this Pull Request be implemented eventually? I'd really like to use SARIF when linting my YAML files.

Sorry I'm closing this pull request for now.

Bibz87:
> > Is there any chance that this Pull Request be implemented eventually? I'd really like to use SARIF when linting my YAML files.
> 
> Sorry I'm closing this pull request for now.

I might be able to find some time to pick this one up and bring it to completion. Do you mind if I start from your changes and apply the suggested changes?

mathroule:
> > > Is there any chance that this Pull Request be implemented eventually? I'd really like to use SARIF when linting my YAML files.
> > 
> > 
> > Sorry I'm closing this pull request for now.
> 
> I might be able to find some time to pick this one up and bring it to completion. Do you mind if I start from your changes and apply the suggested changes?

Sure, feel free to reuse my code.

PR #582, opened 2023-06-24
document-end: Fix spurious "missing document end"

### LINKED ISSUE #560: [BUG] False positive "missing document end" error when directive (%) is present
Consider the following yaml:
```
%YAML 1.2
---
hello: world!
...
```
With `document-end: enable` you get `2:1  error  missing document end "..."  (document-end)`.

That's because the current token (`---`) is `DocumentStartToken` and previous token (`%YAML 1.2`) is `DirectiveToken`, but yamllint [belives](https://github.com/adrienverge/yamllint/blob/b92fc9cb3133673f284ccdef76446673bd81c7a8/yamllint/rules/document_end.py#L106) that accepted previous tokens are [`DocumentEndToken` or `StreamStartToken`](https://github.com/adrienverge/yamllint/blob/b92fc9cb3133673f284ccdef76446673bd81c7a8/yamllint/rules/document_end.py#L100) only.

DimitriPapadopoulos:
Fixes #560 when DocumentStartToken is preceded by `DirectiveToken`.

### DIFF
--- a/tests/rules/test_document_end.py
+++ b/tests/rules/test_document_end.py
@@ -71,3 +71,22 @@ class DocumentEndTestCase(RuleTestCase):
                    '---\n'
                    'third: document\n'
                    '...\n', conf, problem=(6, 1))
+
+    def test_directives(self):
+        conf = 'document-end: {present: true}'
+        self.check('%YAML 1.2\n'
+                   '---\n'
+                   'document: end\n'
+                   '...\n', conf)
+        self.check('%YAML 1.2\n'
+                   '%TAG ! tag:clarkevans.com,2002:\n'
+                   '---\n'
+                   'document: end\n'
+                   '...\n', conf)
+        self.check('---\n'
+                   'first: document\n'
+                   '...\n'
+                   '%YAML 1.2\n'
+                   '---\n'
+                   'second: document\n'
+                   '...\n', conf)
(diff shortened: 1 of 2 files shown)


tests/rules/test_document_end.py
```diff
@@ -71,3 +71,22 @@ def test_multiple_documents(self):
...
+                   '---\n'
+                   'document: end\n'
+                   '...\n', conf)
+        self.check('---\n'
+                   'first: document\n'
+                   '...\n'
+                   '%YAML 1.2\n'
+                   '---\n'
+                   'second: document\n'
+                   '...\n', conf)
```
adrienverge, line 92: We could also check that yamlllint complains if the `...` before `%YAML 1.2` are missing. Looking at [the YAML spec](https://yaml.org/spec/1.2.2/), it seems forbidden to do so:

```yaml
---
first: document
%YAML 1.2
---
second: document
...
```

But maybe this isn't really related to this pull request.

yamllint/rules/document_end.py
```diff
@@ -99,11 +99,13 @@ def check(conf, token, prev, next, nextnext, context):
         prev_is_end_or_stream_start = isinstance(
             prev, (yaml.DocumentEndToken, yaml.StreamStartToken)
         )
+        prev_is_directive = isinstance(prev, (yaml.DirectiveToken))
```
adrienverge, line 102: I allowed myself to update the pull request with a more Pythonic version:
```suggestion
        prev_is_directive = isinstance(prev, yaml.DirectiveToken)
```

adrienverge (approved):
Thanks for fixing this @DimitriPapadopoulos :+1:

tests/rules/test_document_end.py
DimitriPapadopoulos: Not directly related, but still belongs to `test_directives`.

PR #524, opened 2022-12-16
feat: Add --list-files command line option

splattael:
## What?

This PR adds the command line option `--list-files` which list files to lint and exits.

## Why?

This is useful to determine which files are being linted to check configuration changes like `yaml-files` and `ignore` more conveniently.

## Usage

```
$ yamllint -h
usage: yamllint [-h] [-] [-c CONFIG_FILE | -d CONFIG_DATA] [-f {parsable,standard,colored,github,auto}] [-s]
                [--no-warnings] [-v]
                [FILE_OR_DIR ...]

A linter for YAML files. yamllint does not only check for syntax validity, but for weirdnesses like key
repetition and cosmetic problems such as lines length, trailing spaces, indentation, etc.

positional arguments:
  FILE_OR_DIR           files to check

optional arguments:
  -h, --help            show this help message and exit
  -                     read from standard input
  -c CONFIG_FILE, --config-file CONFIG_FILE
                        path to a custom configuration
  -d CONFIG_DATA, --config-data CONFIG_DATA
                        custom configuration (as YAML source)
  --list-files      list files to lint and exit
  -f {parsable,standard,colored,github,auto}, --format {parsable,standard,colored,github,auto}
                        format for parsing output
  -s, --strict          return non-zero exit code on warnings as well as errors
  --no-warnings         output only error level problems
  -v, --version         show program's version number and exit
```

```bash
$ yamllint --list-files .
./.pre-commit-hooks.yaml
./yamllint/conf/default.yaml
./yamllint/conf/relaxed.yaml
./.github/workflows/ci.yaml
``` 

## Docs

Screenshot from 2023-01-07 12-07-45
Screenshot from 2023-01-07 12-07-24

## Refs

* https://github.com/adrienverge/yamllint/issues/396

PR #591, opened 2023-08-21
fix (#542): validate `ignore-from-file` in `validate_rule_conf`

### LINKED ISSUE #542: `ignore-from-file` ignored in `rule` definition
```
$ yamllint --version
yamllint 1.29.0
```
### Description

I have a project where a lot of yaml files have specific rules that need to be ignored.

In order to avoid duplicate code, I want to use the `ignore-from-file` feature in the rule definition but it is ignored.

```
$ cat .yamllint
---
extends: default

rules:
  empty-lines:
    ignore-from-file: .yamlignore
```
```
$ cat .yamlignore
/ignore-empty-lines.yml
```
```
$ cat ignore-empty-lines.yml
---


# These empty lines above just exist.
```

### Expected result

Yamllint should ignore this rule on this file.

### Actual result
```
$ yamllint .
./ignore-empty-lines.yml
  4:1       error    too many blank lines (3 > 2)  (empty-lines)
```

### Workaround
Use the `ignore` keyword. I have a lot of files that need to have multiple errors ignored. The final config file looks like this which leads to duplicate code:

```
$ cat .yamllint
---
extends: default

rules:
  empty-lines:
    ignore:
      - /ignore1.yml
      - /ignore2.yml
      - ...
      - /ignore42.yml
  comments:
    ignore:
      - /ignore1.yml
      - /ignore2.yml
      - ...
      - /ignore42.yml
```
___

The [documentation ](https://yamllint.readthedocs.io/en/v1.29.0/configuration.html#ignoring-paths) does not specify that this keyword is ignored in a `rules` block and the example implicitly gives the impression that it should work.

Either the documentation should be explicit about this behavior or yamllint should work as described above.

ndrwnaguib:
Fix #542. The `ignore-from-file` key wasn't validated in `validate_rule_conf`.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -212,6 +212,14 @@ class SimpleConfigTestCase(unittest.TestCase):
                                   '  colons:\n'
                                   '    ignore: yes\n')
 
+    def test_invalid_rule_ignore_from_file(self):
+        self.assertRaises(
+            config.YamlLintConfigError,
+            config.YamlLintConfig,
+            'rules:\n'
+            '  colons:\n'
+            '    ignore-from-file: 1337\n')
+
     def test_invalid_locale(self):
         with self.assertRaisesRegex(
                 config.YamlLintConfigError,
@@ -660,11 +668,18 @@ class IgnoreConfigTestCase(unittest.TestCase):
     def test_run_with_ignore_from_file(self):
         with open(os.path.join(self.wd, '.yamllint'), 'w') as f:
             f.write('extends: default\n'
-                    'ignore-from-file: .gitignore\n')
+                    'ignore-from-file: .gitignore\n'
+                    'rules:\n'
+                    '  key-duplicates:\n'
+                    '    ignore-from-file: .ignore-key-duplicates\n')
+
         with open(os.path.join(self.wd, '.gitignore'), 'w') as f:
             f.write('*.dont-lint-me.yaml\n'
                     '/bin/\n'
                     '!/bin/*.lint-me-anyway.yaml\n')
+
+        with open(os.path.join(self.wd, '.ignore-key-duplicates'), 'w') as f:
+            f.write('/ign-dup\n')
 
         sys.stdout = StringIO()
         with self.assertRaises(SystemExit):
@@ -686,10 +701,8 @@ class IgnoreConfigTestCase(unittest.TestCase):
             './file-at-root.yaml:3:3: ' + keydup,
             './file-at-root.yaml:4:17: ' + trailing,
             './file-at-root.yaml:5:5: ' + hyphen,
-            './ign-dup/file.yaml:3:3: ' + keydup,
             './ign-dup/file.yaml:4:17: ' + trailing,
             './ign-dup/file.yaml:5:5: ' + hyphen,
-            './ign-dup/sub/dir/file.yaml:3:3: ' + keydup,
             './ign-dup/sub/dir/file.yaml:4:17: ' + trailing,
             './ign-dup/sub/dir/file.yaml:5:5: ' + hyphen,
             './ign-trail/file.yaml:3:3: ' + keydup,
--- a/yamllint/config.py
+++ b/yamllint/config.py
@@ -153,8 +153,21 @@ def validate_rule_conf(rule, conf):
         return False
 
     if isinstance(conf, dict):
-        if ('ignore' in conf and
-                not isinstance(conf['ignore'], pathspec.pathspec.PathSpec)):
+        if ('ignore-from-file' in conf and not isinstance(
+                conf['ignore-from-file'], pathspec.pathspec.PathSpec)):
+            if isinstance(conf['ignore-from-file'], str):
+                conf['ignore-from-file'] = [conf['ignore-from-file']]
+            if not (isinstance(conf['ignore-from-file'], list)
+                    and all(isinstance(line, str)
+                            for line in conf['ignore-from-file'])):
+                raise YamlLintConfigError(
+                    'invalid config: ignore-from-file should contain '
+                    'valid filename(s), either as a list or string')
+            with fileinput.input(conf['ignore-from-file']) as f:
+                conf['ignore'] = pathspec.PathSpec.from_lines(
+                    'gitwildmatch', f)
+        elif ('ignore' in conf and not isinstance(
+                conf['ignore'], pathspec.pathspec.PathSpec)):
             if isinstance(conf['ignore'], str):
                 conf['ignore'] = pathspec.PathSpec.from_lines(
                     'gitwildmatch', conf['ignore'].splitlines())


yamllint/config.py
```diff
@@ -153,8 +153,29 @@ def validate_rule_conf(rule, conf):
         return False
 
     if isinstance(conf, dict):
-        if ('ignore' in conf and
-                not isinstance(conf['ignore'], pathspec.pathspec.PathSpec)):
+        if 'ignore-from-file' in conf and isinstance(
+            conf['ignore-from-file'], (str, list)
+        ):
```
adrienverge, line 158: Is `isinstance(conf['ignore-from-file'], (str, list))` useful? If yes, I don't understand, can you explain? (I would have copied the same logic as for `ignore`, a few lines below.)

It seems to me that it would prevent displaying a nice helpful error `invalid config: ...` if the user made a mistake like `ignore-from-file: 42`.

yamllint/config.py
```diff
@@ -153,8 +153,29 @@ def validate_rule_conf(rule, conf):
...
+            if not (
+                isinstance(conf['ignore-from-file'], list)
+                and all(
+                    isinstance(line, str) and os.path.isfile(line)
+                    for line in conf['ignore-from-file']
+                )
+            ):
+                raise YamlLintConfigError(
+                    'invalid config: ignore-from-file should contain '
+                    'valid filename(s), either as a list or string.'
```
adrienverge, line 170: I propose to reuse the exact same error string as [the not-inside-a-rule error](https://github.com/adrienverge/yamllint/commit/2f8ad7003a8571e0c935bc6ba7fef04ed6f3fde6#diff-443dfa037d6d01528b63511e2301a7bbb7d2ef8d3dc85b1df3e7fabd8b0d52f5R110-R111):
```python
                    'invalid config: ignore-from-file should contain '
                    'filename(s), either as a list or string'
```
(no `valid`, no `.`)

yamllint/config.py
```diff
@@ -153,8 +153,29 @@ def validate_rule_conf(rule, conf):
...
+                )
+            with fileinput.input(conf['ignore-from-file']) as ignore_file:
+                conf['ignore'] = pathspec.PathSpec.from_lines(
+                    'gitwildmatch', ignore_file
+                )
```
adrienverge, line 175: Same kind of remark, for [consistency](https://github.com/adrienverge/yamllint/commit/2f8ad7003a8571e0c935bc6ba7fef04ed6f3fde6#diff-443dfa037d6d01528b63511e2301a7bbb7d2ef8d3dc85b1df3e7fabd8b0d52f5R112-R113) and easier maintenance/debugging/code evolution: can you use `f` instead of `ignore_file`?

tests/test_config.py
```diff
@@ -212,6 +212,18 @@ def test_invalid_rule_ignore(self):
...
+        with self.assertRaisesRegex(
+            config.YamlLintConfigError,
+            'invalid config: ignore-from-file should contain '
+            r'valid filename\(s\), either as a list or string.',
+        ):
+            config.YamlLintConfig(
+                'rules:\n'
+                '  colons:\n'
+                '    ignore-from-file: /invalid_filename\n'
+            )
```
adrienverge, line 225: I wouldn't bother to use a regex here, since we don't for [other situations](https://github.com/adrienverge/yamllint/commit/2f8ad7003a8571e0c935bc6ba7fef04ed6f3fde6#diff-da24b355349b53e4c8047034767d008e4c2ce9ea6c8933a3acd50d0828f00528R470-R476).

This should be enough:
```python
        self.assertRaises(YamlLintConfigError,
                          config.YamlLintConfig,
                          'rules:\n'
                          '  colons:\n'
                          '    ignore-from-file: 1337\n')
```
(like for the `ignore` case just above)

yamllint/config.py
```diff
@@ -153,8 +153,29 @@ def validate_rule_conf(rule, conf):
...
-                not isinstance(conf['ignore'], pathspec.pathspec.PathSpec)):
+        if 'ignore-from-file' in conf and isinstance(
+            conf['ignore-from-file'], (str, list)
+        ):
+            if isinstance(conf['ignore-from-file'], str):
+                conf['ignore-from-file'] = [conf['ignore-from-file']]
+            if not (
+                isinstance(conf['ignore-from-file'], list)
+                and all(
+                    isinstance(line, str) and os.path.isfile(line)
```
adrienverge, line 164: Can you remove `and os.path.isfile(line)`? Because:
\- It's not present for the sibling not-inside-a-rule error.
\- If a file doesn't exists, it will caught just a few lines later with a `FileNotFoundError`, which is better suited.

adrienverge (commented):
Thanks for tackling this @ndrwnaguib!

It looks good, please see my comments. Also:
- Could you use the same code style as the rest of the code? (this project doesn't use the Black format)
- This needs a test to make sure the problem won't come back. Can you update `IgnoreConfigTestCase.test_run_with_ignored_from_file()` to include a rule-scoped `ignore-from-file` like done [in the `ignore` case](https://github.com/adrienverge/yamllint/blob/a68c3aa69e298ea20bdfb51ac388d5bfc4c347aa/tests/test_config.py#L615)?

yamllint/config.py
ndrwnaguib: > Is isinstance(conf['ignore-from-file'], (str, list)) useful? If yes, I don't understand, can you explain? (I would have copied the same logic as for ignore, a few lines below.)

I was only handling those two cases, so it seemed apt. However, I changed it to follow the same check as `ignore`'s.

tests/test_config.py
ndrwnaguib: Fixed that.

yamllint/config.py
ndrwnaguib: > Can you remove and os.path.isfile(line)? 

Sure.

yamllint/config.py
ndrwnaguib: > can you use f instead of ignore_file?

Yes, fixed that.

yamllint/config.py
ndrwnaguib: I am not sure why I chose to unnecessarily pedantic with this line, :). Anyway, I fixed that.

ndrwnaguib:
@adrienverge I rebased onto `master` using GitHub's update branch button. But, I checked the commit history, I do not see any merge commits, would you like me to drop it and rebase interactively instead?--unless you are going to squash the commits.

adrienverge:
Hello,

I allowed myself to force-push to your branch because 2 of my requests were not addressed.

Also the commit title referenced #524 which is unrelated. I amended that and reworded the commit for better understanding and maintenance.

Let's merge now.

ndrwnaguib:
> I allowed myself to force-push to your branch because 2 of my requests were not addressed.

That's alright. Sorry about missing them. 

> Also the commit title referenced #524 which is unrelated. I amended that and reworded the commit for better understanding and maintenance.

Thank you. I meant #542.

PR #596, opened 2023-09-12
tests: Apply pyupgrade suggestions

DimitriPapadopoulos:
* Simplify [super()](https://docs.python.org/3/library/functions.html#super)
  https://github.com/asottile/pyupgrade#super-calls
* UTF-8 is the default for [encode()](https://docs.python.org/3/library/stdtypes.html?highlight=encode#str.encode)
  https://github.com/asottile/pyupgrade#encodeutf-8


adrienverge (approved):
Better indeed :+1:

PR #598, opened 2023-09-23
Apply a couple refurb suggestions

DimitriPapadopoulos:
I kept only those improving readability.


adrienverge (approved):
Better indeed :+1:

PR #600, opened 2023-09-30
style: Use f-strings wherever possible

DimitriPapadopoulos:
They're faster.

### DIFF
--- a/tests/test_spec_examples.py
+++ b/tests/test_spec_examples.py
@@ -39,7 +39,7 @@ from tests.common import RuleTestCase
 #             text = text.replace('\u21d4', '')    # byte order mark
 #             text = text.replace('\u2192', '\t')  # right arrow
 #             text = text.replace('\u00b0', '')    # empty scalar
-#             with open('tests/yaml-1.2-spec-examples/%s' % id, 'w',
+#             with open(f'tests/yaml-1.2-spec-examples/{id}', 'w',
 #                       encoding='utf-8') as g:
 #                 g.write(text)
 
--- a/yamllint/config.py
+++ b/yamllint/config.py
@@ -76,7 +76,7 @@ class YamlLintConfig:
         try:
             conf = yaml.safe_load(raw_content)
         except Exception as e:
-            raise YamlLintConfigError('invalid config: %s' % e)
+            raise YamlLintConfigError(f'invalid config: {e}')
 
         if not isinstance(conf, dict):
             raise YamlLintConfigError('invalid config: not a dict')
@@ -95,7 +95,7 @@ class YamlLintConfig:
             try:
                 self.extend(base)
             except Exception as e:
-                raise YamlLintConfigError('invalid config: %s' % e)
+                raise YamlLintConfigError(f'invalid config: {e}')
 
         if 'ignore' in conf and 'ignore-from-file' in conf:
             raise YamlLintConfigError(
@@ -143,7 +143,7 @@ class YamlLintConfig:
             try:
                 rule = yamllint.rules.get(id)
             except Exception as e:
-                raise YamlLintConfigError('invalid config: %s' % e)
+                raise YamlLintConfigError(f'invalid config: {e}')
 
             self.rules[id] = validate_rule_conf(rule, self.rules[id])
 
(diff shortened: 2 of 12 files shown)


tests/test_spec_examples.py
```diff
@@ -39,8 +39,7 @@
 #             text = text.replace('\u21d4', '')    # byte order mark
 #             text = text.replace('\u2192', '\t')  # right arrow
 #             text = text.replace('\u00b0', '')    # empty scalar
-#             with open('tests/yaml-1.2-spec-examples/%s' % id, 'w',
-#                       encoding='utf-8') as g:
```
DimitriPapadopoulos, line 43: Note that `encoding='utf-8'` is the default.

yamllint/config.py
```diff
@@ -179,16 +179,16 @@ def validate_rule_conf(rule, conf):
...
             # Example: CONF = {option: (bool, 'mixed')}
             #          → {option: true}         → {option: mixed}
             if isinstance(options[optkey], tuple):
                 if (conf[optkey] not in options[optkey] and
                         type(conf[optkey]) not in options[optkey]):
                     raise YamlLintConfigError(
-                        'invalid config: option "%s" of "%s" should be in %s'
-                        % (optkey, rule.ID, options[optkey]))
+                        f'invalid config: option "{optkey}" of "{rule.ID} '
+                        f'should be in {options[optkey]}')
```
adrienverge, line 191: There's a missing `"` after `{rule.ID}`, making the tests fail:

```
FAIL: test_yes_no_for_booleans (tests.test_config.SimpleConfigTestCase)
yamllint.config.YamlLintConfigError: invalid config: option "indent-sequences" of "indentation should be in (<class 'bool'>, 'whatever', 'consistent')

During handling of the above exception, another exception occurred:

  File "/home/runner/work/yamllint/yamllint/tests/test_config.py", line 109, in test_yes_no_for_booleans
    c = config.YamlLintConfig('rules:\n'
AssertionError: "invalid config: option "indent-sequences" of "indentation" should be in " does not match "invalid config: option "indent-sequences" of "indentation should be in (<class 'bool'>, 'whatever', 'consistent')"
```

adrienverge (commented):
Looks good, thanks for taking the time!

Have you changed them one by one and checked they all make sense? Or did you use a script to find and replace them? (in some rare cases like `%02d`, `%.2f` or `%-10s`, f-string cannot be used as is)

yamllint/config.py
DimitriPapadopoulos: You're right about UTF-8, I will revert that:
https://docs.python.org/3/library/functions.html#open

DimitriPapadopoulos: Let me have a look at the rest of the formatting.

tests/test_spec_examples.py
DimitriPapadopoulos: You're right about UTF-8, I will revert that:
https://docs.python.org/3/library/functions.html#open

DimitriPapadopoulos:
> Have you changed them one by one and checked they all make sense? Or did you use a script to find and replace them? (in some rare cases like `%02d`, `%.2f` or `%-10s`, f-string cannot be used as is)

There are not such cases as far as I can see. Yet, it possible to write `f{variable!02d}` or `f{variable!.2f}`, isn't it?

adrienverge:
> There are not such cases as far as I can see. Yet, it possible to write `f{variable!02d}` or `f{variable!.2f}`, isn't it?

Sure it is! I just wanted to make sure you (or your custom script?) looked for them, if any :)

I still see a few `%s` and `%d` in some files (namely `yamllint/linter.py`, `yamllint/rules/line_length.py`, `yamllint/rules/indentation.py` and `yamllint/rules/octal_values.py`). Could you check them?

DimitriPapadopoulos:
As for the tool I use, it's [pyupgrade](https://github.com/asottile/pyupgrade). But I often end up applying changes manually, because:
1. You cannot limit `pyupgrade` to a single class of changes. I prefer an individual commit per class of changes.
2. It's not perfect: it often leaves behind some cases of `%` string interpolation.

DimitriPapadopoulos:
I have left an instance of `%` string interpolation in `yamllint/rules/line_length.py` because I find the f-string interpolation is less readable – unless of course I use additional variables to store the result of complex expressions.

adrienverge (approved):
> I have left an instance of `%` string interpolation in `yamllint/rules/line_length.py`

I had spotted this one too and I agree it's better readable like this :+1: 

I allowed myself to force-push to your branch with a slight rework of modified strings:
```diff
--- a/yamllint/config.py
+++ b/yamllint/config.py
@@ -179,8 +179,8 @@ def validate_rule_conf(rule, conf):
                 continue
             if optkey not in options:
                 raise YamlLintConfigError(
-                    f'invalid config: unknown option "{optkey}" '
-                    f'for rule "{rule.ID}"')
+                    f'invalid config: unknown option "{optkey}" for rule '
+                    f'"{rule.ID}"')
             # Example: CONF = {option: (bool, 'mixed')}
             #          → {option: true}         → {option: mixed}
             if isinstance(options[optkey], tuple):
@@ -216,8 +216,8 @@ def validate_rule_conf(rule, conf):
                 raise YamlLintConfigError(f'invalid config: {rule.ID}: {res}')
     else:
         raise YamlLintConfigError(
-            f'invalid config: rule "{rule.ID}": should be '
-            f'either "enable", "disable" or a dict')
+            f'invalid config: rule "{rule.ID}": should be either "enable", '
+            f'"disable" or a dict')
 
     return conf
 
--- a/yamllint/rules/indentation.py
+++ b/yamllint/rules/indentation.py
@@ -342,11 +342,11 @@ def _check(conf, token, prev, next, nextnext, context):
 
         if found_indentation != expected:
             if expected < 0:
-                message = f'wrong indentation: expected ' \
-                          f'at least {found_indentation + 1}'
+                message = f'wrong indentation: expected at least ' \
+                          f'{found_indentation + 1}'
             else:
-                message = f'wrong indentation: expected {expected} ' \
-                          f'but found {found_indentation}'
+                message = f'wrong indentation: expected {expected} but ' \
+                          f'found {found_indentation}'
             yield LintProblem(token.start_mark.line + 1,
                               found_indentation + 1, message)
 
```

Thank you Dimitri, let's merge!

PR #604, opened 2023-11-08
Add forbid-in-block-sequences feature for detecting empty values in block sequences

ericalmeidasp:
feat: Add `forbid-in-list-items` feature to detect empty values in lists

This commit introduces a new feature, `forbid-in-list-items`, which allows users to check for empty values within lists. The default behavior is set to `false`, but it can be activated by setting the `forbid-in-list-items` configuration variable to `true`.

The `check` function has been updated to handle this feature, and it will now identify and report empty values within list items when `forbid-in-list-items` is enabled.

This enhancement provides users with more control and flexibility when using the linter for YAML files.

### DIFF
--- a/tests/rules/test_empty_values.py
+++ b/tests/rules/test_empty_values.py
@@ -42,7 +42,8 @@ class EmptyValuesTestCase(RuleTestCase):
 
     def test_in_block_mappings_disabled(self):
         conf = ('empty-values: {forbid-in-block-mappings: false,\n'
-                '               forbid-in-flow-mappings: false}\n')
+                '               forbid-in-flow-mappings: false,\n'
+                '               forbid-in-block-sequences: false}\n')
         self.check('---\n'
                    'foo:\n', conf)
         self.check('---\n'
@@ -51,7 +52,8 @@ class EmptyValuesTestCase(RuleTestCase):
 
     def test_in_block_mappings_single_line(self):
         conf = ('empty-values: {forbid-in-block-mappings: true,\n'
-                '               forbid-in-flow-mappings: false}\n')
+                '               forbid-in-flow-mappings: false,\n'
+                '               forbid-in-block-sequences: false}\n')
         self.check('---\n'
                    'implicitly-null:\n', conf, problem1=(2, 17))
         self.check('---\n'
@@ -63,7 +65,8 @@ class EmptyValuesTestCase(RuleTestCase):
 
     def test_in_block_mappings_all_lines(self):
         conf = ('empty-values: {forbid-in-block-mappings: true,\n'
-                '               forbid-in-flow-mappings: false}\n')
+                '               forbid-in-flow-mappings: false,\n'
+                '               forbid-in-block-sequences: false}\n')
         self.check('---\n'
                    'foo:\n'
                    'bar:\n'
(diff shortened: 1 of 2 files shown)


tests/rules/test_empty_values.py
```diff
@@ -258,3 +258,60 @@ def test_in_flow_mappings_comments(self):
                    problem1=(4, 7),
                    problem2=(7, 9),
                    problem3=(10, 5))
+
+    def test_in_list_items_disabled(self):
+        conf = ('empty-values: {forbid-in-block-mappings: false,\n'
+                '               forbid-in-flow-mappings: false,\n'
+                '               forbid-in-list-items: false}\n'
+                'braces: disable\n'
+                'commas: disable\n')
```
adrienverge, line 267: Are these needed?

I guess you copied-pasted them from an example with *flow mappings*, which use `{` and `,` characters. Block sequences don't :)

adrienverge (commented):
Hello Éric, and thanks for contributing to yamllint!

For consistency with [YAML vocabulary](https://yaml.org/spec/1.2.2/#821-block-sequences) and already-existing yamllint options `forbid-in-block-mappings` and `forbid-in-flow-mappings`, this new feature should be named `forbid-in-block-sequences` (and a new option `forbid-in-flow-sequences` shouldn't be necessary because I believe `[1, , 3]` is invalid YAML syntax). Same for the documentation paragraph, the `LintProblem` message and test functions names.

Since the rule `empty-values` is disabled in *default* and *relaxed* configurations, I would set this option enabled by default, like `forbid-in-block-mappings` and `forbid-in-flow-mappings` are. It will create new errors or warnings for some users with specific configurations, but it's easy to fix. What do you think?

tests/rules/test_empty_values.py
ericalmeidasp: sorry, you're right, adjusted.

ericalmeidasp:
Hello! Thank you so much for the code review, 

In response to the valuable feedback provided, I have made the following changes:

1. Renamed the feature from `forbid-in-list-items` to `forbid-in-block-sequences`.

2. Updated the documentation paragraph, LintProblem message, and test function names to align with the new feature name.

3. Set the `forbid-in-block-sequences` option to be enabled by default, similar to `forbid-in-block-mappings` and `forbid-in-flow-mappings`. I agree with u, adjusted to maintain consistency.

In addition to these changes, I've also added more test cases to cover various scenarios. However, due to the `forbid-in-block-sequences` option being set as the default, I had to modify the existing test cases by including `forbid-in-block-sequences: false` to avoid interference with the existing test scenarios in the test_empty_values.py file.

Thanks!

adrienverge (approved):
This looks perfect, thank you! I'm merging this so it goes in the next yamllint release.

⚠️ **To all yamllint users: this is a BREAKING CHANGE if the `empty-values` rule is enabled in your configuration. But if it's the case, you probably want this new option `forbid-in-block-sequences` enabled too, for consistency. In case you want to disable it, just use this configuration in your `.yamllint`:**
```yaml
rules:
  empty-values:
    forbid-in-block-mappings: true
    forbid-in-flow-mappings: true
    forbid-in-block-sequences: false
```

PR #606, opened 2023-11-09
hyphen: Add min-spaces-after and check-scalars options

sbaudoin:
This permits checking style when people want sequences to have a fixed number of spaces greater than 1 for example.
We may also use these options to identify typos where we expect everything starting with an hyphen to be a sequence.


yamllint/rules/hyphens.py
```diff
@@ -72,24 +87,97 @@
...
-        if problem is not None:
-            yield problem
+        if conf['max-spaces-after'] > 0:
+            problem = spaces_after(token, prev, next,
+                                   max=conf['max-spaces-after'],
+                                   max_desc='too many spaces after hyphen')
+            if problem is not None:
+                yield problem
+
+        if conf['min-spaces-after'] > 0:
```
adrienverge, line 168: Doing the check if `conf['min-spaces-after'] == 1` is useless (it cannot be lower than 1) and will degrade performance.

I propose:
```suggestion
        if conf['min-spaces-after'] > 1:
```

adrienverge (commented):
Hello and thanks for contributing!

The `min-spaces-after` change is legitimate, and could be merged soon. Please see my proposals in the comments below.


About the `check-scalars` option, I'm not sure I'm favorable to it. And checking every scalar for a starting `-` in a YAML file is probably very bad for performance. But maybe I don't understand its use and usefulness for yamllint. Can you please propose it in a different pull request or issue, with example use-cases?

(For info, the examples you give:
```yaml
object: [elem1, ...]
-bar:
```
would not be valid YAML anyway if the scalar `-bar:` was turned into a list item `- bar:`.)

sbaudoin:
I'm going to push additional commits. For a cleaner history, will you do the rebase or should I do it on my side?

About the rational behind `check-scalars`, this came from the same code extract as the one you give:
```
foo:
     -broken
```
From a YAML point of view this is a valid scalar syntax but as we see that the scalar starts with an hyphen we may think this syntax is a typo. On the first sight we may think that `hyphen-min-spaces` will identify that but not as we are not considering a sequence because this is a scalar. So I added `check-scalars` to be able to check such situations. But I agree that this might lead to poor performances and may trigger a lot of false positive (that's explain in the doc).

From my point of view, `hyphen-min-spaces` would be incomplete without this additional option; we can go without it but we must be clear about such situations not being taken into account.

adrienverge:
> I'm going to push additional commits. For a cleaner history, will you do the rebase or should I do it on my side?

OK! The best would be to rebase and squash into one commit with a nice title, ready to be merged.

> From my point of view, `hyphen-min-spaces` would be incomplete without this additional option; we can go without it but we must be clear about such situations not being taken into account.

Thanks for the explanation.
I'm +1 to be more clear about such situations not taken into account (and the reason is simple: they are valid YAML). The `quoted-strings` rule can also be enabled to avoid such problems.

PR #608, opened 2023-11-11
Add rules for key types

### LINKED ISSUE #607: Need rules for restricting key types
We need rules for restricting the types that can be used for **keys** in `key: value` maps.

### Node kinds

According to the YAML spec, [nodes](https://yaml.org/spec/1.2.2/#nodes) can be one of three kinds:
- scalar
- sequence
- mapping

I would like to have a rule that specifies that a **key** node must not be a sequence or a mapping.

### Scalar types

I would like to also have a rule that can control which type of scalar can be used for **keys.** The scalar types to consider are:
- str
- int
- float
- bool
- null
- timestamp

For example, JSON only supports string keys, so if I am converting YAML to JSON I want to check that all **keys** are strings.

HenryGessau:
Closes #607


HenryGessau:
@adrienverge Do you have any thoughts or opinions on this?

HenryGessau:
Not aligned with the purpose of yamllint, per https://github.com/adrienverge/yamllint/issues/607#issuecomment-1821792790

PR #620, opened 2023-12-30
tests: Prevent `unittest --buffer` from crashing

Jayman2000:
Before this change, if certain tests were failing in certain ways, then running `python -m unittest --buffer` would cause an `AttributeError` in the `unittest` module itself.

Here’s what `unittest` does when you use the `--buffer` argument:

1. [It sets `sys.stdout` and `sys.stderr` to `StringIO`s][1].
2. It runs a test.
3. [If the test failed, it runs `getvalue()` on `sys.stdout` and `sys.stderr` to get data from its `StringIO`s][2].

`tests/test_cli.py` has its own `RunContext` class that does something similar. Before this change, here’s what could happen:

1. `unittest` sets `sys.stdout` and `sys.stderr` to `StringIO`s.
2. `unittest` runs a test that uses `RunContext`.
3. A `RunContext` gets entered. It sets `sys.stdout` and `sys.stderr` to its own `StringIO`s.
4. The `RunContext` gets exited. It sets `sys.stdout` and `sys.stderr` to `sys.__stdout__` and `sys.__stderr__`.
5. The test fails.
6. `unittest` assumes that `sys.stdout` is still set to one of its `StringIO`s, and runs `sys.stdout.getvalue()`.
7. `unittest` crashes with this error:

    ```
    AttributeError: '_io.TextIOWrapper' object has no attribute 'getvalue'
    ```

[1]: <https://github.com/python/cpython/blob/2305ca51448552542b2414186252123a8dc87db7/Lib/unittest/result.py#L65>
[2]: <https://github.com/python/cpython/blob/2305ca51448552542b2414186252123a8dc87db7/Lib/unittest/result.py#L87>

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -38,13 +38,17 @@ class RunContext:
 
     def __enter__(self):
         self._raises_ctx.__enter__()
+        self.old_sys_stdout = sys.stdout
+        self.old_sys_stderr = sys.stderr
         sys.stdout = self.outstream = StringIO()
         sys.stderr = self.errstream = StringIO()
         return self
 
     def __exit__(self, *exc_info):
-        self.stdout, sys.stdout = self.outstream.getvalue(), sys.__stdout__
-        self.stderr, sys.stderr = self.errstream.getvalue(), sys.__stderr__
+        self.stdout = self.outstream.getvalue()
+        self.stderr = self.errstream.getvalue()
+        sys.stdout = self.old_sys_stdout
+        sys.stderr = self.old_sys_stderr
         return self._raises_ctx.__exit__(*exc_info)
 
     @property


tests/test_cli.py
```diff
@@ -38,13 +38,17 @@ def __init__(self, case):
...
         sys.stderr = self.errstream = StringIO()
         return self
 
     def __exit__(self, *exc_info):
-        self.stdout, sys.stdout = self.outstream.getvalue(), sys.__stdout__
-        self.stderr, sys.stderr = self.errstream.getvalue(), sys.__stderr__
+        sys.stdout = self.old_sys_stdout
+        sys.stderr = self.old_sys_stderr
+        self.stdout = self.outstream.getvalue()
+        self.stderr = self.errstream.getvalue()
```
adrienverge, line 51: To keep the reverse order, can you move these 2 lines first?
```python
def __enter__(self):
    step_1
    step_2
    step_3

def __exit__(self, *exc_info):
    step_3
    step_2
    step_1
```

adrienverge (commented):
Thanks, I confirm the problem (although [Python documentation about `sys.__stdout__`](https://docs.python.org/3/library/sys.html#sys.__stdout__) doesn't warn about potential issues).

And your fix seems to work!

I'm curious: why and when do you need `--buffer`?

tests/test_cli.py
Jayman2000: Done.

Jayman2000:
> I'm curious: why and when do you need `--buffer`?

I don’t know. I tried running the tests on my local system when I ran into #621. I tried using `--buffer` as a part of my debugging process, and when it made things worse, I started investigating. I had only tried `--buffer` because it was mentioned on some StackOverflow question that I had found.

adrienverge (approved):
OK!

Thanks for contributing :+1:

PR #624, opened 2023-12-31
cli: remove format() left over in f63e56f

DimitriPapadopoulos:
See #600 / f63e56f.
- The format() call was missing in the first place for `filename`.
- Other format() calls are obviously useless now that f-strings perform string interpolation.


adrienverge (approved):
Indeed. Thanks!

PR #627, opened 2023-12-31
style: apply ruff suggestion

DimitriPapadopoulos:
Fixes the only error reported by [ruff](https://astral.sh/ruff) using default settings:
```
E721 Do not compare types, use `isinstance()`
```


adrienverge (approved):
Thanks ruff and thank you Dimitri! Better indeed.

PR #629, opened 2023-12-31
build: Replace flake8 with ruff

### LINKED ISSUE #628: flake8 → ruff?
How about changing [flake8](https://flake8.pycqa.org/) to [ruff](https://astral.sh/ruff)? While I wasn't interested initially, ruff does gain momentum, and beyond speed, it replaces all Python linters by a single tool, and is less opinionated than say [black](https://black.readthedocs.io/).

These files would be slightly modified:
* [CONTRIBUTING.rst](https://github.com/adrienverge/yamllint/blob/152ba20f339588a872777eafb888f0073e83cafb/CONTRIBUTING.rst?plain=1#L26)
* [.github/workflows/ci.yaml](https://github.com/adrienverge/yamllint/blob/152ba20f339588a872777eafb888f0073e83cafb/.github/workflows/ci.yaml#L23-L26)
* [pyproject.toml](https://github.com/adrienverge/yamllint/blob/152ba20f339588a872777eafb888f0073e83cafb/pyproject.toml#L29-L30)

The only change in code suggested using the default settings of ruff:
* #627

In addition to being user as a linter, ruff _may_ be used as a formatter, as a less opinionated alternative to black (single quotes are OK).

Finally, it would tick a box in the issues reported by the [Repo-Review](https://learn.scientific-python.org/development/guides/repo-review/?repo=adrienverge%2Fyamllint&branch=master) of [Scientific Python](https://scientific-python.org/):
* [RF001](https://learn.scientific-python.org/development/guides/style#RF001): Has Ruff config
  Must have `[tool.ruff]` section in `pyproject.toml` or `ruff.toml`/`.ruff.toml`.

DimitriPapadopoulos:
- Enable [isort (I)](https://docs.astral.sh/ruff/rules/#isort-i) rules, in addition to the [default set of rules](https://docs.astral.sh/ruff/settings/#select).

Fixes #628.

Requires to be rebased on top of #627 to pass tests.


pyproject.toml
```diff
@@ -26,8 +26,7 @@ dynamic = ["version"]
 [project.optional-dependencies]
 dev = [
     "doc8",
-    "flake8",
-    "flake8-import-order",
+    "ruff",
```
adrienverge, line 29: To keep alphabetical order, it needs to be after `rstcheck`.

adrienverge (commented):
Thanks Dimitri, I approve this :+1:

I don't understand the rule "only one blank line between imports and variables", but if ruff suggests it, I guess it's better.

DimitriPapadopoulos:
Perhaps I can find a ruff setting for that, ruff is more flexible than black. Let me have a look. But yes, I would recommend endorsing whatever ruff suggests and not thinking too much about such details, as far as readability is not impaired.

DimitriPapadopoulos:
Can you have a look. Now ruff adds a line where there was only 1 line. Is `lines-after-imports = 2` really worth it?

tests/rules/test_quoted_strings.py
```diff
@@ -14,7 +14,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from tests.common import RuleTestCase
-
```
adrienverge, line 17: Is there a config to keep this blank line? `tests` are standard, while `yamllint` is third-party.

adrienverge (commented):
> Can you have a look? Now ruff adds a line where there was only 1 line. Is lines-after-imports = 2 really worth it?

Yeah I agree, it's not worth it.

tests/rules/test_quoted_strings.py
DimitriPapadopoulos: I think it might classify `tests` as first-party just like `yamllint`, because it discovers the `tests` directory. Let me try these:
```
[lint.isort]
extra-standard-library = ["tests"]
```
```
[lint.isort]
known-first-party = ["yamllint"]
```

DimitriPapadopoulos: I pushed this setting:
```toml
[lint.isort]
known-first-party = ["yamllint"]
```
Do you like it?

DimitriPapadopoulos: Wrong one. I thought it worked but it doesn't.

DimitriPapadopoulos: This one works slightly better:
```toml
[lint.isort]
extra-standard-library = ["tests"]
```
That's still not what we want, because tests are definitely **not** standard. Rather they are another category of first-party, or maybe third-party.

DimitriPapadopoulos: Therefore I think the closest to what you want is:
```toml
[lint.isort]
known-third-party = ["tests"]
```

adrienverge (approved):
Thanks a lot @DimitriPapadopoulos!

PR #630, opened 2024-01-03
Automatically detect character encoding of YAML files and ignore files

### LINKED ISSUE #218: Can not parse utf-8 strings
```
  File "C:\msys64\mingw64\lib\python3.8\runpy.py", line 192, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\msys64\mingw64\lib\python3.8\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "c:\msys64\mingw64\bin\yamllint.exe\__main__.py", line 7, in <module>
  File "C:\msys64\mingw64\lib\python3.8\site-packages\yamllint\cli.py", line 189, in run
    prob_level = show_problems(problems, 'stdin', args_format=args.format)
  File "C:\msys64\mingw64\lib\python3.8\site-packages\yamllint\cli.py", line 91, in show_problems
    for problem in problems:
  File "C:\msys64\mingw64\lib\python3.8\site-packages\yamllint\linter.py", line 198, in _run
    syntax_error = get_syntax_error(buffer)
  File "C:\msys64\mingw64\lib\python3.8\site-packages\yamllint\linter.py", line 179, in get_syntax_error
    list(yaml.parse(buffer, Loader=yaml.BaseLoader))
  File "C:\msys64\mingw64\lib\python3.8\site-packages\yaml\__init__.py", line 73, in parse
    loader = Loader(stream)
  File "C:\msys64\mingw64\lib\python3.8\site-packages\yaml\loader.py", line 14, in __init__
    Reader.__init__(self, stream)
  File "C:\msys64\mingw64\lib\python3.8\site-packages\yaml\reader.py", line 74, in __init__
    self.check_printable(stream)
  File "C:\msys64\mingw64\lib\python3.8\site-packages\yaml\reader.py", line 143, in check_printable
    raise ReaderError(self.name, position, ord(character),
yaml.reader.ReaderError: unacceptable character #xdc82: special characters are not allowed
  in "<unicode string>", position 279
```
I know https://github.com/adrienverge/yamllint/issues/20 and https://github.com/adrienverge/yamllint/issues/2. But it's on non-Windows. On Windows, LANG, LC_CTYPE does not set in generally. I think yamllint should provide way to read utf-8 string even if LANG/LC_CTYPE is not set.

### LINKED ISSUE #238: Failure to read file in 1.21.0 with Python 2.7 and non-ascii yaml files.

### LINKED ISSUE #347: Character set issue on Windows
I'm seeing the following error when running yamllint on a UTF-8 encoded file that includes characters that don't appear in the Windows 1252 character set:

```
  File "c:\users\mpowell\appdata\local\programs\python\python38-32\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\mpowell\appdata\local\programs\python\python38-32\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\mpowell\AppData\Roaming\Python\Python38\Scripts\yamllint.exe\__main__.py", line 7, in <module>
  File "C:\Users\mpowell\AppData\Roaming\Python\Python38\site-packages\yamllint\cli.py", line 206, in run
    problems = linter.run(f, conf, filepath)
  File "C:\Users\mpowell\AppData\Roaming\Python\Python38\site-packages\yamllint\linter.py", line 240, in run
    content = input.read()
  File "c:\users\mpowell\appdata\local\programs\python\python38-32\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 96: character maps to <undefined>
```

To reproduce, run yamllint on a file containing the following:
```yaml
# “these quotes” will cause the error
```

If I edit line 205 in yamllint's `cli.py` file like this:
```python
--- with io.open(file, newline='') as f:
+++ with io.open(file, newline='', encoding='utf-8') as f:
```
the error goes away. It seems `io.open` is just using the default Windows character set, rather than the actual encoding of the file.

I don't know much python so I doubt this is a robust enough fix to submit as a PR, but if I'm wrong let me know and I'll gladly submit the PR!

Maybe adding a character set parameter to the CLI would be a more flexible way to address?

Jayman2000:
This PR makes sure that yamllint never uses `open()`’s default encoding. Specifically, it uses the character encoding detection algorithm specified in [chapter 5.2 of the YAML spec](https://yaml.org/spec/1.2.2/#52-character-encodings) when reading both YAML files and files that are on [the `ignore-from-file` list](https://yamllint.readthedocs.io/en/stable/configuration.html#ignoring-paths).

There are [two](https://github.com/adrienverge/yamllint/pull/240) [other](https://github.com/adrienverge/yamllint/pull/581) PRs that are similar to this one. Here’s how this PR compares to those two:

- This PR doesn’t have any merge conflicts.
- This PR has a cleaner commit history. You can run the tests and flake8 on each commit in this PR, and they’ll report no errors. I don’t think that you can do that with #240.
- This PR has longer commit messages. I really tried to explain why I think that my changes make sense.
- This PR detects the encoding of files being linted, config files, and files on the `ignore-from-file` list. Those two PRs only detects the encoding of files being linted.
- #240 PR adds a dependency on [`chardet`](https://pypi.org/project/chardet/). This PR doesn’t add any dependencies.
- This PR only supports UTF-8, UTF-16 and UTF-32. Both of those PRs support additional encodings.
- #581 adds support for running tests on Windows. This PR doesn’t.
- The code that this PR adds to the `yamllint` package is simpler.
- The code that this PR adds to the `test` package is much more complicated, but hopefully it tests things more thoroughly.

Fixes #218. Fixes #238. Fixes #347.
Closes #240. Closes #581.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -228,6 +228,10 @@ or:
 
 .. note:: However, this is mutually exclusive with the ``ignore`` key.
 
+.. note:: Files on the ``ignore-from-file`` list should use either UTF-8,
+   UTF-16 or UTF-32. See :doc:`Character Encoding <character_encoding>` for
+   details and workarounds.
+
 If you need to know the exact list of files that yamllint would process,
 without really linting them, you can use ``--list-files``:
 
--- a/tests/common.py
+++ b/tests/common.py
@@ -150,8 +150,9 @@ def build_temp_workspace(files):
         else:
             if isinstance(content, Blob):
                 content = content.text.encode(content.encoding)
-            mode = 'wb' if isinstance(content, bytes) else 'w'
-            with open(path, mode) as f:
+            elif isinstance(content, str):
+                content = content.encode('utf_8')
+            with open(path, 'wb') as f:
                 f.write(content)
 
     return tempdir
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,4 +1,5 @@
 # Copyright (C) 2016 Adrien Vergé
+# Copyright (C) 2023–2025 Jason Yundt
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -13,6 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import codecs
 import contextlib
 from io import StringIO
 import os
@@ -20,6 +22,8 @@ import shutil
 import sys
 import tempfile
 import unittest
+import warnings
+from codecs import CodecInfo
 
 import yaml
 
@@ -27,6 +31,155 @@ from yamllint import linter
 from yamllint.config import YamlLintConfig
 
 
+# Encoding related stuff:
+UTF_CODECS = (
+    'utf_32_be',
+    'utf_32_be_sig',
+    'utf_32_le',
+    'utf_32_le_sig',
+    'utf_16_be',
+    'utf_16_be_sig',
+    'utf_16_le',
+    'utf_16_le_sig',
+    'utf_8',
+    'utf_8_sig'
+)
+
+
+def encode_utf_32_be_sig(obj):
+    return (
+        codecs.BOM_UTF32_BE + codecs.encode(obj, 'utf_32_be', 'strict'),
+        len(obj)
+    )
+
+
+def encode_utf_32_le_sig(obj):
+    return (
+        codecs.BOM_UTF32_LE + codecs.encode(obj, 'utf_32_le', 'strict'),
+        len(obj)
+    )
+
+
+def encode_utf_16_be_sig(obj):
+    return (
+        codecs.BOM_UTF16_BE + codecs.encode(obj, 'utf_16_be', 'strict'),
+        len(obj)
+    )
+
+
+def encode_utf_16_le_sig(obj):
+    return (
+        codecs.BOM_UTF16_LE + codecs.encode(obj, 'utf_16_le', 'strict'),
+        len(obj)
+    )
+
+
+test_codec_infos = {
+    'utf_32_be_sig':
+    CodecInfo(encode_utf_32_be_sig, codecs.getdecoder('utf_32')),
+    'utf_32_le_sig':
+    CodecInfo(encode_utf_32_le_sig, codecs.getdecoder('utf_32')),
+    'utf_16_be_sig':
+    CodecInfo(encode_utf_16_be_sig, codecs.getdecoder('utf_16')),
+    'utf_16_le_sig':
+    CodecInfo(encode_utf_16_le_sig, codecs.getdecoder('utf_16')),
+}
+
+
+def register_test_codecs():
+    codecs.register(test_codec_infos.get)
+
+
+def unregister_test_codecs():
+    if sys.version_info >= (3, 10, 0):
+        codecs.unregister(test_codec_infos.get)
+    else:
+        warnings.warn(
+            "This version of Python doesn’t allow us to unregister codecs.",
+            stacklevel=1
+        )
+
+
+def is_test_codec(codec):
+    return codec in test_codec_infos.keys()
+
+
+def test_codec_built_in_equivalent(test_codec):
+    return_value = test_codec
+    for suffix in ('_sig', '_be', '_le'):
+        return_value = return_value.replace(suffix, '')
+    return return_value
+
+
+def uses_bom(codec):
+    for suffix in ('_32', '_16', '_sig'):
+        if codec.endswith(suffix):
+            return True
+    return False
+
+
+def encoding_detectable(string, codec):
+    """
+    Returns True if encoding can be detected after string is encoded
+
+    Encoding detection only works if you’re using a BOM or the first character
+    is ASCII. See yamllint.decoder.auto_decode()’s docstring.
+    """
+    return uses_bom(codec) or (len(string) > 0 and string[0].isascii())
+
+
+# Workspace related stuff:
+class Blob:
+    def __init__(self, text, encoding):
+        self.text = text
+        self.encoding = encoding
+
+
+def build_temp_workspace(files):
+    tempdir = tempfile.mkdtemp(prefix='yamllint-tests-')
+
+    for path, content in files.items():
+        path = os.fsencode(os.path.join(tempdir, path))
+        if not os.path.exists(os.path.dirname(path)):
+            os.makedirs(os.path.dirname(path))
+
+        if isinstance(content, list):
+            os.mkdir(path)
+        elif isinstance(content, str) and content.startswith('symlink://'):
+            os.symlink(content[10:], path)
+        else:
+            if isinstance(content, Blob):
+                content = content.text.encode(content.encoding)
+            mode = 'wb' if isinstance(content, bytes) else 'w'
+            with open(path, mode) as f:
+                f.write(content)
+
+    return tempdir
+
+
+@contextlib.contextmanager
+def temp_workspace(files):
+    """Provide a temporary workspace that is automatically cleaned up."""
+    backup_wd = os.getcwd()
+    wd = build_temp_workspace(files)
+
+    try:
+        os.chdir(wd)
+        yield
+    finally:
+        os.chdir(backup_wd)
+        shutil.rmtree(wd)
+
+
+def temp_workspace_with_files_in_many_codecs(path_template, text):
+    workspace = {}
+    for codec in UTF_CODECS:
+        if encoding_detectable(text, codec):
+            workspace[path_template.format(codec)] = Blob(text, codec)
+    return workspace
+
+
+# Miscellaneous stuff:
 class RuleTestCase(unittest.TestCase):
     def build_fake_config(self, conf):
         if conf is None:
@@ -81,37 +234,3 @@ class RunContext:
     @property
     def returncode(self):
         return self._raises_ctx.exception.code
-
-
-def build_temp_workspace(files):
-    tempdir = tempfile.mkdtemp(prefix='yamllint-tests-')
-
-    for path, content in files.items():
-        path = os.fsencode(os.path.join(tempdir, path))
-        if not os.path.exists(os.path.dirname(path)):
-            os.makedirs(os.path.dirname(path))
-
-        if isinstance(content, list):
-            os.mkdir(path)
-        elif isinstance(content, str) and content.startswith('symlink://'):
-            os.symlink(content[10:], path)
-        else:
-            mode = 'wb' if isinstance(content, bytes) else 'w'
-            with open(path, mode) as f:
-                f.write(content)
-
-    return tempdir
-
-
-@contextlib.contextmanager
-def temp_workspace(files):
-    """Provide a temporary workspace that is automatically cleaned up."""
-    backup_wd = os.getcwd()
-    wd = build_temp_workspace(files)
-
-    try:
-        os.chdir(wd)
-        yield
-    finally:
-        os.chdir(backup_wd)
-        shutil.rmtree(wd)
--- a/tests/common.py
+++ b/tests/common.py
@@ -87,7 +87,7 @@ def build_temp_workspace(files):
     tempdir = tempfile.mkdtemp(prefix='yamllint-tests-')
 
     for path, content in files.items():
-        path = os.path.join(tempdir, path).encode('utf-8')
+        path = os.fsencode(os.path.join(tempdir, path))
         if not os.path.exists(os.path.dirname(path)):
             os.makedirs(os.path.dirname(path))
 
--- a/tests/test_decoder.py
+++ b/tests/test_decoder.py
@@ -1,0 +1,454 @@
+# Copyright (C) 2023–2025 Jason Yundt
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import codecs
+import os
+import unittest
+
+from tests.common import (
+    UTF_CODECS,
+    encoding_detectable,
+    is_test_codec,
+    register_test_codecs,
+    test_codec_built_in_equivalent,
+    unregister_test_codecs,
+    uses_bom,
+)
+
+from yamllint import decoder
+
+
+class PreEncodedTestStringInfo:
+    def __init__(
+        self,
+        input_bytes,
+        codec_for_input_bytes,
+        expected_output_str
+    ):
+        self.input_bytes = input_bytes
+        self.codec_for_input_bytes = codec_for_input_bytes
+        self.expected_output_str = expected_output_str
+
+
+PRE_ENCODED_TEST_STRING_INFOS = (
+    # An empty string
+    PreEncodedTestStringInfo(
+        b'',
+        None,
+        ''
+    ),
+
+    # A single ASCII character
+    PreEncodedTestStringInfo(
+        b'\x00\x00\x00|',
+        'utf_32_be',
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'\x00\x00\xfe\xff\x00\x00\x00|',
+        'utf_32',
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'|\x00\x00\x00',
+        'utf_32_le',
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xff\xfe\x00\x00|\x00\x00\x00',
+        'utf_32',  # LE with BOM
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'\x00|',
+        'utf_16_be',
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xfe\xff\x00|',
+        'utf_16',  # BE with BOM
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'|\x00',
+        'utf_16_le',
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xff\xfe|\x00',
+        'utf_16',   # LE with BOM
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'|',
+        'utf_8',
+        '|'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xef\xbb\xbf|',
+        'utf_8_sig',
+        '|'
+    ),
+
+    # A string that starts with an ASCII character
+    PreEncodedTestStringInfo(
+        b'\x00\x00\x00W\x00\x00\x00h\x00\x00\x00a\x00\x00\x00t\x00\x00 \x19\x00\x00\x00s\x00\x00\x00 \x00\x00\x00u\x00\x00\x00p\x00\x00\x00?',  # noqa: E501
+        'utf_32_be',
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\x00\x00\xfe\xff\x00\x00\x00W\x00\x00\x00h\x00\x00\x00a\x00\x00\x00t\x00\x00 \x19\x00\x00\x00s\x00\x00\x00 \x00\x00\x00u\x00\x00\x00p\x00\x00\x00?',  # noqa: E501
+        'utf_32',  # BE with BOM
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'W\x00\x00\x00h\x00\x00\x00a\x00\x00\x00t\x00\x00\x00\x19 \x00\x00s\x00\x00\x00 \x00\x00\x00u\x00\x00\x00p\x00\x00\x00?\x00\x00\x00',  # noqa: E501
+        'utf_32_le',
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xff\xfe\x00\x00W\x00\x00\x00h\x00\x00\x00a\x00\x00\x00t\x00\x00\x00\x19 \x00\x00s\x00\x00\x00 \x00\x00\x00u\x00\x00\x00p\x00\x00\x00?\x00\x00\x00',  # noqa: E501
+        'utf_32',  # LE with BOM
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\x00W\x00h\x00a\x00t \x19\x00s\x00 \x00u\x00p\x00?',
+        'utf_16_be',
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xfe\xff\x00W\x00h\x00a\x00t \x19\x00s\x00 \x00u\x00p\x00?',
+        'utf_16',  # BE with BOM
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'W\x00h\x00a\x00t\x00\x19 s\x00 \x00u\x00p\x00?\x00',
+        'utf_16_le',
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xff\xfeW\x00h\x00a\x00t\x00\x19 s\x00 \x00u\x00p\x00?\x00',
+        'utf_16',  # LE with BOM
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'What\xe2\x80\x99s up?',
+        'utf_8',
+        'What’s up?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xef\xbb\xbfWhat\xe2\x80\x99s up?',
+        'utf_8_sig',
+        'What’s up?'
+    ),
+
+    # A single non-ASCII character
+    PreEncodedTestStringInfo(
+        b'\x00\x00\xfe\xff\x00\x01\xf4;',
+        'utf_32',  # BE with BOM
+        '🐻'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xff\xfe\x00\x00;\xf4\x01\x00',
+        'utf_32',  # LE with BOM
+        '🐻'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xfe\xff\xd8=\xdc;',
+        'utf_16',  # BE with BOM
+        '🐻'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xff\xfe=\xd8;\xdc',
+        'utf_16',  # LE with BOM
+        '🐻'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xef\xbb\xbf\xf0\x9f\x90\xbb',
+        'utf_8_sig',
+        '🐻'
+    ),
+
+    # A string that starts with a non-ASCII character
+    PreEncodedTestStringInfo(
+        b'\x00\x00\xfe\xff\x00\x00\x00\xc7\x00\x00\x00a\x00\x00\x00 \x00\x00\x00v\x00\x00\x00a\x00\x00\x00?',  # noqa: E501
+        'utf_32',  # BE with BOM
+        'Ça va?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xff\xfe\x00\x00\xc7\x00\x00\x00a\x00\x00\x00 \x00\x00\x00v\x00\x00\x00a\x00\x00\x00?\x00\x00\x00',  # noqa: E501
+        'utf_32',  # LE with BOM
+        'Ça va?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xfe\xff\x00\xc7\x00a\x00 \x00v\x00a\x00?',
+        'utf_16',  # BE with BOM
+        'Ça va?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xff\xfe\xc7\x00a\x00 \x00v\x00a\x00?\x00',
+        'utf_16',  # LE with BOM
+        'Ça va?'
+    ),
+    PreEncodedTestStringInfo(
+        b'\xef\xbb\xbf\xc3\x87a va?',
+        'utf_8_sig',
+        'Ça va?'
+    )
+)
+TEST_STRINGS_TO_ENCODE_AT_RUNTIME = (
+    "",
+    "y",
+    "yaml",
+    "🇾⁠🇦⁠🇲⁠🇱⁠❗"
+)
+
+
+def setUpModule():
+    register_test_codecs()
+    try:
+        del os.environ['YAMLLINT_FILE_ENCODING']
+    except KeyError:
+        pass
+
+
+tearDownModule = unregister_test_codecs
+
+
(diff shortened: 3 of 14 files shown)


Jayman2000:
I just noticed that one of the checks for this PR is failing. [The coverage for `yamllint/config.py` went down](https://coveralls.io/builds/65586469), but that’s just because the total number relevant lines went down. [There’s only two lines that aren’t covered](https://coveralls.io/builds/65586469/source?filename=yamllint%2Fconfig.py#L97), but [those same two lines aren’t covered in the master branch](https://coveralls.io/builds/65531597/source?filename=yamllint%2Fconfig.py#L97). Is there anything that I need to do here?

adrienverge:
> Is there anything that I need to do here?

At the moment, no. I'm sorry, please excuse the delay, this is a big change with much impact, I need a large time slot to review this, which I couldn't find yet.

tests/common.py
```diff
@@ -13,20 +13,161 @@
...
+import collections
 import contextlib
 from io import StringIO
 import os
 import shutil
 import sys
 import tempfile
 import unittest
+import warnings
+from codecs import CodecInfo as CI
```
adrienverge, line 26: When possible I prefer keeping original names, to avoid confusion (especially when "CI" has other meanings). Would you be OK to use `CodecInfo` at the 4 places where it's used?

tests/common.py
```diff
@@ -13,20 +13,161 @@
...
+    for chunk_size in ('32', '16'):
+        for endianness in ('be', 'le'):
+            for sig in ('', '_sig'):
+                yield f'utf_{chunk_size}_{endianness}{sig}'
+    yield 'utf_8_sig'
+    yield 'utf_8'
+
+
+# Workspace related stuff:
+Blob = collections.namedtuple('Blob', ('text', 'encoding'))
```
adrienverge, line 123: It's smart but not very common, maybe hard to understand for beginners. We could also use something more explicit like:
```python
class Blob:
    def __init__(self, text, encoding):
        self.text = text
        self.encoding = encoding
```
This is just an idea, please choose what you prefer.

tests/common.py
```diff
@@ -13,20 +13,161 @@
...
+
+    try:
+        os.chdir(wd)
+        yield
+    finally:
+        os.chdir(backup_wd)
+        shutil.rmtree(wd)
+
+
+def ws_with_files_in_many_codecs(path_template, text):
```
adrienverge, line 163: For the sake of explicitness and readability, can you change it to:
```suggestion
def temp_workspace_with_files_in_many_codecs(path_template, text):
```
or:
```suggestion
def workspace_with_files_in_many_codecs(path_template, text):
```
(if you choose the second one, we should also remove `temp_` from the other function `temp_workspace()`)

tests/common.py
```diff
@@ -13,20 +13,161 @@
...
+    return uses_bom(codec) or (len(string) > 0 and string[0].isascii())
+
+
+def utf_codecs():
+    for chunk_size in ('32', '16'):
+        for endianness in ('be', 'le'):
+            for sig in ('', '_sig'):
+                yield f'utf_{chunk_size}_{endianness}{sig}'
+    yield 'utf_8_sig'
+    yield 'utf_8'
```
adrienverge, line 119: Smarter than a simple hardcoded list :smile: but maybe a list would be more readable?

tests/test_decoder.py
```diff
@@ -0,0 +1,226 @@
...
+        self.assertTrue(encoding_detectable('t ', 'utf_32_le_sig'))
+
+        self.assertTrue(encoding_detectable('yo', 'utf_16'))
+        self.assertTrue(encoding_detectable('u ', 'utf_16_be_sig'))
+        self.assertTrue(encoding_detectable('do', 'utf_16_le_sig'))
+
+        self.assertTrue(encoding_detectable('wn', 'utf_8_sig'))
+
+
+class DecoderTestCase(unittest.TestCase):
```
adrienverge, line 163: In complement to testing various encoded `test_strings`, could you add a bunch of byte arrays (e.g. `b'\xc3\xa7a va ?'`, `b'\xef\xbb\xbf\xc3\xa7a va ?'`, `b'\xe7\x00a\x00 \x00v\x00a\x00 \x00?\x00'` etc.) hardcoded inside our test files?

yamllint/decoder.py
```diff
@@ -0,0 +1,60 @@
...
+        return 'utf_16'
+    elif stream_data[1:2] == b'\x00':
+        return 'utf_16_le'
+    elif stream_data.startswith(codecs.BOM_UTF8):
+        return 'utf_8_sig'
+    else:
+        return 'utf_8'
+
+
+def auto_decode(stream_data, errors='strict'):
```
adrienverge, line 56: The argument `errors='strict'` isn't used anywhere, except in tests. Can we get rid of it?

adrienverge (commented):
Hello Jason, please excuse the very long delay for reviewing this... This was a big piece and I needed time. I apologize.

The 6 commits are well splitted, well explained, and make the review much easier. Thanks a lot!

In my opinion this PR is good to go. I suspect it can solve problems in several cases (including the issues you pointed out), but I also see a small risk of breakage on exotic systems the day it's released. If this happens, will you be around to help find a solution?

A few notes:
- I notice that you used encoding names with underscores (e.g. `utf_8` vs. `utf-8`). I just read on https://docs.python.org/fr/3/library/codecs.html#standard-encodings that not only are they valid, but they also seem to be the "right" notation:
  > Notice that spelling alternatives that only differ in case or use a hyphen instead of an underscore are also valid aliases; therefore, e.g. 'utf-8' is a valid alias for the 'utf_8' codec.
- I feared that using `open().decode()` would put the whole contents of files in memory before even linting them, and affect performance. But this is already [what yamllint does](https://github.com/adrienverge/yamllint/blob/95e17b33dc147c9c35742853e17bdf3fd508550b/yamllint/linter.py#L231) currently.

Jayman2000:
> Hello Jason, please excuse the very long delay for reviewing this... This was a big piece and I needed time. I apologize.

You shouldn’t be apologizing, I should be thanking you.

Thank you for taking the time to review this PR and to write valuable review comments. Sometimes, maintainers don’t take the time to thoroughly review my contributions. When that happens, my contributions end up being rejected without being understood which is frustrating and saddening. You’re review comments are different, though. They clearly demonstrate that you took the time to read, understand and think about this PR. I’d much rather receive a thoughtful review than a timely one.

> In my opinion this PR is good to go. I suspect it can solve problems in several cases (including the issues you pointed out), but I also see a small risk of breakage on exotic systems the day it's released. If this happens, will you be around to help find a solution?

Sure! I had thought about adding a `--force-encoding` option that would disable encoding autodetection. The idea is that someone could use `--force-encoding shift_jis` to make yamllint decode everything using Shift JIS or `--force-encoding cp1252` to make yamllint decode everything using code page 1252. I decided against adding the `--force-encoding` in this PR because this PR was already so big.

Now that you mention the small risk of breakage on exotic systems, it’s making me think about the `--force-encoding` idea again. It might be wise to implement `--force-encoding` after this PR gets merged and before there’s a new stable release of yamllint. That way, we can give users an easy workaround if encoding autodetection breaks something. If you think that that is a good idea, then I can open another PR that adds a `--force-encoding` option after this PR gets merged.

---

I just pushed a new version of this pull request with the following changes:

- I rebased it on 8513d9b97da3b32453b3fccb221f4ab134a028d7 (the tip of `master`, at the moment).
- I implemented the review suggestions (see the soon-to-be-resolved conversations).
- The variable in `tests/common.py` that used to be named `test_strings` is now called `TEST_STRINGS_TO_ENCODE_AT_RUNTIME`. I made this change for two reasons:
    1. I needed a add a new variable that contained a different type of test strings. I needed to add that variable in order to implement one of the suggestions from your review comments.
    2. The variable name should have been all uppercase anyway because the variable is a constant.
- For similar reasons, I also renamed the `error1` and `error2` variables  to `ERROR1` and `ERROR2` in the function that used to be named `test_detect_encoding()`.
- I added a `repr()` call to the `msg="auto_decode({repr(input_bytes)}) returned the wrong value."` line in `tests/test_decoder.py`. The `repr()` call helps ensure that the error message has proper syntax.
- I fixed a dead link in one of the commit messages (<https://sourceware.org/glibc/manual/html_node/Extended-Char-Intro.html> → <https://www.gnu.org/software/libc/manual/html_node/Extended-Char-Intro.html>).
- I revised the commit message for “decoder: Autodetect detect encoding of YAML files” to (hopefully) make it easier to understand for people who are unfamiliar with this issue.
- I made the commit message for “decoder: Autodetect encoding for ignore-from-file” slightly simpler.
- I improved the commit message for “CI: Fail when open()’s default encoding is used”. I reworded part of it in order to make it more clear why coverage.py is affected by default encoding errors.
- Updated copyright years for some files to account for the fact that I worked on some files in both 2023 and 2024.

tests/common.py
Jayman2000: Done. I only did the `as CI` thing in order to prevent certain lines from being too long. The version of this PR that I just pushed uses `# noqa: E501` instead.

tests/common.py
Jayman2000: I changed it to use `class` syntax instead of `namedtuple()`. I only really use `namedtuple()` out of habit. I don’t really know which option is better, but I agree that people are more likely to be familiar with the class syntax.

tests/common.py
Jayman2000: OK. I renamed it to `temp_workspace_with_files_in_many_codecs()`.

tests/common.py
Jayman2000: Yeah, that function is overkill. I changed it to a hard-coded tuple. That function is 7 lines long. The new hard-coded tuple is 12 lines long. That level of complication isn’t worth it in order to save 6 lines.

tests/test_decoder.py
Jayman2000: Done.

yamllint/decoder.py
Jayman2000: Done. I don’t really remember why I included that argument in the first place.

tests/common.py
```diff
@@ -13,20 +13,169 @@
...
+        len(obj)
+    )
+
+
+test_codec_infos = {
+    'utf_32_be_sig': CodecInfo(encode_utf_32_be_sig, codecs.getdecoder('utf_32')),  # noqa: E501
+    'utf_32_le_sig': CodecInfo(encode_utf_32_le_sig, codecs.getdecoder('utf_32')),  # noqa: E501
+    'utf_16_be_sig': CodecInfo(encode_utf_16_be_sig, codecs.getdecoder('utf_16')),  # noqa: E501
+    'utf_16_le_sig': CodecInfo(encode_utf_16_le_sig, codecs.getdecoder('utf_16')),  # noqa: E501
+}
```
adrienverge, line 81: I understand these `# noqa` originate from https://github.com/adrienverge/yamllint/pull/630#discussion_r1792173070, but instead of overriding the linter, can you use this:
```python
test_codec_infos = {
    'utf_32_be_sig':
    CodecInfo(encode_utf_32_be_sig, codecs.getdecoder('utf_32')),
```
or this:
```python
test_codec_infos = {
    'utf_32_be_sig': CodecInfo(encode_utf_32_be_sig,
                               codecs.getdecoder('utf_32')),
```
?

tests/test_decoder.py
```diff
@@ -0,0 +1,482 @@
...
+        self.input_bytes = input_bytes
+        self.codec_for_input_bytes = codec_for_input_bytes
+        self.expected_output_str = expected_output_str
+
+
+PRE_ENCODED_TEST_STRING_INFOS = (
+    # An empty string
+    PreEncodedTestStringInfo(
+        b'',
+        None,
```
adrienverge, line 51: Good idea.

tests/test_decoder.py
adrienverge: That's great!
A single ASCII character, a string that starts with an ASCII character, a single non-ASCII character, a string that starts with a non-ASCII character: I don't see what else we could test.

tests/test_decoder.py
```diff
@@ -0,0 +1,482 @@
...
+        self.assertTrue(encoding_detectable('wn', 'utf_8_sig'))
+
+
+class DecoderTestCase(unittest.TestCase):
+    def detect_encoding_test_helper(
+        self,
+        original_string,
+        input_bytes,
+        expected_output
+    ):
```
adrienverge, line 348: These variable names confused me for 2 minutes (I thought "output" referred to a string) :sweat_smile: 
Can you get rid of `original_string` (it isn't used) and use something like this:
```suggestion
    def detect_encoding_test_helper(self, input_bytes, expected_codec):
```
… and 3 lines later `actual_output` → `actual_codec`?

PS: For consistency, we could also change `auto_decode_test_helper()` to use `expected_string` instead of `expected_output`. It would be more explicit and easier to grasp.

tests/test_decoder.py
```diff
@@ -0,0 +1,482 @@
...
+                actual_output,
+                msg=ERROR.format(repr(input_bytes))
+            )
+            self.assertIsInstance(actual_output, str)
+        else:
+            try:
+                decoder.auto_decode(input_bytes)
+            except UnicodeDecodeError as exception:
+                return exception
+        return None
```
adrienverge, line 420: Returning `None` is implicit and not stricly needed here. As you prefer.

tests/test_decoder.py
```diff
@@ -0,0 +1,482 @@
...
+                pre_encoded_test_string_info.codec_for_input_bytes,
+                pre_encoded_test_string_info.expected_output_str
+            )
+            if exception is not None:
+                new_exception = self.failureException(
+                    msg=ERROR.format(
+                        repr(pre_encoded_test_string_info.input_bytes)
+                    )
+                )
+                raise new_exception from exception
```
adrienverge, line 436: I understand the logic and I'm OK to keep this code, but if it was me I would simplify all this and just let the exception pop up from `auto_decode_test_helper()`. It's a less clean, but shorter :wink:

(same later)

tests/test_decoder.py
```diff
@@ -0,0 +1,482 @@
...
+                    string
+                )
+                if exception is not None:
+                    at_least_one_decode_error = True
+        self.assertTrue(
+            at_least_one_decode_error,
+            msg=(
+                "None of the TEST_STRINGS_TO_ENCODE_AT_RUNTIME triggered a "
+                + "decoding error."
+            )
```
adrienverge, line 454: The `+` is not needed:

```suggestion
            msg=("None of the TEST_STRINGS_TO_ENCODE_AT_RUNTIME triggered a "
                 "decoding error.")
```

Or if you really like Black :wink::
```suggestion
            msg=(
                "None of the TEST_STRINGS_TO_ENCODE_AT_RUNTIME triggered a "
                "decoding error."
            )
```

tests/common.py
```diff
@@ -13,20 +13,169 @@
...
+    'utf_16_be',
+    'utf_16_be_sig',
+    'utf_16_le',
+    'utf_16_le_sig',
+    'utf_8',
+    'utf_8_sig'
+)
+
+
+def encode_utf_32_be_sig(obj, errors='strict'):
```
adrienverge, line 48: It looks like `errors='strict'` is already the default: https://docs.python.org/3/library/codecs.html#codecs.encode and no yamllint code uses this argument anywhere.
But I'm OK to keep it if you think it's better to be explicit.

tests/test_decoder.py
```diff
@@ -0,0 +1,482 @@
...
+    temp_workspace_with_files_in_many_codecs,
+    test_codec_built_in_equivalent,
+    unregister_test_codecs,
+    uses_bom,
+)
+
+from yamllint import decoder
+
+
+class PreEncodedTestStringInfo():
```
adrienverge, line 35:
```suggestion
class PreEncodedTestStringInfo:
```

adrienverge (commented):
Thanks for following up, again it's a great work.
Amended commits are still perfectly clear and anticipate some of my questions :)

> Sure! I had thought about adding a `--force-encoding` option that would disable encoding autodetection. The idea is that someone could use `--force-encoding shift_jis` to make yamllint decode everything using Shift JIS or `--force-encoding cp1252` to make yamllint decode everything using code page 1252. I decided against adding the `--force-encoding` in this PR because this PR was already so big.
>
> Now that you mention the small risk of breakage on exotic systems, it’s making me think about the `--force-encoding` idea again. It might be wise to implement `--force-encoding` after this PR gets merged and before there’s a new stable release of yamllint. That way, we can give users an easy workaround if encoding autodetection breaks something. If you think that that is a good idea, then I can open another PR that adds a `--force-encoding` option after this PR gets merged.

If I understand correctly, YAML specification says that files should be Unicode-encoded, but in the real world not all YAML files are.
But I'm not sure whether today, with yamllint 1.35.1, it's possible to successfully run yamllint on non-Unicode-encoded YAML files. Do you know whether it's the case?
Here is what I tried on my system (a recent GNU/Linux):
- Python's `sys.getdefaultencoding()` always equals `utf-8`, even when setting environment variables `PYTHONIOENCODING=cp1252`, `PYTHONUTF8=0` and `LANG=fr_FR.CP1252`.
- After creating a CP1252-encoded file (`python -c 'with open("/tmp/cp1252.yaml", "wb") as f: f.write("- éçà".encode("cp1252"))'`), running `PYTHONIOENCODING=cp1252 yamllint /tmp/cp1252.yaml` fails with `UnicodeDecodeError: 'utf-8' codec can't decode …`.

I'm not an expert of how Python handles encoding, so your input is more than welcome on this :slightly_smiling_face:
- *If it's not the case*, then no need for an extra option `--force-encoding`.
- *If it's the case*, then I agree it would be better to add a way to avoid breakage for those users. However I'm not a fan of a new command line option (`--force-encoding`), because ideally this would be temporary and removed after a few yamllint versions (after users fixed their files to use UTF-8), so it would break usage the day when we remove this option.
  Instead, we could use an environment variable `YAMLLINT_IO_ENCODING`, that `detect_encoding()` would use to override the encoding if defined (then we could unit-test 2 or 3 cases in a new `test_detect_encoding_with_env_var_override()`). Removing support for this env var in the future won't break command-line options. In my opinion, this should be fairly simple and should go in the same commit.

> I revised the commit message for “decoder: Autodetect detect encoding of YAML files”

I like the new version too, it's clearer about the Windows case. Did you mean `Autodetect detect` → `Autodetect`?

> I revised the commit message for “decoder: Autodetect detect encoding of YAML files” to (hopefully) make it easier to understand for people who are unfamiliar with this issue.

Indeed, it's slightly clearer :+1: and will allow grepping `warn_default_encoding` or `EncodingWarning` inside Git history.

dalito:
> After creating a CP1252-encoded file (python -c 'with open("/tmp/cp1252.yaml", "wb") as f: f.write("- éçà".encode("cp1252"))'), running PYTHONIOENCODING=cp1252 yamllint /tmp/cp1252.yaml fails with UnicodeDecodeError: 'utf-8' codec can't decode ….

I believe you are expecting something wrong here: From the [docs](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONIOENCODING): _PYTHONIOENCODING: If this is set before running the interpreter, it overrides the encoding used for stdin/stdout/stderr,_ 
So setting "PYTHONIOENCODING" has not effect on the encoding used when opening files which uses what is returned by locale.getencoding which is still utf-8.

(I did not check if you do/did this.) It could be interesting to run tests with https://docs.python.org/3/library/io.html#io-encoding-warning to detect where the code relies on default encoding.

tests/common.py
Jayman2000: Done.

tests/test_decoder.py
Jayman2000: > Can you get rid of `original_string` (it isn't used) and use something like this:
> […]
> … and 3 lines later `actual_output` → `actual_codec`?

Done.

> PS: For consistency, we could also change `auto_decode_test_helper()` to use `expected_string` instead of `expected_output`. It would be more explicit and easier to grasp.

Done.

tests/test_decoder.py
Jayman2000: You’re right! (I always forget about the implicit `None` return thing). I just pushed a new version without the `return None` line.

tests/test_decoder.py
Jayman2000: Now that I look at it again, I think that it makes more sense for the `UnicodeDecodeError` to be raised instead of returned. I just pushed a new version that lets it get raised instead of returned.

tests/test_decoder.py
Jayman2000: Done.

tests/common.py
Jayman2000: The documentation made me think that you needed to have both an `obj` parameter and an `errors` parameter, but it looks like you only need to have a `obj` parameter. I just pushed a new version that gets rid of the `errors` parameter. Thank you for this hint!

tests/test_decoder.py
Jayman2000: Done.

Jayman2000:
I just pushed a new version of this PR. Here’s what’s new:

- I rebased this branch on top of e427005b528f5a4cd59387791b6f9d87ae7dd4ce (the tip of `master`, at the moment) which means that there are no more merge conflicts.
- The commit called “tests: Restore stdout and stderr” now called “tests: Restore stdout”. This was done to account for the changes from e427005b528f5a4cd59387791b6f9d87ae7dd4ce.
- The `YAMLLINT_FILE_ENCODING` environment variable can now be used to override the encoding used to decode files (see below).
- I fixed the “Autodetect detect” thing in that one commit message’s subject line.
- I implemented the review suggestions (see the conversations that just recently got resolved).
- For files that already existed, already had a copyright notice and that I significantly contributed to, I added myself to the copyright notice (I added “Copyright (C) &lt;years&gt; Jason Yundt”).
- For the files that I created myself, I updated the copyright years in order to account for the fact that it’s now 2025.

---

@adrienverge

> If I understand correctly, YAML specification says that files should be Unicode-encoded, but in the real world not all YAML files are. But I'm not sure whether today, with yamllint 1.35.1, it's possible to successfully run yamllint on non-Unicode-encoded YAML files. Do you know whether it's the case? Here is what I tried on my system (a recent GNU/Linux):
>
> * Python's `sys.getdefaultencoding()` always equals `utf-8`, even when setting environment variables `PYTHONIOENCODING=cp1252`, `PYTHONUTF8=0` and `LANG=fr_FR.CP1252`.
>
> * After creating a CP1252-encoded file (`python -c 'with open("/tmp/cp1252.yaml", "wb") as f: f.write("- éçà".encode("cp1252"))'`), running `PYTHONIOENCODING=cp1252 yamllint /tmp/cp1252.yaml` fails with `UnicodeDecodeError: 'utf-8' codec can't decode …`.

Checking `sys.getdefaultencoding()` is a good idea, but I actually don’t think that that particular function is relevant here. [Here’s what the Python documentation has to say about `open()` function’s `encoding` argument:](https://docs.python.org/3.13/library/functions.html#open)

> In text mode, if _encoding_ is not specified the encoding used is platform-dependent: [`locale.getencoding()`](https://docs.python.org/3.13/library/locale.html#locale.getencoding) is called to get the current locale encoding.

That being said, I think that the documentation might actually be wrong here. `locale.getencoding()` seems to always return the actual locale encoding, regardless of whether UTF-8 mode is enabled or disabled. I think that the documentation for `open()` should probably say `locale.getpreferredencoding()` instead of `locale.getencoding()`.

When I run `python -c 'import locale; print(locale.getpreferredencoding())'` with [a plain-ASCII locale](https://jasonyundt.website/posts/terminal-in-ascii-on-linux) and UTF-8 mode disabled, it prints `ANSI_X3.4-1968`. When I run that same command with a plain-ASCII locale and UTF-8 mode enabled, it prints `utf-8`. When I run that same command with a UTF-8 locale, it prints `UTF-8`.

I agree with @dalito about `PYTHONIOENCODING`. It’s probably also a red herring.

That being said, I’m not sure why yamllint was still using UTF-8 when you tested it. I can think of a few possibilities, though:

1. [`fr_FR.CP1252` is not on glibc’s supported locales list.](https://sourceware.org/git/?p=glibc.git;a=blob;f=localedata/SUPPORTED) It’s possible that this is a bug with glibc and that it would work better if you tried a locale that’s on that list.
2. It’s possible that there isn’t actually a locale on your system that’s named `fr_FR.CP1252`. You could try running `locale --all-locales` to see if it’s on the list.
3. `fr_FR.CP1252` is just a name. Technically, you could have accidentally created a locale that has `CP1252` in its name, but actually uses UTF-8. You could try running `LANG=fr_FR.CP1252 locale --keyword-name LC_CTYPE` to verify that the `charmap` really is set to `CP1252`.
4. It’s also possible that your version of glibc doesn’t actually support the `CP1252` character encoding. You could try running `locale --charmaps` to check and see if `CP1252` is available on your system.

If you can’t get it to work on your host system, you can always try [this Nix flake](https://codeberg.org/JasonYundt/yamllint-encoding-test) that I created. That flake will automatically create a virtual machine with a GB 18030 locale and then run yamllint in that virtual machine in order to demonstrate that yamllint 1.35.1 can indeed be used to lint GB 18030–encoded YAML files.

---

> * _If it's not the case_, then no need for an extra option `--force-encoding`.
>
> * _If it's the case_, then I agree it would be better to add a way to avoid breakage for those users. However I'm not a fan of a new command line option (`--force-encoding`), because ideally this would be temporary and removed after a few yamllint versions (after users fixed their files to use UTF-8), so it would break usage the day when we remove this option.
>   Instead, we could use an environment variable `YAMLLINT_IO_ENCODING`, that `detect_encoding()` would use to override the encoding if defined (then we could unit-test 2 or 3 cases in a new `test_detect_encoding_with_env_var_override()`). Removing support for this env var in the future won't break command-line options. In my opinion, this should be fairly simple and should go in the same commit.

OK, that sounds like a good idea. I’ve implemented it, but I decided to call the environment variable `YAMLLINT_FILE_ENCODING` instead of `YAMLLINT_IO_ENCODING`. `YAMLLINT_IO_ENCODING` sounds too similar to `PYTHONIOENCODING`, and (as far as I know) the `PYTHONIOENCODING` variable affects stdin, stdout and stderr, not regular files. People who are familiar with `PYTHONIOENCODING` might see the name `YAMLLINT_IO_ENCODING` and assume that it has something to do with stdin, stdout and stderr.

---

> I like the new version too, it's clearer about the Windows case. Did you mean `Autodetect detect` → `Autodetect`?

Yep. Thanks for pointing that out.

---

@dalito

> (I did not check if you do/did this.) It could be interesting to run tests with https://docs.python.org/3/library/io.html#io-encoding-warning to detect where the code relies on default encoding.

I agree, `-X warn_default_encoding` is a really useful way to track down these kinds of bugs. I’m really glad that was added to Python. A lot of the changes in this PR come from me running `python -X warn_default_encoding -W error::EncodingWarning -m unittest discover` and seeing what broke. The last commit in this PR makes the CI tests run with `-X warn_default_encoding -W error::EncodingWarning` to help us avoid these kinds of mistakes in the future.

docs/configuration.rst
```diff
@@ -228,6 +228,21 @@ or:
...
+   set the ``YAMLLINT_FILE_ENCODING`` environment variable to the name of the
+   character encoding that you want yamllint to use for ignore files.
+   Specifically, ``YAMLLINT_FILE_ENCODING`` should be set to `the name of one
+   of Python’s standard character encodings
+   <https://docs.python.org/3/library/codecs.html#standard-encodings>`_. Please
+   note, this should only be used as a temporary solution in order to make it
+   easier to migrate from older versions of yamllint to newer versions of
+   yamllint. See :doc:`Character Encoding Override
+   <character_encoding_override>` for details.
+
```
adrienverge, line 245: Is it worth mentioning here? Why not, you decide!
By the way, the same also applies to the configuration file `.yamllint`.

If you keep it, maybe we can simplify it like this?

```rst
.. note:: Files on the ``ignore-from-file`` list should use either UTF-8, UTF-16
   or UTF-32. See :doc:`Character Encoding Override
   <character_encoding_override>` for details and wordarounds.
```

docs/character_encoding_override.rst
```diff
@@ -0,0 +1,32 @@
+Character Encoding Override
+===========================
+
+When yamllint reads a file, it will try to automatically detect that file’s
```
adrienverge, line 4: Idea:
```suggestion
When yamllint reads a file (either configuration or a file to lint), it will try
to automatically detect that file’s …
```
(this goes along with my next comment ↓)

adrienverge (commented):
@dalito @Jayman2000 thanks for your help and explanations on system encodings! Indeed I misunderstood `PYTHONIOENCODING`.
@Jayman2000, https://jasonyundt.website/posts/terminal-in-ascii-on-linux is a nice and useful article :+1: For information, as you suspected, `fr_FR.CP1252` was not in my glibc locales list. Since `fr_FR.iso88591` is present, next time I'll use this one.

I rereviewed the code and it looks ready for merging. Before that, I have 2 remarks:
1. > OK, that sounds like a good idea. I’ve implemented it, but I decided to call the environment variable YAMLLINT_FILE_ENCODING instead of YAMLLINT_IO_ENCODING

   Your argument makes sense, `YAMLLINT_FILE_ENCODING` sounds better to avoid confusion. But yamllint can also lint stdin (`echo -e 'é: v\né: v' | yamllint -`) and write reports to stdout (`duplication of key "é" in mapping`). Does your latest implementation with `YAMLLINT_FILE_ENCODING` also handle the encoding of stdin/stdout?

   By the way, good idea to properly warn about the temporariness of `YAMLLINT_FILE_ENCODING`.

2. The new documentation page on "Character Encoding Override" is a great addition. Thanks! To prepare for potential future additions, may I suggest to rename it to "Character Encoding", and put the last 2 paragraphs in a section "Override character encoding"?

docs/character_encoding_override.rst
Jayman2000: I like that suggestion. I incorporated your suggestion into the version of the pull request that I just pushed, but I used slightly different wording (I wanted to make the sentence slightly easier to understand and sound more natural). What do you think about the wording that I used?

docs/configuration.rst
Jayman2000: I like the shorter version that you proposed better. I just pushed a new version of this pull request that shortens it the way that you suggested.

docs/character_encoding_override.rst
adrienverge: It's perfect, thanks! (I'm not a native English speaker so such enhancements are always welcome :slightly_smiling_face:)

adrienverge (approved):
Hello Jason,

> I don’t think that we need to make sure that we use the right character encoding for `stdout`. I think that Python will choose the right character encoding for us. I took a look at each of yamllint’s output formats to try and figure out what the ideal character encoding would be:

Agreed. In details, I agree with your reasoning about the 4 output `--format`s for stdout.

About stderr, it seems logical to do the same as for stdout (here, let Python use the locale encoding).

The commits are still well split and give all the context needed. Thank you for that.

Let's release this!

PR #637, opened 2024-01-08
build: enable refurb rules in ruff

DimitriPapadopoulos:


adrienverge:
I'm not of than of these rules. Sometimes it helps readability to be able to write what refurb dislike.

DimitriPapadopoulos:
Agreed. Some of the rules do improve code and performance, other rules may result in less readable code, such as:
```
if ...:
    return ....
return ....
```
instead of:
```
if ...:
    return ....
else:
    return ....
```

PR #639, opened 2024-01-10
build: requires setuptools-scm

DimitriPapadopoulos:


adrienverge:
About the first commit `build: format requires as in setuptools doc`, the [mentioned documentation](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html) shows various examples, including a majority of "one space on each side":

    BazSpam == 1.1
    pywin32 >= 1.0

Moreover the "one space on each side" style is consistent with the `dependencies` section. So I propose to drop this change.

About the second commit `build: requires setuptools-scm`, I'm pretty sure you are right to add it, but I (and potential reviewers of this PR) don't have the reason. Can you document why inside the commit message?
(I see a note about `setuptools-scm` on https://setuptools.pypa.io/en/latest/userguide/dependency_management.html, but it's not explicit why it would apply to yamllint.)

DimitriPapadopoulos:
I'll drop the first commit.

As for the second commit, I thought setuptools_scm is used by yamllint, but it is not.  Maybe an effect of my mixing versions of pip/python?

PR #641, opened 2024-01-13
build: enable flake8-blind-except in ruff

DimitriPapadopoulos:


adrienverge (commented):
I'm -1 for this one.
Changing these would satisfy a linter, but `YamllintConfig.extend()` can raise something else than an `AssertionError` (moreover we may change in the future). With `except Exception as e`, any present or future problem will continue to be caught.

DimitriPapadopoulos:
Agreed. The idea is that you must know and document the exceptions raised by a piece code. While this might nice be in simple cases and deep in function call trees, it is not always feasible higher in a function call tree, especially when like here you **do** want to catch any exception.

PR #642, opened 2024-01-13
build: enable ruff-specific rules in ruff

DimitriPapadopoulos:


tests/test_cli.py
```diff
@@ -70,7 +70,7 @@ def setUpModule():
     env_vars_that_could_interfere = (
         'YAMLLINT_CONFIG_FILE',
         'XDG_CONFIG_HOME',
-        # These variables are used to determine where the user’s home
```
adrienverge, line 73: I thought `’` was a valid, and even "more correct":
- it's the character used in Python documentation: https://docs.python.org/3/library/os.path.html#os.path.expanduser
- it's what a renowned English dictionary uses: https://dictionary.cambridge.org/grammar/british-grammar/apostrophe

adrienverge (commented):
Review would be easier with a small explanation in the commit message :slightly_smiling_face: For example it took me 2 minutes to find https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf and understand what this change does. (This remark applies to all pull requests.)

Although the `Unused noqa directive` is cool and useful, I don't like the others (see my inline comment). They look like over-linting. I propose to only add `RUF100`.

tests/test_cli.py
DimitriPapadopoulos: I did add an explanation in the commit message, perhaps it is not clear enough:
> ```
> RUF003 Comment contains ambiguous `’` (RIGHT SINGLE QUOTATION MARK). Did you mean ``` (GRAVE ACCENT)?
> ```
> Doesn't make much sense, however let's be consistent and use `'` apostrophes, as in all other code comments.

Both are somehow correct. I decided to address this specific occurrence of the warning, which I agree might not make sense in other contexts, because:
- you use `'` everywhere else, `’` is used only for this specific apostrophe,
- the Python documentation does **not** use `’` in the source of the documentation, `’` is changed into `’` when rendering the documentation to HTML and other formats:
  
  [**python/cpython/Doc/library/os.path.rst**](https://github.com/python/cpython/blob/e6495159f6f2ca4fe3b5143c62fc2ad5873ef54f/Doc/library/os.path.rst?plain=1#L157-L160)
  Lines 157 to 160 in [e649515](https://github.com/python/cpython/commit/e6495159f6f2ca4fe3b5143c62fc2ad5873ef54f)
  ```rst
  .. function:: expanduser(path)
  
     On Unix and Windows, return the argument with an initial component of ``~`` or
     ``~user`` replaced by that *user*'s home directory.
  ```

adrienverge: > I did add an explanation in the commit message, perhaps it is not clear enough

Sorry, my bad, I missed it! (I'm used to commit messages copied-pasted in the pull request message, but I shouldn't rely on that.)

Both apostrophes types are correct, but `’` is "more" correct, typographically speaking. Some years ago using such Unicode characters could cause problems, but nowadays there are no reasons not to use the "right" characters (and emojis, if we wanted to...)
The fact that we use another "less right" apostrophe `'` at many other places is not perfectly satisfying, I agree, but it's not a good reason to suppress the correct one.

> * the Python documentation does **not** use `’` in the source of the documentation, `'` is changed into `’` when rendering the documentation

I didn't know, that's nice :+1:
About yamllint, comments in Python code are not meant to be post-processed, so we can use `’` right away.

OK for you to keep `’`?

adrienverge:
> Removed all `RUF` rules but `RUF100`.

Thanks!

Issue #587, opened 2023-08-15
Consider automatically changing the settings based on yaml version


SubaruArai:
Yaml's spec is quite complicated, and has evolved over time.

There are also some confusion around what version yamllint supports (#101), and default config between different yaml versions (#559, #540, #430, #344, #232, #247, #158).

May I suggest to add:

1. A config key to specify the targeted yaml version
   e.g. `yaml-version: 1.1`
2. Automatically change the default rules based on the yaml version
   e.g. if ver>=1.2, `truthy.check-keys = true`
3. (Optionally) add an option to automatically change rules based on the individual file's directive
   e.g. if `%YAML 1.2`, change config to 1.2.2

1 and 2 should be relatively easy, but 3 needs some consideration on how to implement.

Information: [yaml spec changelog](https://yaml.org/spec/1.2.2/ext/changes)

adrienverge:
Hello,

Although it's not 100% accurate, yamllint aims to support the latest version of the YAML specification (1.2.2). I believe that supporting multiple versions based on tags would be complicated, and cumbersome to most users.

PS: I don't see how the issues you pointed out (#559, #540, #430, #344, #232, #247, #158) are related to the YAML spec version.

SubaruArai:
> I believe that supporting multiple versions based on tags would be complicated, and cumbersome to most users.

Fair enough.

> I don't see how the issues you pointed out (#559, #540, #430, #344, #232, #247, #158) are related to the YAML spec version.

I thought `truthy`'s intend was to detect potential error between yaml 1.1/1.2 parsers.
The example `{y: 1, yes: 2, on: 3, true: 4, True: 5} -> {y: 1, true: 5}` will only happen in yaml<=1.1 parsers.

adrienverge:
> I thought `truthy`'s intend was to detect potential error between yaml 1.1/1.2 parsers.
> The example `{y: 1, yes: 2, on: 3, true: 4, True: 5} -> {y: 1, true: 5}` will only happen in yaml<=1.1 parsers.

@SubaruArai you're right. I've opened https://github.com/adrienverge/yamllint/pull/650 to implement your 3rd suggestion (and while I disagree with the 2nd one, I like the 1st :+1:).

Issue #559, opened 2023-04-04
Consider changing `truthy` linter's `check-keys` default


gravieure:
The default configuration for the `truthy` linter [applies it to keys as well as values](https://github.com/adrienverge/yamllint/blob/6bfd6756e2414e0a16a256eaf6575e922d7e2ac9/yamllint/rules/truthy.py#L43).  The syntax for [GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) definitions uses an [`on`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on) key to determine what event the workflow applies to.

The upshot is that, out of the box, yamllint will blow up on any YAML document configuring a GitHub Action.  This certainly can be fixed by disabling `check-keys`, but the presence of that option is non-obvious.

Given that GitHub Actions are a very common YAML usecase, do you think it'd be reasonable to disable this linter's `check-keys` option by default?

adrienverge:
Hello, is it a duplicate of #540, #430, #344, #232, #247, #158 and others?

gravieure:
No; those all assume that the behavior is a bug needing to be fixed.  I understand that it is not a bug, but an option which is enabled by default.  My question is: would it make sense to change that default?  The numerous issues filed on this subject are strong evidence that the default is confusing.

If changing the default is disagreeable, what do you think about the `truthy` linter printing some info about the `check-keys` option in this scenario?  I think that would also reduce confusion and duplicate bug reports, since people would be guided to the right option without needing to file an issue about it.

SubaruArai:
I was thinking to suggest using a `%YAML 1.2` tag to automatically disable `check-keys`.
But then again, found out that `pyyaml` (still!) doesn't support yaml 1.2.

Switching to `ruamel.yaml` might be an option, but that's for another discussion.

777arc:
I think this is very reasonable, either disable check-keys or disable it for files inside of .github/workflows

adrienverge:
TL;DR: For those seeking a quick-win to silent the error on GitHub Actions, here are 4 solutions:
\- use [config option](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy) `rules: {truthy: {check-keys: false}}`
\- use [config option](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy) `rules: {truthy: {allowed-keys: [on, …]}}`
[...]

Issue #540, opened 2023-02-24
Truthy rule gives confusing result for common Github Actions yml that uses "on" syntax


AndrewLaneAtPowerSchool:
The truthy rule seems to see the "on" keyword in the following YAML and think it should be true/false (assuming because on/off can be seen as truthy and falsy?).

Example YML:
```
name: Run tests

on: [push, pull_request]

jobs:  
  test:
    runs-on: ubuntu-latest

    steps:
    - name: …
```

Snippet of yamllint output:

```
  3:1       warning  truthy value should be one of [false, true]  (truthy)
``

Expected: should not throw warning

andrewimeson:
This is not a false positive. This is a documented and expected behavior. Your options are:

1. Quote the key

   ```yaml
   ---
   name: Run tests

   "on": [push, pull_request]
   ```

2. Disable the rule with a line comment

   ```yaml
   ---
   name: Run tests

   on: [push, pull_request]  # yamllint disable-line rule:truthy
   ```

3. Disable key checking in the yamllint configuration (e.g. `.yamllint.yaml`)

   ```yaml
   ---
   extends: default
   rules:
     truthy:
       check-keys: false
   ```


This is a duplicate of #430 and others

AndrewLaneAtPowerSchool:
Thanks!

Issue #344, opened 2020-11-26
Allow configuration of truthy keys


mondeja:
Linting Github Actions configuration files I'm getting the warning `truthy value should be one of [false, true]  (truthy)` with [`on` keys](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#on). For discussion, is not this default behaviour too much intrusive?

The solution would be set `check-keys: false` in `truthy` configuration, but this removes the verification for all keys. Maybe an `allowed-keys` configuration can help in this case?

Ignore solution if someone needs it:

```yaml
rules:
  truthy:
    allowed-values:
      - 'true'
      - 'false'
    check-keys: true
    ignore: |
      .github/workflows
```

adrienverge:
Hello,

> The solution would be set `check-keys: false` in `truthy` configuration, but this removes the verification for all keys.

What do you mean exactly with "but this removes the verification for all keys"? Can you give an example of what values you would put in this `allowed-keys` list?

mondeja:
> What do you mean exactly with "but this removes the verification for all keys"?

If I've another key, for example `On` in the file would not be checked, not?

- Can you give an example of what values you would put in this `allowed-keys` list?

```yaml
allowed-keys:
  - on
```

So with this `On` would be treated as warning but not `on`.

adrienverge:
> If I've another key, for example `On` in the file would not be checked, not?

Sure, but why would you want to detect and report `On` (and other possibly truthy values), if you're not interested by reporting `on`? Why would you want to detect and report `On`, but not `Hey` or `Álvaro`?

If I understand your use-case, you either want to:
- use the rule `truthy` with `{check-keys: false}` (keys can be whatever they want),
- or have a new rule (not `truthy`) that would check keys and forbid them if they match a list.

mondeja:
Sorry, but I don't understand your point. What is the solution that you propose for Github Actions configuration files linting? This next?

```yaml
rules:
  truthy:
    allowed-values:
      - 'true'
      - 'false'
    check-keys: true
    ignore: |
      .github
```

Ignore the file?

adrienverge:
If you want to use `on` as a key in your YAML file, you need to disable the `truthy` rule or use `check-keys: false`. From what you post, it seems like you don't need `check-keys: true`.

mondeja:
Wouldn't it be helpful if you could define the keys for which the rule works?

adrienverge:
[...]

PR #650, opened 2024-01-31
truthy: Adapt forbidden values based on YAML version

adrienverge:
Specification of YAML ≤ 1.1 has 22 boolean values:

    y     | Y            | n     | N
    yes   | Yes   | YES  | no    | No    | NO
    true  | True  | TRUE | false | False | FALSE
    on    | On    | ON   | off   | Off   | OFF

Whereas YAML 1.2 spec recognizes only 6 [^1]:

    true  | True  | TRUE | false | False | FALSE

For documents that explicit state their YAML spec version at the top of the document, let's adapt the list of forbidden values.

In the future, we should:
- implement a configuration option to declare the default YAML spec version, e.g. `default-yaml-spec-version: 1.2`,
- consider making 1.2 the default in a future release (this would be a slight breaking change, but yamllint always tried to be 1.2-compatible).
- consider adapting yamllint to other 1.1 vs. 1.2 differences [^2].

Solves: https://github.com/adrienverge/yamllint/issues/587

Related to: #559 #540 #430 #344 #247 #232 #158

[^1]: https://yaml.org/spec/1.2.2/#1032-tag-resolution
[^2]: https://yaml.org/spec/1.2.2/ext/changes/#changes-in-version-12-revision-120-2009-07-21

### DIFF
--- a/tests/rules/test_truthy.py
+++ b/tests/rules/test_truthy.py
@@ -27,7 +27,8 @@ class TruthyTestCase(RuleTestCase):
                    'True: 1\n', conf)
 
     def test_enabled(self):
-        conf = 'truthy: enable\n'
+        conf = ('truthy: enable\n'
+                'document-start: disable\n')
         self.check('---\n'
                    '1: True\n'
                    'True: 1\n',
@@ -35,7 +36,8 @@ class TruthyTestCase(RuleTestCase):
         self.check('---\n'
                    '1: "True"\n'
                    '"True": 1\n', conf)
-        self.check('---\n'
+        self.check('%YAML 1.1\n'
+                   '---\n'
                    '[\n'
                    '  true, false,\n'
                    '  "false", "FALSE",\n'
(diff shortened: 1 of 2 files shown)


adrienverge:
@perlpunk today I attended a great talk in Brussels which said there were 22 values (not 18), so I updated the commit message:

```diff
-Specification of YAML ≤ 1.1 has 18 boolean values:
+Specification of YAML ≤ 1.1 has 22 boolean values:
```

Notes:
\- I haven't adapted the yamllint rule code, because PyYAML treats the 4 extra values `y`, `Y`, `n` and `N` as strings (so it's neither 1.1- nor 1.2-complete).
\- I had a hard time finding [the exact list](https://yaml.org/type/bool.html) on yaml.org (it's not directly on the [YAML 1.1 specification page](https://yaml.org/spec/1.1/current.html)).

fosdem2024

By the way, I read that between 1.1 and 1.2 "*the merge `<<` and value `=` special mapping keys have been removed*" but I heard that some 1.2-compatible YAML parsing libs implement the merge marker `<<` as an option. Do you think yamllint should complain when it finds both a tag `%YAML 1.2` and a merge marker `<<`?

perlpunk:
@adrienverge thanks very much!
Also, nice to meet you :)

Yeah, I think the non-standard behaviour of PyYAML makes it a bit more difficult.
And I think more libraries will implement things like `<<` optionally. I think it's a good idea because `<<` is well known and very useful. Of course it makes our job harder...

Regarding `%YAML 1.2`, I'm not sure how often it will be used in the future. It's meant for that kind of thing, but I think people might avoid it because they don't want to type it in every YAML file.

For my YAML::PP library I made an option `schema`, which is an array, so you can say `schema => ['Core', 'Merge'])` for example, so I was thinking to do something similar for yamltidy. But while `Core` stands for a set of tags, `Merge` is just one tag. In YAML::PP, both are a class, e.g. `YAML::PP::Schema::Merge`, that's why it works. For getting the PyYAML behaviour I would probably just add another class `YAML::PP::Schema::PyYAML1_1`.

So I don't have a clear idea here yet how a linter/formatter config option could look like...

The so called tag repository is kind of hidden, yeah. You can find it in the 1.1 spec under [YAML tag repository](https://yaml.org/type/index.html).
YAML 1.1 was always a draft actually, and then for 1.2 it wasn't used anymore.
So then I put together this [page](https://perlpunk.github.io/yaml-test-schema/schemas.html).
The data in that repository can be used as the official test suite for resolving tags.
I hope we can move that into yaml-test-suite at some point.

perlpunk:
...and I just realized that I forgot `No` in the test data (I just have `NO` and `no`), that's why https://perlpunk.github.io/yaml-test-schema/data.html only shows 21 booleans!

adrienverge:
@perlpunk thank you! I'll bookmark https://perlpunk.github.io/yaml-test-schema/schemas.html as it's very handy (and I'm glad you could detect a missing entry in the test list thanks to my message :))

I agree that prepending explicit tags like `%YAML 1.2` is not practical and few users would do it. I'm attracted by the idea of making yamllint 1.2-compatible by default, but that would partially defeat the purpose of some rules (e.g. [`octal-values`](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.octal_values) or [`truthy`](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy) which help detecting values valid in 1.1 but not in 1.2).

Anyway, I believe the change of this pull request is a good move so I'll merge it.
And let's not lint *`<<` + YAML 1.2* cases since they are used in practice and quite useful.

perlpunk:
> @perlpunk thank you! I'll bookmark https://perlpunk.github.io/yaml-test-schema/schemas.html

Also, like I said, you can use the data https://github.com/perlpunk/yaml-test-schema/tree/master/data for your test suite.
I will add the `No` thing ASAP :)

PR #654, opened 2024-02-06
cli: Cleanly skip broken symlinks that are ignored

adrienverge:
Before this commit, yamllint would output "[Errno 2] No such file or directory" when running on a directory which contained a broken symbolic link, even if the file is set to be ignored in yamllint configuration.

This commit fixes that, and adds corresponding tests.

As a side effect this changes `yamllint.linter.run(stream, config)`, so tools that would use this API need to filter ignored files beforehand.

Fixes https://github.com/adrienverge/yamllint/issues/399

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -14,8 +14,10 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import contextlib
+from io import StringIO
 import os
 import shutil
+import sys
 import tempfile
 import unittest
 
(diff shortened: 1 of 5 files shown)


adrienverge:
Hello @andrewimeson, this change is related to https://github.com/adrienverge/yamllint/pull/524#discussion_r1061698032 and https://github.com/adrienverge/yamllint/pull/524#discussion_r1052261762. After reading the code, I think it's a good change, i.e. `yamllint.linter.run(stream, config)` should lint the YAML input it is given (independently of the filename), and the filter for ignored files should be done upstream. But if you think differently don't hesitate to say so :)

PR #682, opened 2024-08-12
docs: note about .yamllint file

erlangparasu:


adrienverge:
Hello, such new mention should come with details about other configuration locations. But this isn't really needed, because all this is already stated at [the beginning of the documentation about configuration](https://yamllint.readthedocs.io/en/stable/configuration.html#configuration).

erlangparasu:
as a new user i am used to reading the readme at the beginning, and earlier i was confused about what the default file name for config was. so i intend to add this so that maybe there are new users so they are not confused at the beginning.

but if you think this is not necessary, just ignore it.

PR #688, opened 2024-09-17
docs: fix position of backquotes

hituzi-no-sippo:


adrienverge (approved):
Thanks @hituzi-no-sippo!

PR #693, opened 2024-10-17
Apply new-lines check to all lines of a file, not just the first one

### LINKED ISSUE #475: rule new-lines: checks only the first newline character in a file
Currently, the `check()` function of `rules/new_lines.py` is only called once per file, checking only the first new-line-character of a file. ([see also this comment](https://github.com/adrienverge/yamllint/pull/474#discussion_r902714071))

To also be able to detect mixed newlines (and report the wrong ones) all lines of a file should be checked.

## Example

Consider this yaml-file (with the newline characters explicitly shown)

```yaml
---\n
test\r\n
\n
```

If it is checked for UNIX-style newlines, it should report an error on line 2 and if it is checked for DOS-style newlines it should report two errors on lines 1 and 3.

Currently, this file would be valid if checked with `type: unix` and invalid with `type: dos` (with only one error on line 1).

## Discussion

It should be discussed if all wrong newline characters should be reported.

In my opinion, this would clutter the error output too much and the amount of errors resulting from this should be caped. Either only one new-lines error per file or only a small amount (maybe max of 5?). Then the error message should say something like _"more new-line errors possible, but not showing..."_

---

### side effects

PR #474 introduced a new type and tests for it. Some test where supposed to test this type of scenario and therefore are currently broken. They can be reactivated once this issue is fixed. See:

https://github.com/adrienverge/yamllint/blob/40cab7f999b7d8f4f143f22f6932126274238e0d/tests/rules/test_new_lines.py#L77-L82

https://github.com/adrienverge/yamllint/blob/40cab7f999b7d8f4f143f22f6932126274238e0d/tests/rules/test_new_lines.py#L91-L96

dbk-rabel:
fixes #475

### DIFF
--- a/tests/rules/test_new_lines.py
+++ b/tests/rules/test_new_lines.py
@@ -72,7 +72,7 @@ class NewLinesTestCase(RuleTestCase):
             self.check('\n', conf)
             self.check('\r\n', conf, problem=(1, 1))
             self.check('---\ntext\n', conf)
-            self.check('---\r\ntext\r\n', conf, problem1=(1, 4), problem2=(2, 5)) # yamllint disable-line rule:line-length
+            self.check('---\r\ntext\r\n', conf, problem1=(1, 4), problem2=(2, 5))  # yamllint disable-line rule:line-length
             self.check('---\r\ntext\n', conf, problem=(1, 4))
             self.check('---\ntext\r\nfoo\n', conf, problem=(2, 5))
             self.check('---\ntext\r\n', conf, problem=(2, 5))
@@ -84,5 +84,5 @@ class NewLinesTestCase(RuleTestCase):
             self.check('---\r\ntext\r\n', conf)
             self.check('---\ntext\n', conf, problem1=(1, 4), problem2=(2, 5))
             self.check('---\ntext\r\n', conf, problem=(1, 4))
-            self.check('---\r\ntext\nfoo\r\n', conf, problem=(2, 4))
-            self.check('---\r\ntext\n', conf, problem=(2, 4))
+            self.check('---\r\ntext\nfoo\r\n', conf, problem=(2, 5))
+            self.check('---\r\ntext\n', conf, problem=(2, 5))
--- a/tests/rules/test_new_lines.py
+++ b/tests/rules/test_new_lines.py
@@ -72,7 +72,7 @@ class NewLinesTestCase(RuleTestCase):
             self.check('\n', conf)
             self.check('\r\n', conf, problem=(1, 1))
             self.check('---\ntext\n', conf)
-            self.check('---\r\ntext\r\n', conf, problem1=(1, 4), problem2=(2, 5))  # yamllint disable-line rule:line-length
+            self.check('---\r\n#\r\n', conf, problem1=(1, 4), problem2=(2, 2))
             self.check('---\r\ntext\n', conf, problem=(1, 4))
             self.check('---\ntext\r\nfoo\n', conf, problem=(2, 5))
             self.check('---\ntext\r\n', conf, problem=(2, 5))
--- a/tests/rules/test_new_lines.py
+++ b/tests/rules/test_new_lines.py
@@ -31,7 +31,7 @@ class NewLinesTestCase(RuleTestCase):
         self.check('---\ntext\n', conf)
         self.check('---\r\ntext\r\n', conf)
 
-    def test_unix_type_first_occurence(self):
+    def test_unix_type(self):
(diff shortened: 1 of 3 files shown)


adrienverge (commented):
Hello David,

Thanks for contributing!

Could you update the commit to report only the first wrong line (at most once per file)? Cf. discussion on the issue tracker #475, especially https://github.com/adrienverge/yamllint/issues/475#issuecomment-1163321459

dbk-rabel:
> Hello David,
> 
> Thanks for contributing!
> 
> Could you update the commit to report only the first wrong line (at most once per file)? Cf. discussion on the issue tracker #475, especially [#475 (comment)](https://github.com/adrienverge/yamllint/issues/475#issuecomment-1163321459)


Hello Adrian.
I think I missunderstood in the first place. Sorry.
Yes, I can update the commit. I'm just not a 100% sure how this is done the cleanest way. Are there other checks that implement the same thing? (I.e. checking the whole file but only yielding the first occurence?)

dbk-rabel:
Sorry, I don't seem to find a way to do it. I think we would somehow have to keep track of the files we already found an erronous line in. I'm not quite sure how this should be done and I doubt it can be done elegantly.

To be honest, I do not agree with the discussion in #475 . I personally think that it would be correct to produce a warning for each line.

I'd still love to implement it, if there was a good way to do it. Maybe you could help @adrienverge  ?

adrienverge:
Hello David,

> I personally think that it would be correct to produce a warning for each line.

I'm :+1: to fix a bug (wrong newlines are currently not detected if on line 2+) but it would be a bad idea to change current behavior, i.e. report 1000 errors for files containing 1000 lines, where 1 were previously reported. Many users would complain, and in my opinion they would be right. And it would be inconsistent with rule `syntax`.

> Maybe you could help @adrienverge ?

I'm willing to review and release ready-to-merge PRs on my free time, but at the moment I don't have the bandwidth to implement nor guide implementation of development requests, sorry :/

dbk-rabel:
> 
> I'm 👍 to fix a bug (wrong newlines are currently not detected if on line 2+) but it would be a bad idea to change current behavior, i.e. report 1000 errors for files containing 1000 lines, where 1 were previously reported. Many users would complain, and in my opinion they would be right. And it would be inconsistent with rule `syntax`.
> 

You convinced me. I agree: People would be upset and would complain. So such a drastic change in behaviour is probably a bad idea.

It still leaves me a litte bit sad, because I do think that 1000 errors would be the correct behaviour. :) But the argument holds: To change behaviour so drastically over night is probably never a good idea.

> > Maybe you could help @adrienverge ?
> 
> I'm willing to review and release ready-to-merge PRs on my free time, but at the moment I don't have the bandwidth to implement nor guide implementation of development requests, sorry :/

No worries. I actually wasn't aware that yamllint is something you maintain in your free time. My misstake. When a tool is used so widely, I always assume that there is some paid developers behind it.

Sorry if I was too demanding. I really appreciate your work, Thank you!

I hope I will find the time to take a closer look at this issue. But I'm not sure yet when that will be the case.

dbk-rabel:
@adrienverge What is your opinion on my today's approach? Would something like this be an alternative?

It feels a little bit unclean. But it seems to work, does not break current behaviour and could be used by other rules as well in the future, if needed.

adrienverge:
Hello David,

It looks like your last push added a new rule option `disable_after_first_occurence`, which is not really what we want.

The goal is rather to fix the existing rule, by reporting the first wrong newline character even if it's not on the first line (but only the wrong newline: at most once per file).

dbk-rabel:
Salut Adrien.

My idea with the new parameter `disable_after_first_occurence` was to be able not to hardcode some special behaviour for one specific rule in the linter.py . So I implemented some general parameter that than was used only by one of the rules.

My latest push would be the alternative: Tell the linter explicitly to handle "new-lines" differently than all other rules.

I fear that you might not like this solution either. But I'm giving my best and this is the cleanest solution I came up with.

Yours
  David

dbk-rabel:
@adrienverge
Happy new year! Hope you're doing well.

May I ask your opinion on my latest version of this pull request?

adrienverge:
Hello David, thanks for your message, and please excuse the long delay…

As you suspected, this `disable_by_force()` trick is not clean and cannot be merged. I'm sorry but I lack time, I can review this PR only if it's ready and implements the spec discussed in above comments. I apologize.

dbk-rabel:
Hello Adrian.

Thanks for the answer.

I tried everything I could think of. Maybe then there is no clean solution and I have to just live with this bug.

Yours
  David

PR #696, opened 2024-11-03
Update cli.py: encoding='utf-8'

BaseMax:
The issue happened in our project at https://github.com/SalamLang/Salam/issues/265 in `pre-commit` for lining YAML files.

> UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 445: character maps to

### DIFF
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -219,7 +219,7 @@ def run(argv=None):
     for file in find_files_recursively(args.files, conf):
         filepath = file[2:] if file.startswith('./') else file
         try:
-            with open(file, newline='') as f:
+            with open(file, newline='', encoding='utf-8') as f:
                 problems = linter.run(f, conf, filepath)
         except OSError as e:
             print(e, file=sys.stderr)


BaseMax:
cc @adrienverge @jbampton

adrienverge:
Hello and thanks for the proposal. Could you check out other pull requests related to character encoding? How does this one differ from them?

BaseMax:
Hi @adrienverge, happy connecting.

The https://github.com/adrienverge/yamllint/pull/630/files#diff-2e0288fc9fc3cda09f90a25f76bedb9ce0cea019d01147b436e575c71a3e674eR222 merge request looks fine but it doesn't have the change I applied.

My problem is that I have Farsi utf8 text in my yamel file and the problem was related to the 'cli.py' file.

Related to my issue https://github.com/adrienverge/yamllint/pull/240/files looks a good patch as it can automatically detect the encoding and then use that in reading the file.

adrienverge:
Hello Max, thanks. It looks like https://github.com/adrienverge/yamllint/pull/630 solves the same problem but is more complete and future-proof. Also, your PR doesn't fix encoding problems for other files such as configuration. What do you think?

> My problem is that I have Persian UTF8 text in my YAML files and the problem was related to the 'cli.py' file.

In the meantime, a solution is to tell Python to read files as UTF-8 by default:
```bash
export PYTHONUTF8=1
yamllint your-file.yaml
```

BaseMax:
Thank you @adrienverge, I added PYTHONUTF8 var to our `pre-commit` env config. https://github.com/SalamLang/Salam/commit/db7e870e233f9997e765ff00e7af91b515d6ef2a

@jbampton and I will do more testing.

asears:
I started a similar PR on this before seeing this one.  Having the option to do errors="ignore" and other encoding / error handling options would be beneficial.  The other in-flight PR looks really comprehensive though it might take awhile to release.  This seems like a useful summary of the options for improvements in that other PR. https://llego.dev/posts/comprehensive-guide-opening-closing-files-python/

In our scenario, we have a few emoji characters in yaml causing the charmap error.  

I was able to simply fix this issue with `with open(file, newline='', errors="ignore") as f:`
The fix in this PR to set to UTF-8 also works for us on Windows, as does the env var fix.  

Could this PR be tested, merged and released if it supports the solution while that much larger PR goes through the process?  It will help with Github workflows and other yaml files which may use emoji characters.  Might solve some long standing issues on Windows.

If there are any concerns with the feature it could be an experimental or preview arg feature toggle option?

Thanks for the helpful tool  Looking forward to a future --fix option.

adrienverge:
Hello @asears, I understand the proposal but no, sorry, I don't want to take any risk: only the long-term PR https://github.com/adrienverge/yamllint/pull/630 should be merged.

While waiting for it, if you would be OK for an opt-in experimental or preview arg feature toggle option, I guess you would be OK to use the `PYTHONUTF8=1` solution? It's not ideal, but it's easy and works now.

DimitriPapadopoulos:
Note that [`open()`](https://docs.python.org/3/builtins/functions.html#open) will use `encoding='utf-8'` by default on most recent Linux platforms:
> _encoding_ is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever [`locale.getencoding()`](https://docs.python.org/3/library/locale.html#locale.getencoding) returns), but any [text encoding](https://docs.python.org/3/glossary.html#term-text-encoding) supported by Python can be used.

Issue #673, opened 2024-05-27
ignore-from-file should use path of .yamllint.yaml as base, not current working dir


jellehelsen:
If there is a `.yamllint.yaml` at the base of the git repo, with `ignore-from-file` pointing to .gitignore, yamllint stops working in subdirectories. It will pickup the .yamllint.yaml file in a higher level directory, but then tries to find .gitignore in the current folder. This causes it to fail with  `FileNotFoundError: [Errno 2] No such file or directory: '.gitignore'` .

Making this extra annoying is that editor plugins like flycheck, that use yamllint, tend to use the directory where the file is located as the working dir. So having `ignore-from-file` in your project level `.yamllint.yaml` breaks in-editor linting.

adrienverge:
Hello and thanks for the report.

Yamllint doesn't handle Git repos differently than regular directories: if `ignore-from-file` points to a relative file, it will be searched from the current working directory. Looking for the *ignore file* in special locations (e.g. looking for a Git repo root in parent directories) would change current behavior and might annoy some users.

I suggest configuring flycheck to work like most Unix tools (if possible of course). Or maybe use an absolute path in `ignore-from-file`?

jellehelsen:
It's not really git related. Imho it does not make sense use the `.yamllint.yaml` that is in a parent directory and then fail because it can't find the `ignore-from-file` that is in that same folder.
Either don't use the `.yamllint.yaml` from a parent directory, or also find the files the `.yamllint.yaml` points to with a relative path. In the latter case the relative path should be relative to the `.yamllint.yaml` that is pointing to it.
In the current implementation, yamllint breaks when you go into a child folder.

adrienverge:
Hello Jelle, and sorry about the delay.

So the idea is that any file referenced by yamllint config would be relative to its own path.
But what about when this config is passed in another manner, e.g. via an environment variable (`yamllint -d …`)? It would then be relative to the current working directory?

@ndrwnaguib what do you think? (you contributed `ignore-from-file` in https://github.com/adrienverge/yamllint/pull/363 and https://github.com/adrienverge/yamllint/pull/591)

jellehelsen:
When there is no config file, the original behavior remains. Since there is no file to calculate a relative path to, the path gets used as is, making it relative to the current working directory.

PR #701, opened 2024-11-26
fix: ignore-from-file relative to config file

jellehelsen:
This fixes the problem (#673) that files mentioned in ignore-from-file can not be found when running in a subdirectory. Values in ignore-from-file are now treated as being relative to the config file.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -121,72 +121,78 @@ class SimpleConfigTestCase(unittest.TestCase):
         self.assertEqual(c.rules['hyphens'], False)
 
     def test_validate_rule_conf(self):
+        c = config.YamlLintConfig('rules:\n'
+                                  '  indentation:\n'
+                                  '    spaces: 2\n'
+                                  '    indent-sequences: yes\n'
+                                  '    check-multi-line-strings: false\n')
+
         class Rule:
             ID = 'fake'
 
-        self.assertFalse(config.validate_rule_conf(Rule, False))
-        self.assertEqual(config.validate_rule_conf(Rule, {}),
+        self.assertFalse(c.validate_rule_conf(Rule, False))
+        self.assertEqual(c.validate_rule_conf(Rule, {}),
                          {'level': 'error'})
 
-        config.validate_rule_conf(Rule, {'level': 'error'})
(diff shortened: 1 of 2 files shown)


adrienverge:
Hello!

Depending on the outcome of https://github.com/adrienverge/yamllint/issues/673 and what @ndrwnaguib thinks, this PR could be welcome and merged :+1:

I haven't reviewed yet, but I can already ask:
- Can you please minimize all the diff that can be? E.g. don't change `config` to `c`; keep `validate_rule_conf()` independant (not in a class) and pass it `config_dir` if needed; etc.
- Make the commit *clean* and ready to review. For example I saw a `self.maxDiff = None`.
- Document this new behavior by updating documentation, for example inside a new note below this one: https://github.com/adrienverge/yamllint/blob/8513d9b/docs/configuration.rst?plain=1#L229
- Please keep single quotes :pray:
- Add test cases for at least:
  \- an `ignore-from-file: file` when `file` is in the root dir but current dir is a sub dir (should find it)
  \- an `ignore-from-file: file` when `file` is in a sub dir but current dir is this sub dir (should error)
  \- an `ignore-from-file: file` when configuration is not passed via a file (found)
  \- an `ignore-from-file: file` when configuration is not passed via a file (error)
  \- other cases?

jellehelsen:
Hi,

I think I've addressed all your remarks. 
- `validate_rule_conf()` is back out of the class and get passed the config_dir with None as default value, restored corresponding test
- Tests added and existing tests cleaned. 
- Quotes restored (I should not have run `ruff format`. Apologies)
- Added a line of documentation at the suggested location to indicate that the ignore-from-file path is relative to the config file.

jellehelsen:
Hi Adrien,

My apologies this took so long.
I implemented your suggested tests. Where they differ, I commented. I also added some extra tests to verify the cases where the gitignore is not in one of the parents of the cwd and where cwd is a deep child of where gitignore is.

A gitignore file only applies to the folder it is in and it's descendants. So if the gitignore is not in a parent of the cwd it should be ignored.
In order the have the paths relative to the ignore file, the relative path to dirname(ignore_file) is being prepended to the pattern. 

Let me know if you have any questions or want any changes.

Issue #700, opened 2024-11-25
Backslash-escaped non-printable character in double-quoted string throws exception when `quoted-strings` are `only-when-needed`


jmknoble:
The following line of YAML:

```yaml
  echo_format: "\u001b[1;34m+ {command}\u001b[0m"
```

with the following .yamllint config:

```yaml
  quoted-strings:
    quote-type: any
    required: only-when-needed
```

Throws an exception (shown below).

Yamllint version:  yamllint 1.35.1
Python version:  Python 3.12.7

Dependency versions:

```
└── yamllint v1.35.1 (group: dev)
    ├── pathspec v0.12.1
    └── pyyaml v6.0.2
```

Reproducible problem documented at https://github.com/jmknoble/yamllint-quoted-escaped-nonprintable-char-problem .

The problem seems to be in   `yamllint.rules.quoted_strings._quotes_are_needed()` .  My hunch is that the already-decoded string is being passed, which `yaml.reader.check_printable()` interprets as unprintable.

Traceback:

```
Traceback (most recent call last):
  File "/Users/jmknoble/.local/bin/yamllint", line 8, in <module>
    sys.exit(run())
             ^^^^^
  File "/Users/jmknoble/.local/share/uv/tools/yamllint/lib/python3.12/site-packages/yamllint/cli.py", line 227, in run
    prob_level = show_problems(problems, file, args_format=args.format,
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jmknoble/.local/share/uv/tools/yamllint/lib/python3.12/site-packages/yamllint/cli.py", line 102, in show_problems
    for problem in problems:
                   ^^^^^^^^
  File "/Users/jmknoble/.local/share/uv/tools/yamllint/lib/python3.12/site-packages/yamllint/linter.py", line 199, in _run
    for problem in get_cosmetic_problems(buffer, conf, filepath):
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jmknoble/.local/share/uv/tools/yamllint/lib/python3.12/site-packages/yamllint/linter.py", line 136, in get_cosmetic_problems
    for problem in rule.check(rule_conf,
                   ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #703, opened 2024-12-03
fix: allow backslash-escaped special characters in double-quoted strings (#700)

jmknoble:
- Add test case to reproduce #700 
- When checking whether quotes are needed, assume that double-quoted strings containing special characters have already been escaped (and claim the quotes are in fact needed)

### DIFF
--- a/tests/rules/test_quoted_strings.py
+++ b/tests/rules/test_quoted_strings.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from tests.common import RuleTestCase
+import yaml.reader
 
 from yamllint import config
 
@@ -474,6 +475,30 @@ class QuotedValuesTestCase(RuleTestCase):
                    '- foo bar\n'
                    '- "foo bar"\n',
                    conf, problem1=(3, 3), problem2=(7, 3), problem3=(11, 3))
+
+    def test_only_when_needed_special_characters(self):
+        conf = 'quoted-strings: {required: only-when-needed}\n'
+        self.check('---\n'
+                   'k1: "\\u001b"\n',
+                   conf)
+        self.check('---\n'
+                   "k1: '\\u001b'\n",
+                   conf, problem=(2, 5))
+        self.check('---\n'
+                   'k1: \\u001b\n',
+                   conf)
+        self.assertRaises(yaml.reader.ReaderError, self.check,
+                          '---\n'
+                          'k1: "\u001b"\n',
+                          conf)
+        self.assertRaises(yaml.reader.ReaderError, self.check,
+                          '---\n'
+                          "k1: '\u001b'\n",
+                          conf)
+        self.assertRaises(yaml.reader.ReaderError, self.check,
+                          '---\n'
+                          'k1: \u001b\n',
+                          conf)
 
     def test_octal_values(self):
         conf = 'quoted-strings: {required: true}\n'
--- a/yamllint/rules/quoted_strings.py
+++ b/yamllint/rules/quoted_strings.py
@@ -204,10 +204,18 @@ def _quote_match(quote_type, token_style):
             (quote_type == 'double' and token_style == '"'))
 
 
-def _quotes_are_needed(string, is_inside_a_flow):
+def _quotes_are_needed(string, style, is_inside_a_flow):
     # Quotes needed on strings containing flow tokens
     if is_inside_a_flow and set(string) & {',', '[', ']', '{', '}'}:
         return True
+
+    if style == '"':
+        try:
+            yaml.reader.Reader('').check_printable('key: ' + string)
+        except yaml.reader.ReaderError:
+            # Special characters in a double-quoted string are assumed to have
+            # been backslash-escaped
+            return True
 
     loader = yaml.BaseLoader('key: ' + string)
     # Remove the 5 first tokens corresponding to 'key: ' (StreamStartToken,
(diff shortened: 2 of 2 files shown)


adrienverge:
Hello @jmknoble, thanks for contributing!

This fix looks good.

> Any wider change in this area likely requires more thought to understand whether it would be a breaking change.

Understood. I agree with your conclusion: let's fix double-quotes for now.

However I'm not comfortable with relying on an English string (`e.reason == "special characters are not allowed"`), it could change in the future and break. Instead, what about using `yaml.reader.Reader.check_printable()`, [which is the source of this string](https://github.com/yaml/pyyaml/blob/69c141adcf805c5ebdc9ba519927642ee5c7f639/lib/yaml/reader.py#L137-L144)? For example:
```python
>>> from yaml.reader import Reader
>>> Reader('').check_printable('\u0020')
>>> Reader('').check_printable('\u001b')
yaml.reader.ReaderError: unacceptable character #x001b: special characters are not allowed
```

> Repro test case fails properly before fix applied:
> Tests pass, including repro test case, after fix and linting applied:

:ok_hand:

Also can you squash all your commits into one, with a commit message starting with `quoted-strings: …` and a minimal rationale + reproducer inside it?

tests/rules/test_quoted_strings.py
```diff
@@ -475,6 +476,28 @@ def test_only_when_needed_extras(self):
...
+                     'k1: "\u001b"\n')
+        yamltext2 = ('---\n'
+                     # single-quoted unescaped special chars: yuck"
+                     "k1: '\u001b'\n")
+        yamltext3 = ('---\n'
+                     # unquoted unescaped special chars: yuck"
+                     'k1: \u001b\n')
+        self.assertRaises(yaml_reader_ReaderError, self.check, yamltext1, conf)
+        self.assertRaises(yaml_reader_ReaderError, self.check, yamltext2, conf)
+        self.assertRaises(yaml_reader_ReaderError, self.check, yamltext3, conf)
```
adrienverge, line 499: Can you please merge the two functions into one? I would find it be more easily understandable that they're about the same issue (like other tests).

Also you can remove all comments `ok / yuck`, we try to avoid those in all tests to be brief.

tests/rules/test_quoted_strings.py
```diff
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from tests.common import RuleTestCase
+from yaml.reader import ReaderError as yaml_reader_ReaderError
```
adrienverge, line 17: We never do such imports renaming: can you use simply `ReaderError`? Or `yaml.reader.ReaderError`?

adrienverge (commented):
> How does this look?

This looks good! Except for the last added empty line: it seems unrelated?

tests/rules/test_quoted_strings.py
adrienverge: Thanks for the change. Indeed I prefer the new version. I understand the readability argument, but it also applies when reading the whole set of tests: having one single `test_…()` is consistent with other tests for bugfixes, and eases reading yamllint tests globally.

adrienverge:
Hello @jmknoble,

Thanks for the update!

No need to open a new pull request, let's track this feature's history at a single place. You can push force to this one. Otherwise I'll squash myself before merging, with a commit message like:

    quoted-strings: Fix crash with only-when-needed and escaped special characters
   
    Prevent yamllint crash upon non-printable characters:
   
        $ yamllint -d 'rules: {quoted-strings: {required: only-when-needed}}' - <<<'key: "\u001b"'
        …
        yaml.reader.ReaderError: unacceptable character #x001b: special characters are not allowed
          in "<unicode string>", position 5
   
    Accept escaped special characters in double quoted strings (`key:
    "\u001b"`). But still throw exceptions for unescaped (embedded) special
    characters, whether or not quotes are present.
   
    When checking whether quotes are needed, assume that double-quoted
    strings containing special characters have already been escaped (and
    claim the quotes are in fact needed).
   
    Fixes https://github.com/adrienverge/yamllint/issues/700

Before this I just have a question, regarding your comment https://github.com/adrienverge/yamllint/pull/703#issuecomment-2515822973. Do you confirm that writing `k1: '\u001b'` or `k1: \u001b` is valid YAML, and that the 2 are equivalent? If my understanding is correct, then what do you think of adding the following code to make it clear?
```diff
     def test_only_when_needed_special_characters(self):
         conf = 'quoted-strings: {required: only-when-needed}\n'
         self.check('---\n'
                    'k1: "\\u001b"\n',
                    conf)
+        self.check('---\n'
+                   "k1: '\\u001b'\n",
+                   conf, problem=(2, 5))
+        self.check('---\n'
+                   'k1: \\u001b\n',
+                   conf)
         self.assertRaises(yaml.reader.ReaderError, self.check,
                           '---\n'
                           'k1: "\u001b"\n',
                           conf)
         self.assertRaises(yaml.reader.ReaderError, self.check,
                           '---\n'
                           "k1: '\u001b'\n",
                           conf)
         self.assertRaises(yaml.reader.ReaderError, self.check,
                           '---\n'
                           'k1: \u001b\n',
                           conf)
```

adrienverge (approved):
Hello @jmknoble, I hope you'll excuse the delay of my response...

Thanks a lot for checking this, and confirming with a Python script. Let's go then!

PS: I just allowed myself to force-push with this fix:
```diff
--- a/tests/rules/test_quoted_strings.py
+++ b/tests/rules/test_quoted_strings.py
@@ -13,9 +13,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from tests.common import RuleTestCase
 import yaml.reader
 
+from tests.common import RuleTestCase
+
 from yamllint import config
 
 ```

PR #706, opened 2024-12-17
style: Apply ruff rule RUF039

DimitriPapadopoulos:
RUF039 First argument to `re.compile()` is not raw string


adrienverge (approved):
Thanks Dimitri!

PR #715, opened 2025-03-06
Apply ruff/refurb rule FURB188

DimitriPapadopoulos:
FURB188 Prefer `str.removeprefix()` over conditionally replacing with slice.


adrienverge (approved):
Thank you Dimitri

PR #716, opened 2025-03-06
Apply ruff/flake8-pie rule PIE800

DimitriPapadopoulos:
PIE800 Unnecessary spread `**`

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -747,7 +747,7 @@ class CommandLineConfigTestCase(unittest.TestCase):
                                  (0, './a.yml:1:1: [warning] missing document '
                                      'start "---" (document-start)\n', ''))
 
-                with temp_workspace({**workspace, **{conf_file: conf}}):
+                with temp_workspace({**workspace, conf_file: conf}):
                     with RunContext(self) as ctx:
                         cli.run(('-f', 'parsable', '.'))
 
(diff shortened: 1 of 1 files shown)


adrienverge (approved):
:+1:

ssbarnea:
@adrienverge @DimitriPapadopoulos While I find this change useful, I need to ask why there is no pipeline running ruff on yamllint codebase?

The lack of it makes it hard to do a proper review of it (to see if it fixes the issue in all places) and more importantly, does not prevent regressions in the future.

DimitriPapadopoulos:
See https://github.com/adrienverge/yamllint/issues/628#issuecomment-1921677157.

ssbarnea:
@DimitriPapadopoulos That's the way to do it with any tool switch: "progressive". On Ansible DevTools projects (>10) we did run flake8 and ruff in parallel for about an year until we decided to remove flake8 because the number of stuff reported by flake8 was so small that it did not worth the burden of keep running it. Funny aspect that months later we started to remove black because we found ruff formatting good enough to replace black.

There is no such thing as perfect replacement. There is always something you might potential lose, is just to find the sweet spot on when to do it.

Nowadays we are considering removing pylint for the same reasons, most rules already implemented in ruff and the remaining ones were either not important enough.

Level of maintenance is another critical reason when picking one tool or another and ruff is progressing at blazing speed.

PR #720, opened 2025-03-11
Add ignore-regex option to comments rule

### LINKED ISSUE #545: Feature: support config rule comments.ignore-regex
In many of our YAML files, we got markers starting with `# renovate:` and those can be quite long depending on how much information we need to include in this marker, they cannot be divided into multiple lines, thus often violating the `rules.line-length` rule.

I have seen similar problems with `#cloud-info`, violating the `rules.comments.require-starting-space: true` rule.

I, therefore, propose (as has been done by @adrienverge multiple times already (https://github.com/adrienverge/yamllint/issues/428#issuecomment-991512486, https://github.com/adrienverge/yamllint/issues/416#issuecomment-951981783, https://github.com/adrienverge/yamllint/issues/361#issuecomment-784197416)) to add the option, `comments.ignore-regex`, example usage:

```yaml
rules:
  comments:
    ignore-regex:
      - ^#!\s*
      - ^#cloud-config
      - ^# renovate:
 ```

peschmae:
Fixes #545 

This adds a `ignore-regex` flag to the `comments` rule.

If a comment matches the regex, `require-starting-space` as well as `min-spaces-from-content` are skipped.

The regex is only matched against the comment itself, and not against the full line.


adrienverge (commented):
Hello @peschmae, thanks for contributing!

This looks good, but I have a few requests.

1. First, I disagree with myself of 2021 when I proposed `ignore-regex` as the option name. Indeed, other rules don't use "regex" in their option names when their content should be a regex. Similarly, rules don't tell "option-…-string" when their value should be a string.

   Instead, I propose `ignore-comments`. This will allow adding `ignore-lines` in the future, if one day we want to silent errors based on the preceding YAML line contents, in case of inline comment:
   ```yaml
   part of the line that should match the regex  # comment
   ```
3. Can you reword `Fixes #545` to `Partially fixes #545` ? Because the original request is about silenting both the `comments` and `line-length` rules, so in my opinion the issue #545 should remain open.
2. Finally, [write a good commit message](https://github.com/adrienverge/yamllint/blob/master/CONTRIBUTING.rst) (basically: the commit title should be `comments: …` and the message could be what you wrote in the first PR message).

PR #722, opened 2025-03-14
Give context about how many spaces were expected

mikalsande:
When telling someone that there are too few spaces before a comment it can be useful to include how many were expected.

### DIFF
--- a/yamllint/rules/comments.py
+++ b/yamllint/rules/comments.py
@@ -90,7 +90,8 @@ def check(conf, comment):
             comment.pointer - comment.token_before.end_mark.pointer <
             conf['min-spaces-from-content']):
         yield LintProblem(comment.line_no, comment.column_no,
-                          'too few spaces before comment')
+                          'too few spaces before comment: expected '
+                          f'{conf["min-spaces-from-content"]}')
 
     if conf['require-starting-space']:
         text_start = comment.pointer + 1


yamllint/rules/comments.py
```diff
@@ -86,11 +86,12 @@
...
 def check(conf, comment):
-    if (conf['min-spaces-from-content'] != -1 and comment.is_inline() and
+    expected = conf['min-spaces-from-content']
+    if (expected != -1 and comment.is_inline() and
             comment.pointer - comment.token_before.end_mark.pointer <
-            conf['min-spaces-from-content']):
+            expected):
         yield LintProblem(comment.line_no, comment.column_no,
-                          'too few spaces before comment')
+                          f'too few spaces before comment, expected {expected}')
```
adrienverge, line 94: - No need for a new variable, this will be good:
  ```python
                            'too few spaces before comment, expected '
                            f'{conf["min-spaces-from-content"]}')
  ```
- Let's use a colon `:` instead of `,`, to be consistent with other rules (like `indentation` or `new-lines`).

adrienverge (commented):
Hello, thanks for contributing!

It looks like the linter reported a line too long: can you check that?

When it's ready please squash everything into one commit (to be consistent and follow the guidelines it can be named `comments: …`).

yamllint/rules/comments.py
mikalsande: Sure, fixed now.

prestonvanloon:
Hit this issue today and came here to contribute exactly this. I hope it lands in a release soon, the current error message is terribly unhelpful. Thanks!

adrienverge:
Hello Preston, I just released it in yamllint version 1.37.1.

PR #725, opened 2025-03-16
restore full test suite to sdist

bollwyvl:
Congratulations on 1.36.1, and as always thank you for maintaining this tool!

[Downstream](https://github.com/conda-forge/yamllint-feedstock/pull/45) we noted that the `tests` folder is incomplete, failing to import with `tests.common` being missing. We shipped just getting those files from the GitHub-generated tarball, but it is generally to convenient to have a single canonical, immutable file (and SHA) to track.

I don't _think_ `sdist`-only files can be added directly in `pyproject.toml` with the `setuptools` legacy backend, so this PR adds a one-liner `MANIFEST.in` to preserve all the files in `tests` in the sdist. 

Thanks again!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+recursive-include tests *.*


adrienverge:
Hello Nicholas,

Thanks for the report!

Indeed this needs fixing. The present PR does not fully resolve the problem, because other files (e.g. documentation in `docs/`, `CHANGELOG.rst`, `CONTRIBUTING.rst`, etc.) are also missing.
Also we don't want to use a greedy pattern like`*.*`, otherwise we would ship dozens of unneeded files like `tests/rules/common.pyc` or the contents of `tests/__pycache__`.

So I implemented https://github.com/adrienverge/yamllint/pull/726 instead. Can you confirm it solves the problem on your side?

bollwyvl:
Sure, explicit is better than implicit!

Just shipped the (tested) `1.36.2` from https://github.com/conda-forge/yamllint-feedstock/pull/46 with just content from the PyPI sdist.

Frankly the best way to ensure this kind of thing is to have a CI excursion that does the equivalent of what a downstream would do:

- in one job
  - build the `sdist`  
  - upload with `actions/upload-artifacts`
- in another job (in the same workflow, and maybe on another python)
  - _don't_ check out the code
  - download with `actions/upload-artifacts`
  - install the sdist
    - `pip check` to catch any runtime dependency issues
  - install the sdist with the `[test]` extra
    - unpack (just) the tests
    - run the tests with `-W` and a coverage threshold to catch any packaging or test dependency issues 
  - reinstall the sdist with `[docs]` extra
    - build the docs with `-W` to catch any docs issues

Issue #721, opened 2025-03-12
Publish to PyPI from GHA with PEP 740 digital attestations


webknjaz:
This would allow showing more metadata as verified on PyPI and include verifiable dist signatures uploaded there too.

Here's the guide to follow: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/. By implementing it, the digital attestations will be included automatically. Ping me for review if need be.

adrienverge:
@webknjaz thanks, it's a good idea! I implemented this at https://github.com/adrienverge/yamllint/pull/724.

I'll merge now and create a new tag to make sure it works on the real PyPI, but feel free to review the pull request: we can make edits afterwards if needed.

webknjaz:
@adrienverge looks like you'll need to set up a trusted publisher on TestPyPI or drop that job. Although, if you want to keep it for regular PR merges, you need some extra version computation for untagged commits since (TestPyPI) won't allow version reuse. I usually use `setuptools-scm` for this.

adrienverge:
I did set up a trusted publisher on TestPyPI (same procedure as on PyPI) prior to pushing anything on the GitHub repo, so it looks like the HTTP 400 sometimes returned by test.pypi.org are rather due to duplication of pushes, as your comment suggests.

I want to fix this but I see a benefit in keeping these test publications to TestPyPI, as regular sanity checks.

To avoid pushing releases with the same version, we could either:
- run the whole worflow only if on master branch:
  ```yaml
  on:
    push:
      branches:
        - master
  ```
- run the workflow on all pushes, but the TesPyPI job only on master branch:
  ```yaml
    publish-to-testpypi:
      if: github.ref == 'refs/heads/master'
  ```

What do you think?
Is there a preferred option? An alternative option? Are there best practices?

adrienverge:
In the meantime I added `skip-existing: true` on TestPyPI to prevent these HTTP 400s, see https://github.com/adrienverge/yamllint/pull/727.

adrienverge:
Thanks for all these details Sviatoslav, it's interesting and valuable.

Here, the workflow can remain simple so I'd like to avoid variables (I understand they're useful in more complex cases like some of yours) and we can assume there won't be tags on anything else than on release commits.
And like most repo maintainers, I'd like tags to trigger publications, despite the drawback you pointed out. Less work (as long as everything goes well), easier to understand.

So in the end, I propose to simply change:

```diff
[...]

PR #727, opened 2025-03-17
CI: Ignore version duplicates when publishing to TestPyPI

adrienverge:
This commit prevents the failure of the GitHub Actions job "Publish distribution package to TestPyPI" on most pushes (see discussion at https://github.com/adrienverge/yamllint/issues/721#issuecomment-2727167030), since recent commit a3e1325 "CI: Publish PyPI releases using GitHub Actions workflows".

Indeed, TestPyPI doesn't allow repushing the same version. Use `skip-existing` to prevent to such failures on the CI.

An alternative would be to use the `setuptools-scm` plugin to define a unique version for each commit (e.g. `1.36.2.dev2+gd4f1c14`), but it requires more changes (define the version in `pyproject.toml`, fetch tags from GitHub in `publish.yaml`, exclude `.*` + `.github/**` files in `MANIFEST.in`). Moreover it doesn't work as is: to prevent errors like `HTTP 400 The use of local versions in '1.36.2.dev2+gd4f1c14' is not allowed` we would need `local_scheme = "no-local-version"`, which would push duplicated versions on TestPyPI anyway.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -46,6 +46,7 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           repository-url: https://test.pypi.org/legacy/
+          skip-existing: true
 
   publish-to-pypi:
     name: Publish distribution package to PyPI


webknjaz:
> we would need `local_scheme = "no-local-version"`, which would push duplicated versions on TestPyPI anyway.

@adrienverge that's not true. It would remove the local part (`+g` and the commit abbreviation). But the `.devX` bit would remain, and it'd be unique for as long as this is happening on the same branch (e.g. there's no parallel workflows attempting to publish the same commit).

adrienverge:
> @adrienverge that's not true. It would remove the local part (`+g` and the commit abbreviation). But the `.devX` bit would remain, and it'd be unique for as long as this is happening on the same branch (e.g. there's no parallel workflows attempting to publish the same commit).

Removing the `+g`  + commit abbreviation doesn't solve everything, at least with current configuration. Indeed, there can be multiple `.devX` with the same `X`, because the workflow job currently runs on every branch and push events (including on pull requests).
But I assume this is going to change soon, see https://github.com/adrienverge/yamllint/issues/721#issuecomment-2741267613.

webknjaz:
> > @adrienverge that's not true. It would remove the local part (`+g` and the commit abbreviation). But the `.devX` bit would remain, and it'd be unique for as long as this is happening on the same branch (e.g. there's no parallel workflows attempting to publish the same commit).
> 
> Removing the `+g` + commit abbreviation doesn't solve everything, at least with current configuration. Indeed, there can be multiple `.devX` with the same `X`, because the workflow job currently runs on every branch and push events (including on pull requests). But I assume this is going to change soon, see [#721 (comment)](https://github.com/adrienverge/yamllint/issues/721#issuecomment-2741267613).

FWIW, what you're doing in #728 is effectively the same I meant here, except for no `setuptools-scm` integration.

adrienverge:
> FWIW, what you're doing in #728 is effectively the same I meant here, except for no `setuptools-scm` integration.

Indeed, that's what I tried to do :)

PR #728, opened 2025-03-21
CI: Publish each master commit with a unique version on TestPyPI

adrienverge:
This follows commit 7d52df7 "CI: Ignore version duplicates when publishing to TestPyPI" with a better design:
- Only publish builds from `master` branch on TestPyPI.
- Version every non-tag commit with a `.devN` suffix, e.g. `1.36.2.dev1`. This prevents duplicates on TestPyPI.
- `twine check` built packages.

See discussion at
https://github.com/adrienverge/yamllint/issues/721#issuecomment-2731027692 for more details.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -8,19 +8,29 @@ jobs:
   build:
     name: Build distribution package
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
+      - name: Fetch tags
+        if: github.ref_type != 'tag'
+        run: git fetch --prune --unshallow --tags
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: "3.x"
       - name: Install pypa/build
-        run: python -m pip install --user build
+        run: python -m pip install build twine
+      - name: Add '.devN' to version for non-tag builds
+        if: github.ref_type != 'tag'
+        run:
+          sed -i
+          "/^APP_VERSION = /s/'$/.dev$(git describe --tags | cut -d- -f2)'/"
+          yamllint/__init__.py
       - name: Build a binary wheel and a source tarball
         run: python -m build
+      - name: Twine check the distribution packages
+        run: python -Im twine check --strict dist/yamllint-*
       - name: Store the distribution packages
         uses: actions/upload-artifact@v4
         with:
@@ -29,6 +39,7 @@ jobs:
 
   publish-to-testpypi:
     name: Publish distribution package to TestPyPI
+    if: github.ref_name == github.event.repository.default_branch
     needs: build
     runs-on: ubuntu-latest
     environment:
@@ -46,11 +57,10 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           repository-url: https://test.pypi.org/legacy/
-          skip-existing: true
 
   publish-to-pypi:
     name: Publish distribution package to PyPI
-    if: startsWith(github.ref, 'refs/tags/')  # only for tags
+    if: github.ref_type == 'tag'
     needs: build
     runs-on: ubuntu-latest
     environment:


.github/workflows/publish.yaml
```diff
@@ -8,19 +8,28 @@ jobs:
...
+      - name: Add '.devN' to version for non-tag builds
+        if: ${{ ! startsWith(github.ref, 'refs/tags/') }}
+        run:
+          sed -i
+          "/^APP_VERSION = /s/'$/.dev$(git describe --tags | cut -d- -f2)'/"
+          yamllint/__init__.py
       - name: Build a binary wheel and a source tarball
         run: python -m build
+      - name: Twine check the distribution packages
+        run: python -m twine check --strict dist/yamllint-*
```
adrienverge, line 32: @webknjaz `python -I` (that you suggested at https://github.com/adrienverge/yamllint/issues/721#issuecomment-2741928694) didn't work  because in isolated mode, `python` doesn't find user-installed libraries:

   /opt/hostedtoolcache/Python/3.13.2/x64/bin/python: No module named twine

webknjaz: Oh, I didn't realize. In general, it's discouraged to use `--user` because it's a global location with some things (like installed binaries) shared across several Python interpreters which may override each other.
I should really drop that from the guide… Nowadays, I default to having `-I` in all (or most of) my `python` invocations for safety reasons.

Technically, GH-provided VMs have `pipx` installed. So you could get away with just having `pipx run build` followed by a `pipx run twine check --strict dists/*`. If you want to pin versions, you can add `--spec` there.

.github/workflows/publish.yaml
```diff
@@ -29,6 +38,7 @@ jobs:
 
   publish-to-testpypi:
     name: Publish distribution package to TestPyPI
+    if: github.ref == 'refs/heads/master'
```
webknjaz, line 41: I like using the `github.ref_name` var (it was added later than I first wrote the guide, I think: `github.ref_name` and `github.ref_type` are a bit more convenient). Also, there's a way to get the repo default branch w/o having to hardcode it, been using it for years.
```suggestion
    if: >-
      github.ref_name == github.event.repository.default_branch
```

.github/workflows/publish.yaml
```diff
@@ -8,19 +8,28 @@ jobs:
...
+        run: git fetch --prune --unshallow --tags
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: "3.x"
       - name: Install pypa/build
-        run: python -m pip install --user build
+        run: python -m pip install --user build twine
+      - name: Add '.devN' to version for non-tag builds
+        if: ${{ ! startsWith(github.ref, 'refs/tags/') }}
```
webknjaz, line 24: I find it more convenient to use `>-` for escaping stuff and making it less nested. This allows us to drop the mustache:
```suggestion
        if: >-
          !startsWith(github.ref, 'refs/tags/')
```

Though, there's a nicer way to check for the same thing nowadays, more elegant:

```suggestion
        if: github.ref_type != 'tag'
```

.github/workflows/publish.yaml
```diff
@@ -8,19 +8,28 @@ jobs:
   build:
     name: Build distribution package
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
+      - name: Fetch tags
```
webknjaz, line 15: This can also be skipped for release requests:
```suggestion
      - name: Fetch tags
        if: github.ref_type != 'tag'
```

.github/workflows/publish.yaml
```diff
@@ -8,19 +8,28 @@ jobs:
...
+        run: git fetch --prune --unshallow --tags
```
webknjaz, line 16: Although, I just tend to build this right into the checkout step:
```suggestion
      - uses: actions/checkout@v4
        with:
          fetch-depth: >-
            ${{
              github.ref_type == 'tag'
              && 1
              || 0
            }}
          persist-credentials: false
```

.github/workflows/publish.yaml
adrienverge: These are 2 good points.

Updated with your last suggestion.

(I also changed to `if: github.ref_type == 'tag'` in job `publish-to-pypi` for consistency)

adrienverge (commented):
Thanks again @webknjaz!

.github/workflows/publish.yaml
webknjaz: > Smart, it would certainly work, but isn't more lightweight to only `git fetch --prune --unshallow --tags` (I think it fetches all ancestors of the current branch + all tags), rather to pull all history for _all branches_ and tags (that's what `fetch-depth: 0` does)?

It might be a little more efficient. I don't remember if `--unshallow` would get other branches or not.

> By the way, it would be better to use `fetch-tags: true` in `actions/checkout@v4`, but I tried and it didn't work. According to the internet it might be currently bugged.

I don't think having tags detached from the tree would work. They have to be grafted and exist in the context. `git describe --tags` would need all the commits between `FETCH_HEAD` and the last tag to (1) count them, but also to know which of those tags is closest on the tree. And whether those tags are reachable from `FETCH_HEAD`. Timestamps mean nothing, it should be a tag of the branch. Imagine there's a tag on some other branch — that wouldn't be what `git-describe` finds. Similarly, imagine there's a tag on the very first initial commit in the repo that is dated 2099 — it also won't be what `git-describe` finds.

webknjaz (approved):
Looks ready.

.github/workflows/publish.yaml
adrienverge: > I don't think having tags detached from the tree would work. They have to be grafted and exist in the context. `git describe --tags` would need all the commits between `FETCH_HEAD` and the last tag to (1) count them, but also to know which of those tags is closest on the tree. And whether those tags are reachable from `FETCH_HEAD`. Timestamps mean nothing, it should be a tag of the branch. Imagine there's a tag on some other branch — that wouldn't be what `git-describe` finds. Similarly, imagine there's a tag on the very first initial commit in the repo that is dated 2099 — it also won't be what `git-describe` finds.

:+1:

PR #731, opened 2025-03-28
Add (optional) multi-processing functionality

danudey:
Add a command-line option to allow an end-user to specify a number of
separate worker processes to use, in order to potentially speed up
linting of large repositories.

Default the option to 1 to (largely?) preserve the current behavior.
Possibly worth noting that, technically, we are still multiprocessing
with a value of '1', as we still spin up a separate worker process from
the main process and then just wait on that process to finish the work.

### DIFF
--- a/yamllint/cli.py
+++ b/yamllint/cli.py
@@ -18,6 +18,7 @@ import locale
 import os
 import platform
 import sys
+import multiprocessing
 
 from yamllint import APP_DESCRIPTION, APP_NAME, APP_VERSION, linter
 from yamllint.config import YamlLintConfig, YamlLintConfigError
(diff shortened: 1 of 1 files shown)


yamllint/cli.py
```diff
@@ -143,6 +144,18 @@ def find_project_config_filepath(path='.'):
...
+def process_file(args):
+    file, conf = args
+    filepath = file.removeprefix('./')
+    try:
+        with open(file, mode='rb') as f:
+            problems = linter.run(f, conf, filepath)
+    except OSError as e:
+        print(e, file=sys.stderr)
+        sys.exit(-1)
+    return list(problems), file
```
danudey, line 156: Because we can't pass a generator back from one process to another, we convert the `problems` generator into a list. This is _most likely_ fine, though technically it will increase memory use temporarily.

I only really see this being a problem if someone has a large number of files with a large number of problems, all of which finish scanning at the same time but are forced to buffer while another file's results are being printed.

danudey: Resolved this with `concurrent.futures`.

danudey:
Open to discussion about whether the `--continue-on-error` behavior should be a separate pull request, or if it's acceptable to have it in here as well.

danudey:
Did my best to clean up some of the code coverage. Some branches are Python version-dependent (as we use `os.get_process_cpu()` if it's available). Others I'm not sure why those lines are being flagged as I don't believe that any of my changes should have resulted in different code paths. If they did, happy to make those changes.

Issue #275, opened 2020-05-27
quoted-strings only-when-needed doesn't work properly for multiline strings


tjanez:
Example YAML file `input.yml`:

```yaml
---
multiline_url_with_quotes: "https://github.com/cloudposse/\
  github-authorized-keys/releases/download/1.0.3/\
  github-authorized-keys_linux_amd64"

multiline_url_without_quotes: https://github.com/cloudposse/\
  github-authorized-keys/releases/download/1.0.3/\
  github-authorized-keys_linux_amd64

```

Yamllint 1.23.0 reports:
```
$ yamllint -d '{extends: default, rules: {quoted-strings: {required: only-when-needed}}}' input.yml
input.yml
  2:28      error    string value is redundantly quoted with any quotes  (quoted-strings)

```

However, the quotes are actually not redundant:

```python
>>> import yaml
>>> with open("input.yml") as f:
...     data = yaml.safe_load(f)
... 
>>> data['multiline_url_with_quotes']
'https://github.com/cloudposse/github-authorized-keys/releases/download/1.0.3/github-authorized-keys_linux_amd64'
>>> data['multiline_url_without_quotes']
'https://github.com/cloudposse/\\ github-authorized-keys/releases/download/1.0.3/\\ github-authorized-keys_linux_amd64'
>>>
```

adrienverge:
Thanks for the report! Contributions to fix this are welcome.

julienfalque:
The issue also happens with values that would be parsed as octal values:

```yaml
foo: '0123456789'
```

PR #742, opened 2025-04-22
quoted-strings: Double quotes on multiline are not redundant

tomasmiguez:
Those cases where getting flagged:

```bash
$ python yamllint -d 'rules: {quoted-strings: {required: only-when-needed}}' \
  <(echo 'multiline string 4:\n  "word 1\\\n  word 2"\n')
/proc/self/fd/11
  2:3       error    string value is redundantly quoted with any quotes  (quoted-strings)
```

But that actually has usecases. It's the only way of splitting a string in multiple lines without adding a whitespace. For things like URLs, or regexps this is needed.

Fixes https://github.com/adrienverge/yamllint/issues/275

### DIFF
--- a/tests/rules/test_quoted_strings.py
+++ b/tests/rules/test_quoted_strings.py
@@ -129,7 +129,7 @@ class QuotedValuesTestCase(RuleTestCase):
                    'multiline string 4:\n'
                    '  "word 1\\\n'
                    '   word 2"\n',
-                   conf, problem1=(9, 3), problem2=(12, 3))
+                   conf, problem1=(9, 3))
 
     def test_quote_type_double(self):
         conf = 'quoted-strings: {quote-type: double}\n'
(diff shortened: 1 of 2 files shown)


adrienverge:
Hello @tomasmiguez, thanks for tacking this problem :+1:

Unfortunately this pull request cannot be merged:
- It would create a regression, the following redundant quotes wouldn't be reported anymore:
  ```yaml
  quotes are not needed:
    "word1
    word2"
  ```
- Another problem: it returns very early, but doesn't check that the type of quote (single vs. double) is the right one.
  
- In addition, it would needed some rework (the Python code style not consistent with the rest of the code, unneeded parentheses, the logic should be added inside `_quotes_are_needed()`, commit title…)

Instead, I created the following fix, which checks for `\` at the end of lines, checks the type of quotes, and adds non-regression tests: https://github.com/adrienverge/yamllint/pull/746
Please let me know if it solves the problem for you.

adrienverge:
Hello @tomasmiguez, thank you for testing and reporting :+1:

I just released it in yamllint version 1.37.1.

PR #746, opened 2025-04-30
quoted-strings: Fix only-when-needed on multiline with backslash

adrienverge:
On double-quoted multiline strings, quotes aren't needed if lines are broken on spaces, e.g.:

    multiline:
      "this is a sentence
      cut into words"

But quotes are needed when at least one line ends with a backslash character (`\`), meaning that the next spaces should be removed:

    multiline:
      "https://example.com/a/very/very\
      /very/very/long/URL"

This commit fixes that.

Fixes https://github.com/adrienverge/yamllint/issues/275


adrienverge:
See https://github.com/adrienverge/yamllint/pull/742#issuecomment-2841949268.

Closes https://github.com/adrienverge/yamllint/pull/742.

Issue #274, opened 2020-05-26
TypeError: 'NoneType' object is not iterable


mgedmin:
I noticed that my vim stopped highlighting YAML syntax errors.  I ran yamllint directly on a file I was editing and saw

```
Traceback (most recent call last):
  File "/home/mg/.local/bin/yamllint", line 8, in <module>
    sys.exit(run())
  File "/home/mg/.local/pipx/venvs/yamllint/lib/python3.8/site-packages/yamllint/cli.py", line 168, in run
    conf = YamlLintConfig(file=user_global_config)
  File "/home/mg/.local/pipx/venvs/yamllint/lib/python3.8/site-packages/yamllint/config.py", line 42, in __init__
    self.parse(content)
  File "/home/mg/.local/pipx/venvs/yamllint/lib/python3.8/site-packages/yamllint/config.py", line 83, in parse
    for rule in self.rules:
TypeError: 'NoneType' object is not iterable
```

Turns out this was caused by my `~/.config/yamllint/config` containing

```yaml
extends: relaxed
rules:
line-length:
  max: 200
```

I'm pretty sure this used to work before I did some pip upgrades.

I fixed my problem by changing the config to explicitly set `rules: []`, but I wonder if you'd want to catch this kind of type error in the YAML config and produce better error messages?  Or even allow `null` values and treat them as empty lists/dicts?

adrienverge:
Hi Marius,

I'm glad you managed to get it to work again :+1: 

If using an empty `rules:` once worked, it was a hidden feature that was never meant to exist. For clarity and consistency, I think it's better to stick to what `rules` is (a dict). I agree a more friendly error message would be better.

mgedmin:
Oh, it's a dict!  I'll go and fix my config again, before stricter typechecking breaks my editor plugins, again :)

adrienverge:
Hehe. You can see an example of a full configuration here: https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration

PR #747, opened 2025-05-01
Add error check for invalid config

koyuki7w:
In the current version, the error message states `TypeError: 'NoneType' object is not iterable` when the 'rules' in the config file is not a dict (gh-274.) It would be helpful to improve the error message.

Additionally, it is for consistency, as other elements(`ignore`, `locale`, etc.) have similar error checks.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -199,6 +199,10 @@ class SimpleConfigTestCase(unittest.TestCase):
     def test_invalid_rule(self):
         with self.assertRaisesRegex(
                 config.YamlLintConfigError,
+                'invalid config: rules should be a dict'):
+            config.YamlLintConfig('rules:\n')
+        with self.assertRaisesRegex(
+                config.YamlLintConfigError,
                 'invalid config: rule "colons": should be either '
                 '"enable", "disable" or a dict'):
             config.YamlLintConfig('rules:\n'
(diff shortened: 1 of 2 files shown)


adrienverge (approved):
Hello Yuki,

Thanks for this addition, this is a good idea.

I wonder whether "dict" is the best user-facing word (in "*rules should be a dict*"), because "dict" seems a Python-specific word to me. But anyway, we can always change it later if needed.

koyuki7w:
According to [the YAML specification](https://yaml.org/spec/1.2.2/#3211-nodes), "mapping" seems to be a more appropriate term. The current yamllint documentation uses "mapping" and does not mention "dict." 

Perhaps all three instances of "dict" in the error message should be changed to "mapping." 

https://github.com/adrienverge/yamllint/blob/79a6b2b1392eaf49cdd32ac4f14be1a809bbd8f7/yamllint/config.py#L82 https://github.com/adrienverge/yamllint/blob/79a6b2b1392eaf49cdd32ac4f14be1a809bbd8f7/yamllint/config.py#L86 https://github.com/adrienverge/yamllint/blob/79a6b2b1392eaf49cdd32ac4f14be1a809bbd8f7/yamllint/config.py#L237-L238

adrienverge:
Thanks for your input @koyuki7w. I agree and I created https://github.com/adrienverge/yamllint/pull/751 to address this.

PR #749, opened 2025-05-04
Fix link markup in the documentation

koyuki7w:
Appended a missing underscore so the link works correctly.


adrienverge (approved):
Hello Yuki, thank you for the fix :+1:

PR #750, opened 2025-05-04
Python 3 is the default nowadays

DimitriPapadopoulos:


adrienverge:
Hello Dimitri,

Thank for the proposal, I agree that every simplification is good to take.

I see you closed the pull request, I guess it's related to the warning output on GitHub action:

> *Linters*
> The `python-version` input is not set. The version of Python currently in `PATH` will be used.

-- https://github.com/adrienverge/yamllint/actions/runs/14820565906

DimitriPapadopoulos:
Yes, it's mostly the warning.

DimitriPapadopoulos:
Removing `python-version` would make `actions/setup-python` a no-op, as it falls back on the version of Python bundled with the runner image.

Instead, `python-version: 3.x` means: latest release of Python 3.

adrienverge:
:+1:

PR #755, opened 2025-05-27
Fix deprecated license specification format

koyuki7w:
In the current version, `pip install .` emits the following warnings:
```
********************************************************************************
Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

By 2026-Feb-18, you need to update your project and remove deprecated calls
or your builds will no longer be supported.

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
********************************************************************************
```
```
********************************************************************************
Please consider removing the following classifiers in favor of a SPDX license expression:

License :: OSI Approved :: GNU General Public License v3 (GPLv3)

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
********************************************************************************
```
(These warnings can be seen by using `--log` option.)

So fixed the way to specify the license as these warnings suggest. Also see https://peps.python.org/pep-0639/#deprecate-license-classifiers and https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,14 +3,13 @@ name = "yamllint"
 description = "A linter for YAML files."
 readme = {file = "README.rst", content-type = "text/x-rst"}
 requires-python = ">=3.9"
-license = {text = "GPL-3.0-or-later"}
+license = "GPL-3.0-or-later"
 authors = [{name = "Adrien Vergé"}]
 keywords = ["yaml", "lint", "linter", "syntax", "checker"]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Environment :: Console",
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
     "Programming Language :: Python",
     "Topic :: Software Development",
     "Topic :: Software Development :: Debuggers",


DimitriPapadopoulos:
Indeed [PEP 639](https://peps.python.org/pep-0639/) support requires Setuptools [77](https://setuptools.pypa.io/en/stable/history.html#v77-0-0).

However, that's not a problem because users should upgrade to the latest version any way. From [Installing Packages](https://packaging.python.org/en/latest/tutorials/installing-packages/)
> ### [Ensure pip, setuptools, and wheel are up to date](https://packaging.python.org/en/latest/tutorials/installing-packages/#ensure-pip-setuptools-and-wheel-are-up-to-date):
> 
> While `pip` alone is sufficient to install from pre-built binary archives, up to date copies of the `setuptools` and `wheel` projects are useful to ensure you can also install from source archives:

DimitriPapadopoulos:
I'm not sure it must be dropped. Isn't `setuptools >= 77` required when packaging (`python -m build`)

adrienverge:
Thanks for the answers @koyuki7w and @DimitriPapadopoulos.

The first commit is good to go and useful, so I'll merge it right now.

About the `setuptools >= 61` vs. `77` requirement, I haven't searched but it looks like Dimitri could be right and this could be a build-only dependency, in which case it shouldn't impact end-users. Depending on what you think about it, can you open a new pull request so we can merge it there (if relevant)?

PR #759, opened 2025-06-21
style: Apply assorted ruff rules

DimitriPapadopoulos:
Mostly [flake8-simplify (SIM)](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim) rules.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -101,7 +101,7 @@ def unregister_test_codecs():
 
 
 def is_test_codec(codec):
-    return codec in test_codec_infos.keys()
+    return codec in test_codec_infos
 
 
 def built_in_equivalent_of_test_codec(test_codec):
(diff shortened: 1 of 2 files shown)


adrienverge:
Hello Dimitri, thanks for the housekeeping!

I like those changes, except for the first commit which I find harder to understand, in this precise case. If you insist I can merge it, but if you agree with me, you could avoid it.

Also can you just start commit titles with `style:`?

adrienverge:
Hello, I see you pushed a new version of the PR :+1: 
Next time don't hesitate to post a small message to tell what you think, and so I get a notification.

For reference, it looks like you indeed removed the first commit (thanks) which was:

    Apply ruff/flake8-simplify rule SIM103

    SIM103 Return the negated condition directly

PR #760, opened 2025-06-22
style: Enforce ruff rules

DimitriPapadopoulos:
Implement https://github.com/adrienverge/yamllint/pull/716#issuecomment-2740687520.

Requires #759 to be applied first.

### DIFF
--- a/yamllint/rules/trailing_spaces.py
+++ b/yamllint/rules/trailing_spaces.py
@@ -33,7 +33,7 @@ Use this rule to forbid trailing spaces at the end of lines.
     this document contains     """ """
     trailing spaces
     on lines 1 and 3         """ """
-"""
+"""  # noqa: ISC001
 
 
 import string
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -21,10 +21,11 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
       - run:
-          pip install flake8 flake8-import-order sphinx sphinx_rtd_theme
+          pip install flake8 flake8-import-order ruff sphinx sphinx_rtd_theme
           rstcheck[sphinx] doc8
       - run: pip install .
       - run: flake8 .
+      - run: ruff check --output-format=github
       - run: doc8 $(git ls-files '*.rst')
       - run: rstcheck --ignore-directives automodule $(git ls-files '*.rst')
       - run: yamllint --strict $(git ls-files '*.yaml' '*.yml')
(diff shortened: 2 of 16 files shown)


yamllint/rules/trailing_spaces.py
```diff
@@ -30,10 +30,10 @@
    the following code snippet would **FAIL**:
    ::
 
-    this document contains     """ """
+    this document contains     
     trailing spaces
-    on lines 1 and 3         """ """
-"""
+    on lines 1 and 3
```
DimitriPapadopoulos, line 35: Not sure about removing `""" """` after trailing spaces. What's your opinion?

DimitriPapadopoulos:
The flake8-simplify rules appear to be the least liked _ruff_ rules. I'll remove `SIM103`. Note that [SIM108](url) is about embracing new Python constructs, you'll get used to them eventually — but I'll remove it too.

Concerning the isort rule `I001`, I presume I can find an option that does what you want. However, note that most maintainers prefer ordering imports using the default ordering: i) standard Python libraries, ii) external libraries, and finally iii) internal imports. They can't be collectively wrong. Well, maybe they can, but that's the standard and everyone is getting used to it. I guess your implicit argument is that you would like to avoid extensive changes for the mere purpose of reordering imports.

As for the ruff rules, I used to find them intrusive and used to disable them. I guess the idea is to:
1. avoid ambiguity (which exact Unicode code point am I seeing here?),
2. be consistent in the use of punctuation.

I'm OK to remove them if you want me to.

adrienverge (commented):
Thanks a lot for your feedback @DimitriPapadopoulos @Jayman2000.

Dimitri I see you dropped the RUF001 + RUF002 + RUF003 commits: given what the Unicode Core Specification says, I feel it's the right thing to do :+1:

- `style: Apply ruff/flake8-implicit-str-concat rule ISC001`: ✅
- `style: Ignore ruff/Perflint rule PERF203`: ✅
- `style: Apply ruff/pyupgrade rule UP015`: ✅
- `style: Apply ruff/pyupgrade rule UP031`: ✅
- `style: Enforce ruff rules`: ✅
- `style: Apply ruff/isort rule I001`: I indeed prefer your latest push, because my gut feeling is that the `force-sort-within-sections` way of sorting imports (e.g. sort `from io import StringIO` inside "i") is more natural. Hence I would lean towards this, especially since the existing codebase mostly does like this. However, if you think that top Python projects on the internet use the other way of sorting (e.g. sort `from io import StringIO` inside "S"), then let's align with them. I admit I'm not aware of how they do.
- `CI: Specify the latest version of Python 3` and `CI: Test more versions of Python`: can you move them in a dedicated pull request? The review and merging should be easy.

PR #768, opened 2025-07-31
build: PEP 639 support requires setuptools ≥ 77

DimitriPapadopoulos:
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files


adrienverge (approved):
Thanks!

PR #769, opened 2025-07-31
build: time to retire `setup.py`

DimitriPapadopoulos:


adrienverge (commented):
I have the same will to retire this file :+1:
But I haven't checked whether some packagers could still be using it: do you think it won't impact anybody now?

PR #770, opened 2025-07-31
build: revert 0cb7b0d

DimitriPapadopoulos:
Most if not all build backends pick up files with standard names by default, especially `LICEN[CS]E*`. That's the case with setuptools:
https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html

> * All files specified by the `license-files` configuration parameter in `pyproject.toml` and/or equivalent in `setup.cfg`/`setup.py`; note that if you don’t explicitly set this parameter, `setuptools` will include any files that match the following glob patterns: `LICEN[CS]E*`, `COPYING*`, `NOTICE*`, `AUTHORS**`;

No need to explicitly specify licence files, implicit is fine.


adrienverge (approved):
OK. Too bad we merged https://github.com/adrienverge/yamllint/pull/766 just last week, then :smile:

PR #791, opened 2025-11-12
Include Alpine Linux installation for yamllint

harsh183:
https://pkgs.alpinelinux.org/package/edge/community/x86/yamllint

Looks like this exists and is regularly updated. Adding this in the quickstart guide can help people find it, especially in Docker contexts.


adrienverge:
OK :+1: 

Thanks for contributing!

PR #281, opened 2020-07-08
Add Dockerfile and docker-compose.yml to run tests and linter

wolfgangwalther:
Had trouble setting up a local dev environment on Windows while working on #280, so I created some Docker and Docker Compose files. In case you want to add those to your repo, I opened this PR.

Just run `docker-compose up test` and `docker-compose up lint` to run tests and the linter.

PR #191, opened 2019-07-14
Add Dockerfile

arbourd:
Adds Dockerfile for official builds.

Can be easily paired with an [automated build on Docker Hub](https://docs.docker.com/docker-hub/builds/) or Travis CI to create fresh containers for each release.

This would be amazing for those of us who use yamllint in container-based CI platforms.

Closes #118

PR #179, opened 2019-05-23
Added dockerfile

xtellurian:
* This PR adds a dockerfile

* The image let's you run yamllint with the command:

```
docker run -v `pwd`:/files IMAGE:TAG /files
```

PR #794, opened 2025-11-16
Container: Add initial container for yamllint

oliv3r:
Just a quick set of files to enable the build of (tagged) containers, allowing us to use yamllint cli with a quick docker run, instead of having to build/install stuff locally.

The README can be updated once a few containers have been tagged, published and validated.

Note, do not run with `docker run --tty` when wanting to redirect stdin!

docker run --rm --interactive yamllint:master -s - < /path/to/file

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+docs/
+tests/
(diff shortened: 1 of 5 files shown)


adrienverge:
Hello, thanks for the proposal, but please check existing pull requests: #281, #179, #191, #154, #118, #434, #575, etc.

Issue #800, opened 2026-01-06
pathspec=1.0.0 causes deprecation warnings


ssbarnea:
Use of pathspec needs to be updated to avoid runtime deprecation warnings coming from [pathspec 1.0.0](https://github.com/cpburnz/python-pathspec/releases/tag/v1.0.0) that was released few hours ago.

```
DeprecationWarning: GitWildMatchPattern ('gitwildmatch') is deprecated. Use 'gitignore' for GitIgnoreBasicPattern or GitIgnoreSpecPattern instead.
```

Note: we run CI with warnings as errors so for the time being we will be ceiling down pathspec.

PR #802, opened 2026-01-12
[pathspec >= 1.0.0] Update with `GitIgnoreSpec.from_lines`

hswong3i:
To follow Git's implementation, use GitIgnoreSpec.from_lines(...). The change for using this is Git allows including files from excluded directories which directly contradicts the gitignore docs.

See https://github.com/adrienverge/yamllint/issues/800#issuecomment-3716842204

Fixes https://github.com/adrienverge/yamllint/issues/800

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
     "Topic :: Software Development :: Testing",
 ]
 dependencies = [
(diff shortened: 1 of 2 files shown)


hswong3i:
@adrienverge thank you for your quick release, your changes looks much elegant than mine ;-)

PR #803, opened 2026-01-13
Require pathspec ≥ 1.0.0 and follow Git's gitignore implementation

adrienverge:
To follow Git's implementation, use `GitIgnoreSpec.from_lines(…)`. The change for using this is Git allows including files from excluded directories which directly contradicts the gitignore docs.

Thanks to Wong Hoi Sing Edison for the initial proposal.

Fixes https://github.com/adrienverge/yamllint/issues/800.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
     "Topic :: Software Development :: Testing",
 ]
 dependencies = [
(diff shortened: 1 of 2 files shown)


hswong3i:
@adrienverge lovely and thank you very much ;-)

PR #810, opened 2026-04-07
fix: ignore list not being used when passing files/directories as arguments

Zamiell:
Imagine that I put "1.yml" on the ignore list. Then, if I invoke yamllint like this:

```sh
yamllint 1.yml 2.yml 3.yml 4.yml 5.yml 6.yml
```

Then, instead of linting 2-6, it will actually lint 1-6! Which seems almost certainly to be a bug, because other industry-standard tools such as [Prettier](https://prettier.io/) will correctly skip 1.yml.

The behavior occurs with both files and directories. I didn't bother opening an issue in favor of just fixing the bug myself in a pull request.

One other situation that I thought about was when the user does not know that "1.yml" is on the ignore list and types `yamllint 1.yml`, expecting it to find errors, but instead yamllint passes. This could lead the user to think that yamllint rules are turned off or yamllint is bugged in some way. To address this, I also decided to print a warning to standard error when this happens, to explicitly let the user know that the file is being skipped (as a quality of life feature). However, note that Prettier does not do this, so it might not be needed.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -151,7 +151,7 @@ It is possible to exclude specific files or directories, so that the linter
 doesn't process them. They can be provided either as a list of paths, or as a
 bulk string.
 
-You can either totally ignore files (they won't be looked at):
+You can exclude files found during directory traversal:
 
 .. code-block:: yaml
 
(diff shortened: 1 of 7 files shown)


DimitriPapadopoulos:
But then other tools such as [Ruff](https://docs.astral.sh/ruff/) do override their `exclude` list when explicitely passing an ignored file as an argument. I'm not certain which behaviour is best, but It doesn't look like there's consensus on the subject.

Zamiell:
@DimitriPapadopoulos Interesting, I didn't know that Ruff did that!

For context, in my company we have the following linting algorithm:

1) Detect the file paths changed in the pull request.
2) Feed them to `uv run yamllint a.yml b.yml c.yml` and so on.

Thus, the behavior in my pull request is preferable for this pattern. Because currently, when someone submits a pull request for a file on the ignore list, they erroneously get hit by yamllint errors!

Ruff's behavior does make more sense for humans typing in the command manually. But with that said, I think the explicit warning that I put into this PR is a great middle ground that covers both sides!

DimitriPapadopoulos:
Ruff has [`exclude`](https://docs.astral.sh/ruff/settings/#exclude) and [`force-exclude`](https://docs.astral.sh/ruff/settings/#force-exclude) for that purpose.

Zamiell:
ok, I revamped the PR to add "force-exclude", matching the ruff implementation. LLM summary:

> This PR adds Ruff-style exclusion behavior:
> 
>   - Explicitly specified files are linted even if they match ignore or ignore-from-file.
>   - --force-exclude or force-exclude: true makes those patterns apply to explicit files too, skipping them silently before opening.
>   - Files discovered through directory traversal continue to respect ignore patterns.
>   - Rule-specific ignores remain effective, and --list-files follows the same file selection.
> 
>   This supports automation that passes changed files directly:
> 
>   yamllint --force-exclude a.yml b.yml c.yml
> 
>   The configuration option defaults to false and supports inheritance. Tests and documentation cover both behaviors.
> 

I guess it is a little inconsistent, since yamllint uses "ignore" instead of "exclude", so maybe it should be called "force-ignore", but I also like the idea of aligning with ruff, so it could go either way probably.

PR #813, opened 2026-06-18
Support FORCE_COLOR and NO_COLOR

hashar:
When running on CI the process stdout is not a tty and coloring is disabled unless one uses `--format colored`.

Add support for forcing colored output via the environment variable `FORCE_COLOR` and add `NO_COLOR` (which takes precedence).

The spec from https://force-color.org/ states an environment variable is taken in account when:
- it is present
- AND NOT an empty string

> Command-line software which outputs colored text should check for a FORCE_COLOR environment variable. When this variable is present and not an empty string (regardless of its value), it should force theaddition of ANSI color.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -22,6 +22,7 @@ import shutil
 import sys
 import tempfile
 import unittest
+from unittest import mock
 
 from tests.common import (
     RunContext,
@@ -44,6 +45,44 @@ def utf8_available():
         return False
     else:
         return True
+
+
+class SupportsColorTestCase(unittest.TestCase):
+
+    def assertSupportsColor(self, env):
+        self._assert_color(True, env)
+
+    def assertNotSupportsColor(self, env):
+        self._assert_color(False, env)
+
+    def _assert_color(self, expected, env):
+        with mock.patch.dict('os.environ', env, clear=True):
+            self.assertEqual(expected, cli.supports_color())
+
+    def test_no_color_set(self):
+        self.assertNotSupportsColor({'NO_COLOR': '1'})
+        self.assertNotSupportsColor({'NO_COLOR': '0'})
+        self.assertNotSupportsColor({'NO_COLOR': 'whatever'})
+        self.assertNotSupportsColor({'NO_COLOR': '1', 'FORCE_COLOR': '1'})
+
+    def test_force_color_set(self):
+        self.assertSupportsColor({'FORCE_COLOR': '1'})
+        self.assertSupportsColor({'FORCE_COLOR': '0'})
+        self.assertSupportsColor({'FORCE_COLOR': 'whatever'})
+
+    def test_empty_no_color_does_not_change_default(self):
+        with mock.patch.dict('os.environ', clear=True):
+            if cli.supports_color():
+                self.assertSupportsColor({'NO_COLOR': ''})
+            else:
+                self.assertNotSupportsColor({'NO_COLOR': ''})
+
+    def test_empty_force_color_does_not_change_default(self):
+        with mock.patch.dict('os.environ', clear=True):
+            if cli.supports_color():
+                self.assertSupportsColor({'FORCE_COLOR': ''})
+            else:
+                self.assertNotSupportsColor({'FORCE_COLOR': ''})
 
 
 class CommandLineTestCase(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -45,6 +46,42 @@ def utf8_available():
...
+
+    # Tests
+
+    def test_no_color_set(self):
+        self.assertNotSupportsColor({'NO_COLOR': ''})
+        self.assertNotSupportsColor({'NO_COLOR': '1'})
+        self.assertNotSupportsColor({'NO_COLOR': '1', 'FORCE_COLOR': '1'})
+
+    def test_force_color_set(self):
+        self.assertSupportsColor({'FORCE_COLOR': ''})
```
adrienverge: This looks like an error, right?

From https://force-color.org/, if `FORCE_COLOR` is empty, it should be ignored:

> When this variable is present and not an empty string, it should force the addition of ANSI color.

yamllint/cli.py
```diff
@@ -38,6 +38,13 @@ def find_files_recursively(items, conf):
 
 
 def supports_color():
+    no_color = os.environ.get('NO_COLOR')
+    force_color = os.environ.get('FORCE_COLOR')
+    if (no_color is not None and no_color != '0'):
```
adrienverge: See previous comment, I believe there is an error, it should be:
```suggestion
    if no_color and no_color != '0':
```
in order to handle empty strings as "undefined".

adrienverge (commented):
Hello Antoine,

Thanks for this proposal. `NO_COLOR` and `FORCE_COLOR` seem pretty standard, so it looks like a good idea :+1:

Please see my comments in the code, and the tests must be reworked:
- Do no test the internal function `cli.supports_color()`, instead, test the real yamllint program. See inside in `CommandLineTestCase`, there are already `test_run_piped_output_nocolor()`, `test_run_format_colored()`, `test_run_format_colored_warning()`. Please add the new tests there.
- Also test `NO_COLOR` and `FORCE_COLOR` in conjunction with `--format standard` and `--format colored` (the command line argument must win).
- No need for comments `# Assertion helpers` and `# Tests`: the readers will infer that :wink:

tests/test_cli.py
hashar: It is **definitely** an error which has two causes:

- I did not read the spec, despite it being written in bold, it is again a case of Read The Fine Manual. Thank you for having quoted it.

- `no_color[0] != '\0'`, I have misread that as _the first character is zero_ and my implementation does not check that.  I took a few courses at C language 30 years ago went for pascal and assembly instead and never looked back. That `\0` is `NULL` or the end of a string eg it is an empty string :-]

I'll fix it later today. **thank you**.

yamllint/cli.py
hashar: I have removed the `!= '0'`  since I have misread the c code `\0` as being a zero when it is about detecting a NULL terminated string, ie an empty string.

tests/test_cli.py
hashar: I have changed the logic to check the absence of the env variable and it NOT being an empty string:
```
    if (no_color is not None and no_color != ''):
        return False
    if (force_color is not None and force_color != ''):
        return True
```

hashar:
TODO address the items mentioned in the cover message which I have missed:

> - [ ] Do no test the internal function cli.supports_color(), instead, test the real yamllint program. See inside in CommandLineTestCase, there are already test_run_piped_output_nocolor(), test_run_format_colored(), test_run_format_colored_warning(). Please add the new tests there.
> - [ ] Also test NO_COLOR and FORCE_COLOR in conjunction with --format standard and --format colored (the command line argument must win).
> - [ ] No need for comments # Assertion helpers and # Tests: the readers will infer that 😉

That will be for after the lunch break :]

hashar:
> [x] Do no test the internal function cli.supports_color(), instead, test the real yamllint program. See inside in CommandLineTestCase, there are already test_run_piped_output_nocolor(), test_run_format_colored(), test_run_format_colored_warning(). Please add the new tests there.

I have kept the unit tests covering `supports_color()` I think there is value in them and I guess they can be extended to cover Windows/ANSICON/TERM/whether stdout is a tty.

> [x] Also test NO_COLOR and FORCE_COLOR in conjunction with --format standard and --format colored (the command line argument must win).

I have added test cases for the following combinations:

| Format | Env | Colored?
|--|--|--
| auto | NO_COLOR |  :no_entry_sign: 
| auto | FORCE_COLOR |:ok:
| standard | NO_COLOR | :no_entry_sign: 
| standard | FORCE_COLOR |  :no_entry_sign: 
| colored | NO_COLOR | :ok:
| colored | FORCE_COLOR | :ok:

> [x] No need for comments # Assertion helpers and # Tests: the readers will infer that 😉

I have removed them :-]

PR #815, opened 2026-06-26
truthy: Don't carry %YAML version into the next document

sarathfrancis90:
The `truthy` rule reads a document's YAML version from its `%YAML` directive, but it only forgets that version on a `...` document-end marker. When documents are separated by `---` alone, the previous document's version leaks into the next one — which has no directive of its own and should fall back to YAML 1.1:

    %YAML 1.2
    ---
    on: 1
    ---
    on: 2

`on` in the second document is a 1.1 truthy value and should be flagged, but the leaked 1.2 version silently suppresses it.

I reset the cached version when a `---` marker starts a document that has no directive of its own (a directive always precedes the marker, so a document that declares one keeps it). Added a test for the 1.2 → directive-less case; the existing multi-document test still passes. flake8, ruff and the full unittest suite are green.

### DIFF
--- a/tests/rules/test_truthy.py
+++ b/tests/rules/test_truthy.py
@@ -186,6 +186,19 @@ class TruthyTestCase(RuleTestCase):
                    'boolean6: !!bool NO\n',
                    conf)
 
+    def test_explicit_yaml_version_does_not_leak_to_next_document(self):
+        conf = ('truthy: enable\n'
+                'document-start: disable\n')
+        # A %YAML directive only applies to the document it introduces. A
+        # following directive-less document (separated by '---', without
+        # '...') must be linted as YAML 1.1, where 'on' is truthy.
+        self.check('%YAML 1.2\n'
+                   '---\n'
+                   'on: 1\n'
+                   '---\n'
+                   'on: 2\n',
+                   conf, problem1=(5, 1))
+
     def test_check_keys_disabled(self):
         conf = ('truthy:\n'
                 '  allowed-values: []\n'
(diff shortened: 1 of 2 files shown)


sarathfrancis90:
The Linters job is red on the `make -C docs linkcheck` step — it hits a 404 on `https://yaml.org/spec/1.2.2/#52-character-encodings` (yaml.org now returns Not Found for that path). That link lives in `docs/character_encoding.rst` and `yamllint/decoder.py`, so it's unrelated to this change; the unit tests pass on 3.10–3.14. Glad to send a separate PR to update the dead spec link if that'd help.

adrienverge (approved):
Perfect, thanks @sarathfrancis90!


> The Linters job is red on the `make -C docs linkcheck` step — it hits a 404 on `https://yaml.org/spec/1.2.2/#52-character-encodings` (yaml.org now returns Not Found for that path). That link lives in `docs/character_encoding.rst` and `yamllint/decoder.py`, so it's unrelated to this change; the unit tests pass on 3.10–3.14. Glad to send a separate PR to update the dead spec link if that'd help.

I'm a bit surprised that yaml.org changed their website so much, and broke permalinks to the specifications. It could be an error, because the home page https://yaml.org/ has a link to the spec 1.2.2, which points to this precise https://yaml.org/spec/1.2.2/ and triggers the same HTTP 404. So I'm in favor of not fixing anything in yamllint documentation for now, since it looks like a temporary error.
EDIT: It looks like they are aware of it: https://github.com/yaml/www.yaml.org/issues/193

PR #819, opened 2026-07-19
CI: Add support for Python 3.15

DimitriPapadopoulos:


.github/workflows/tests.yaml
```diff
@@ -43,16 +43,18 @@ jobs:
           - '3.11'
           - '3.12'
           - '3.13'
-          - '3.13t'
+          - '3.13'
```
adrienverge: I understand that you removed Python 3.13t because *free-threading* is tested in 3.14t below. Right?

In this case, you can remove the double `'3.13'` line.

adrienverge (commented):
Thank you Dimitri

.github/workflows/tests.yaml
DimitriPapadopoulos: Fixed. The rationale was that 3.13t was a test release, not supported anymore, and 3.14t is the first real free-threaded release.

adrienverge: OK!

PR #823, opened 2026-07-27
quoted-strings: Fix false positive on flow values starting with ':'

### LINKED ISSUE #818: [quoted-strings] false positive on flow-mapping values starting with ':'
## Summary

With `quoted-strings: {required: only-when-needed}`, yamllint flags a double-quoted string starting with `:` inside a **flow mapping** as redundantly quoted — but removing the quotes produces a real YAML parse error.

## Environment

- yamllint 1.38.0 (also reproduced on older 1.x)
- Python 3.12, PyYAML 6.x

## Reproduction

`.yamllint`:
```yaml
rules:
  quoted-strings:
    quote-type: double
    required: only-when-needed
```

`test.yaml`:
```yaml
foo:
  within: {dimension: email_created_at, last: {days: ":days"}}
```

```
$ yamllint test.yaml
test.yaml
  2:54  error  string value is redundantly quoted with double quotes  (quoted-strings)
```

## Expected behavior

The quotes are **required**. Removing them yields invalid YAML:

```
$ python3 -c 'import yaml; yaml.safe_load("foo: {days: :days}")'
yaml.parser.ParserError: while parsing a flow node
expected the node content, but found ':'
```

The leading `:` inside a flow mapping is interpreted as a flow-value indicator, so the scalar must be quoted. yamllint's own parser agrees — it reports `syntax error: expected the node content, but found ':'` on the unquoted form.

## Actual behavior

yamllint reports the quoted form as redundantly quoted, forcing users to choose between a lint error and a parse error.

## Related

#799 (self-closed as "already fixed") covered *trailing*-colon strings in **block** context (`- text: "foo:"`). This report is a distinct case: *leading* colon in **flow-mapping** context.

## Workaround

Rewrite the mapping in block style — a value of `:days` parses fine unquoted outside a flow mapping:

```yaml
foo:
  within:
    dimension: email_created_at
    last:
      days: :days
```

soodoku:
Fixes #818.

With `required: only-when-needed`, a quoted scalar starting with `:` inside a flow mapping is reported as redundantly quoted, but removing the quotes produces invalid YAML:

```yaml
foo:
  within: {dimension: email_created_at, last: {days: ":days"}}
```

```
  2:54  error  string value is redundantly quoted with double quotes  (quoted-strings)
```

```python
>>> yaml.safe_load('foo: {days: :days}')
yaml.parser.ParserError: while parsing a flow node
expected the node content, but found ':'
```

`_quotes_are_needed()` decides this by re-parsing the value and seeing whether it survives as a plain scalar, but it always builds the probe in block context:

```python
loader = yaml.BaseLoader('key: ' + token.value)
```

`key: :value` is valid, so the quotes look redundant — while the token actually lives in a flow mapping, where `{key: :value}` is not valid. The function already receives `is_inside_a_flow`; it just wasn't using it here.

This builds the probe in whichever context the token appears in, and expects the matching end token. The surrounding logic is unchanged: the flow probe yields the same five leading tokens, and for `:value` the scanner emits a `ValueToken` where a `ScalarToken` is expected, so the existing check already concludes the quotes are needed.

Block-context behaviour is untouched — `k5: :wq` unquoted and `k6: ":wq"` flagged, as covered by the existing corner-case tests.

## Verification

- `python -m unittest discover`: 432 tests before, 433 after, all passing — the only delta is the added test.
- The new test fails on master (it reports the extra flow-mapping problem) and passes with the change; I confirmed both directions by reverting just the rule file.
- The reproduction from the issue now exits 0.
- `flake8 .` clean, and `yamllint --strict $(git ls-files '*.yaml' '*.yml')` clean.

The added test also asserts that genuinely redundant quotes inside a flow mapping (`flow: {k1: "value"}`) are still reported, so this narrows the rule rather than disabling it there.

I left `CHANGELOG.rst` alone, since recent contributor PRs don't seem to touch it — happy to add an entry if you'd prefer.


soodoku:
The red `Linters` job here isn't caused by this PR — `master` itself is currently failing.

ruff 0.16.0 (2026-07-23) stabilized `ISC004` out of preview, and since `ruff` is unpinned in `dependency-groups.dev`, CI started enforcing it against pre-existing code. All 7 errors are in `tests/test_cli.py` and `tests/test_config.py`, neither of which this branch touches; running ruff 0.16.0 against a clean `master` tree reproduces them identically.

Opened #824 to fix those. I'll rebase this branch once that lands.

DimitriPapadopoulos:
@soodoku Don't forget to rebase to fix the linter issue.

PR #824, opened 2026-07-27
style: Apply ruff rule ISC004

soodoku:
ruff 0.16.0 (released 2026-07-23) [stabilized](https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#stabilization) `ISC004` ([`implicit-string-concatenation-in-collection-literal`](https://docs.astral.sh/ruff/rules/implicit-string-concatenation-in-collection-literal)) out of preview.

Since `ruff` is unpinned in `dependency-groups.dev` and `pyproject.toml` selects the whole `ISC` family, the `Linters` job now fails on `master` and on every open PR with 7 errors in `tests/test_cli.py` and `tests/test_config.py`. Master's last green run was 2026-07-20, three days before the release.

This applies the rule to the 7 affected sites by parenthesizing the concatenations. Where the extra indentation would have pushed a line past the 79-column limit, the enclosing `assertEqual(...)` call is wrapped instead of re-splitting the strings, so **every string literal is byte-identical to `master`** — verified by comparing all `ast.Constant` string nodes before and after (435/435 in `test_cli.py`, 539/539 in `test_config.py`).

Verified locally with ruff 0.16.0: `ruff check`, `flake8`, `doc8`, `rstcheck`, `yamllint --strict` and `make -C docs html` all clean, and `python -X warn_default_encoding -W error::EncodingWarning -m unittest discover` passes 432 tests.

Separately, you may want to consider upper-bounding `ruff` in `dependency-groups.dev` so that a ruff release can't turn unrelated PRs red — but that's a policy call, and the `style: Apply ruff rule X` history suggests you may prefer riding latest.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


adrienverge:
Thank you, let's fix the style issue.

PR #827, opened 2026-08-01
tests: make test_find_files_recursively independent of TMPDIR ordering

### LINKED ISSUE #373: test_cli.py: test_find_files_recursively() breaks when build directory sorts before `/etc`
see https://github.com/adrienverge/yamllint/blob/0fff4e29e42d170cd0d1ad17736acdc53df97fc9/tests/test_cli.py#L152-L158

The semantics of `os.path.join` are such that:

```python
import os
pwd = '/an/inconvenient/dir'
absolute = os.path.join(pwd, '/etc/another/file')
relative = os.path.join(pwd, 'sub/directory.yaml/empty.yml')
absolute < relative  # False
pwd = '/home/user'
relative = os.path.join(pwd, 'sub/directory.yaml/empty.yml')
absolute < relative  # True
```

Recommend turning `/etc/another/file` into a relative path.

Eljees:
`test_find_files_recursively` fails whenever the temporary directory sorts before `/etc`, although nothing is broken.

Two things line up to cause it: `os.path.join(self.wd, '/etc/another/file')` silently discards `self.wd` (the second argument is absolute), and the expected list for that case is hand-ordered on the assumption that `/etc/another/file` sorts *before* the working directory - which is only true when the workspace lives under `/tmp`. Run the suite with `TMPDIR=/dev/shm/ylt` (as, for example, nixpkgs builds do - see NixOS/nixpkgs#118304) and the test fails:

```
TMPDIR=/dev/shm/ylt python -m unittest tests.test_cli.CommandLineTestCase.test_find_files_recursively
...
FAILED (failures=1)
```

The fix passes the absolute path explicitly (so the intent - a non-existing explicit path is returned as-is - is visible) and sorts the expected list, making the case independent of where the workspace lives. The full suite is green under both the default TMPDIR and `/dev/shm`: `Ran 432 tests ... OK (skipped=3)` (Python 3.12).

Fixes #373.

AI-assisted (LLM used for drafting); the analysis and runs are mine.


Eljees:
Ping on this one and #826 - both open since 1 August with no review.

On the red `Linters` job: it does not come from either patch. `ruff` is unpinned in the `dev` dependency group while `ISC` is in `extend-select`, so CI installs the newest ruff; 0.16.1 started flagging ISC004 in code that predates both PRs - six occurrences in `tests/test_cli.py` (384, 391, 754, 777, 813, 822) and one in `tests/test_config.py:869`. The same seven errors reproduce on unmodified master. #824 by @soodoku already applies that rule, so I have not duplicated it here.

#827 makes `test_find_files_recursively` independent of TMPDIR ordering; #826 strips the leading `./` from paths in `--format github` output.

Eljees:
Merged current `master` in (0dcc198). The red `Linters` here was not this branch:

```
tests/test_cli.py:384:28: ISC004 Unparenthesized implicit string concatenation in collection
tests/test_cli.py:391:28: ISC004 ...
tests/test_cli.py:754:38, 777:38, 813:30, 822:30
tests/test_config.py:869:21
```

Seven findings, none of them on a line this branch touches - it changes four lines in one test. They were fixed on `master` by f4ad063 ("style: Apply ruff rule ISC004"), which landed after this PR's last run, so the run was still being made against the base that had them.

Net diff against `master` is unchanged at four lines in `tests/test_cli.py`.

soodoku (approved):

Eljees:
Status note, since the thing I flagged in August has been resolved by someone else in the meantime.

On 14 August I wrote that the red `Linters` job here was not from either patch: `ruff` is unpinned in the `dev` dependency group while `ISC` is in `extend-select`, so CI picked up a newer rule set than the code was written against. That is settled — #824 landed the `ISC004` fixes on 19 August, and after the merge of current `master` into this branch all eight jobs on `1b17698b` pass:

`Unit tests` on 3.10, 3.11, 3.12, 3.13, 3.14, 3.14t and 3.15, plus `Linters`. `mergeable` is `true`, state `clean`.

@soodoku approved this one on 4 September. Nothing has been pushed to the branch since, so the approval still applies to the head that is green.

Mentioning it because #835 went in on 19 September, so the queue is clearly moving and this may simply have slipped past rather than being held for a reason. Four changed lines in one test; it makes `test_find_files_recursively` independent of the order `TMPDIR` happens to yield.

#826 is the sibling from the same day and is also clean and green, though nobody has looked at it yet. Happy to rebase either of them, and equally happy to hear that one is not wanted.

Issue #771, opened 2025-08-11
TypeError in indentation rule when parsing invalid YAML file


scharfy:
When running yamllint (v1.37.1) on Python 3.11, I encountered a TypeError in the indentation rule when processing a specific YAML file used in my unit tests.
The file is intentionally invalid YAML (garbage content).

workflow-invalid.yml
```yaml
THI IS:
}¬} invlid"
    >YAML
```

Unfortunately, I have not been able to create a minimal reproducible example from scratch — the crash only happens with this particular file in my project.

When I add this file to the ignore section in .yamllint, the error disappears and yamllint runs normally again.

```bash
Traceback (most recent call last):
  File "/home/user/.local/bin/yamllint", line 8, in <module>
    sys.exit(run())
  File "/home/user/.local/lib/python3.11/site-packages/yamllint/cli.py", line 227, in run
    prob_level = show_problems(problems, file, args_format=args.format,
  File "/home/user/.local/lib/python3.11/site-packages/yamllint/cli.py", line 102, in show_problems
    for problem in problems:
  File "/home/user/.local/lib/python3.11/site-packages/yamllint/linter.py", line 201, in _run
    for problem in get_cosmetic_problems(buffer, conf, filepath):
  File "/home/user/.local/lib/python3.11/site-packages/yamllint/linter.py", line 136, in get_cosmetic_problems
    for problem in rule.check(rule_conf,
  File "/home/user/.local/lib/python3.11/site-packages/yamllint/rules/indentation.py", line 582, in check
    yield from _check(conf, token, prev, next, nextnext, context)
  File "/home/user/.local/lib/python3.11/site-packages/yamllint/rules/indentation.py", line 343, in _check
    if expected < 0:
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

Expected behavior:
Invalid YAML should result in a parsing error message, not a crash.

Workaround:
Adding the file to .yamllint’s ignore list resolves the problem.

Environment:

yamllint: 1.37.1

Python: 3.11

OS: Ubuntu 24.04.3 LTS

Maybe just make sure expected is never NoneType?

Greetings from austria

mike-whittaker-work:
Sorry if this seems like a "me too" , but this error has occurred in our build pipeline as well. The None should be caught and processed correctly, regardless of having the test file available.

> > $ yamllint -d relaxed --no-warnings -f colored .
> Traceback (most recent call last):
[...]

PR #829, opened 2026-08-17
Fix IndexError in key-ordering on unmatched closing token

VXNCXNX:
The `key-ordering` rule crashes with an `IndexError` on invalid YAML that closes a flow collection which was never opened, instead of reporting the syntax error.

```
$ printf -- '---\n{a: 1, b: 2}}\n' > y.yaml
$ yamllint -d "{rules: {key-ordering: enable}}" y.yaml
  File "yamllint/rules/key_ordering.py", line 142, in check
    context['stack'].pop()
IndexError: pop from empty list
```

`[a, b, c]]` does the same. On invalid input PyYAML can emit a closing token with no matching opening one, leaving the stack empty. `key_duplicates.py` already guards its `pop()` for this reason, so this applies the same `if len(context['stack']) > 0` guard to `key_ordering.py`, which was missed. It is the sibling of #771 / PR #816, which fixed the pattern in `indentation.py` only.

After the fix the same file reports `2:13 error syntax error: expected '<document start>', but found '}'`. Test added covering the mapping and sequence forms. Full suite passes (433), flake8 clean.

AI disclosure: written with Claude Code. I ran the CLI before and after and checked the mutation myself.


VXNCXNX:
Rebased onto current `master`, no conflicts. Full suite on the new base: 433 passed, 6 subtests passed.

Reverting `key_ordering.py` to `master` still reproduces the crash the test targets, so the coverage is real:

```
E   IndexError: pop from empty list
yamllint/rules/key_ordering.py:142: IndexError
FAILED tests/rules/test_key_ordering.py::KeyOrderingTestCase::test_unmatched_closing_flow_token
```

PR #830, opened 2026-08-17
tests: Wrap strings in parentheses for ISC004

VXNCXNX:
ruff 0.16.3 added ISC004 (unparenthesized implicit string concatenation in a collection). It fires on 7 pre-existing sites in `tests/test_cli.py` and `tests/test_config.py`, so the Linters job now fails on every pull request:

```
tests/test_cli.py:384:28: ISC004
tests/test_cli.py:391:28: ISC004
tests/test_cli.py:754:38: ISC004
tests/test_cli.py:777:38: ISC004
tests/test_cli.py:813:30: ISC004
tests/test_cli.py:822:30: ISC004
tests/test_config.py:869:21: ISC004
```

`ISC` is in `extend-select` and ruff is not pinned in the workflow, so CI picked the new rule up on its own. master still shows green because it was last checked with an older ruff.

I wrapped the strings in parentheses rather than adding an ignore, since the rule looks intentional. No string value changes: I compared the string constants of both files before and after via the AST, 435 and 539 respectively, identical in both. `ruff check` is clean, the suite passes (432, plus 6 subtests) and flake8 is clean.

This is separate from #829 and branched from master, so they can go in either order.

AI disclosure: written with Claude Code. I reproduced the failure locally, made the change, and checked the constants myself.


adrienverge:
Hello @VXNCXNX and Claude, thank you for fixing this issue. A pull request was already open for that: https://github.com/adrienverge/yamllint/pull/824.

VXNCXNX:
You are right, #824 covers the same two files and it merged yesterday. I should have found it before opening this. Closing.

Issue #757, opened 2025-06-18
Support custom indent for sequences


psss:
Would it be possible to add support for custom number of spaces to be used for indenting sequences? Consider the following example:

```yaml
summary:
    Basic smoke test for the httpd web server
provision:
    how: virtual
    memory: 4096
prepare:
  - name: packages
    how: install
    package: [httpd, curl]
  - name: service
    how: shell
    script: systemctl start httpd
execute:
    how: tmt
    script:
      - echo foo > /var/www/html/index.html
      - curl http://localhost/ | grep foo
```

Or this one:

```yaml
summary:
    Support for multiple configurations
description:
    All steps support multiple configurations. In this way it is
    possible for example to discover tests from multiple repos.
    The 'execute' steps supports only a single configuration.
discover:
  - name: tier0
    how: fmf
    filter: 'tier: 0'
    url: https://github.com/teemtee/tmt
  - name: tier1
    how: fmf
    filter: 'tier: 1'
    url: https://github.com/teemtee/tmt
  - name: all
    how: fmf
    url: https://github.com/teemtee/tmt
prepare:
  - how: ansible
    name: packages
    playbook: ansible/packages.yml
  - how: shell
    name: services
    script: systemctl start service
execute:
    how: tmt
```

I find this way of indenting the sequence content visually in the same way as corresponding dict content as nicely consistent. The values, regardles of their type, all share the same level of indentation. But currently it's not possible with `yamllint` to enforce this.

What about introducing a new config option, let's say `sequence-spaces`, which would allow to specify the number of spaces to be enforced for sequences? Or allow `indent-sequences` to accept an integer which would specify the number of the desire spaces before the `-` character?

adrienverge:
Hello,

This is an interesting idea. I would be in favor of extending `indent-sequences` to accept integers, rather than adding a new option.

However I think it would be a substantial work to implement such a change, especially to adapt and enhance all tests of the `indentation` rule.

PR #832, opened 2026-08-19
indentation: Allow indent-sequences to be an integer

ChrisJr404:
This extends the indentation rule so that `indent-sequences` also accepts a non-negative integer, closing #757, where you were in favor of teaching the option to take integers rather than adding a separate one.

The four existing values (`true`, `false`, `whatever`, `consistent`) only let you say whether a block sequence is indented, not by how much: `true` derives the amount from `spaces`, and `false` forbids indentation entirely. That leaves no way to line the `-` up with the mapping values around it, which is what the reporter was after:

```yaml
prepare:
    - name: packages
      how: install
    - name: service
      how: shell
```

With an integer the `-` is required to sit exactly that many spaces past its parent key, independently of `spaces`. `0` ends up behaving like `false` and a value equal to `spaces` behaves like `true`, so the old shortcuts still read naturally. Everything else is untouched and the default stays `true`.

I added rule tests covering the fixed offset (including a case where it differs from `spaces`, and the `0` case), a config test that the integer form is accepted, and a docs example. The full suite, flake8, doc8 and the docs build all pass.


ChrisJr404:
Heads up: the Linters job is red, but it's not from this change — ruff 0.16.3 (just released) started flagging ISC004 on some pre-existing implicit string concatenations in tests/test_cli.py, so it fails the same way on master and on other open PRs. The unit test matrix here is green across 3.10-3.15. Happy to fold a fix for those into a separate PR if that's useful.

DimitriPapadopoulos:
@ChrisJr404 Rebasing should fix the linter issue.

PR #836, opened 2026-09-19
style: Apply more ruff rules

DimitriPapadopoulos:


adrienverge:
I'm OK to apply FURB105, however I don't fully agree with SIM105, for readability reasons. Is it possible to either drop SIM105, or split into 2 pull requests?

DimitriPapadopoulos:
The SIM105 fix is quite readable. Perhaps it's just you weren't aware of `suppress`

Besides, agreeing or not is not the issue. Python has grown for a simple to a complex language in the last decennia. To tame that complexity it's best to standardise, like contemporary languages such as Go and Rust do, and `ruff check` and `ruff format` are the current golden standard for Python.

adrienverge:
I'm aware of `suppress` and sometimes use it, but I don't want to force its usage. In my opinion, there are certain cases where an explicit `except FooError:` is better for readability / understanding.

DimitriPapadopoulos:
Again, personal opinion don't matter. What **is** best for readability / understanding.is simplification trough standardisation. Standardisation is pushed by people who have spent quite some time weighing the pros and cons of each rule, mote than us in any case. Also, that's what current languages like Go do, they have learnt from the errors of the past.

adrienverge:
I understand your point of view, but I don't agree. I my opinion, deciding "what is best for …" is quite subjective.

DimitriPapadopoulos:
A consensus is not subjective, a personal opinion is.

